PR #57, opened 2020-02-06
Update package documentation; clarify faucet deployment

### LINKED ISSUE #22: Reconsider docker deployment
The original faucet from IOV used Docker for deployment and was never pushed to npm. With this monorepo, we get npm deployment for free because we use the same process for all packages. Now we have a good amount of duplication in the build process (for npm and for docker).

I wonder of we should instead
1. Install the faucet build from npm inside the Docker. This would require an external Dockerfile, since the code is not yet uploaded to npm when we create a version tag.
2. Don't support Docker at all and instead document how to install and run via npm (like [it was done with @iov/cli](https://github.com/iov-one/iov-core/tree/v2.0.0-alpha.7/packages/iov-cli#installation-and-first-run)). Whoever wants to containerize it can do so with a few simple commands.

webmaster128:
Closes #22


docs/faucet.Dockerfile
```diff
@@ -0,0 +1,15 @@
+# The only officially supported distribution channel of the faucet binary is @cosmwasm/faucet on npmjs.com
+# This is an example file of how to wrap the faucet in a docker container.
+# This file will be outdated very soon, but should work as a template for your custom docker setup.
+
+# Coose from https://hub.docker.com/_/node/
+FROM node:12.14.1-alpine
+
+RUN yarn global add @cosmwasm/faucet@0.0.3
```
ethanfrey: Nice, published ;)

packages/faucet/README.md
```diff
@@ -74,29 +76,33 @@ as soon as there is
...
-docker run --read-only --rm cosmwasm/faucet:manual help
+docker run --read-only --rm local-cosmwasm-faucet:manual version
+docker run --read-only --rm local-cosmwasm-faucet:manual help
 ```
 
 - Run faucet locally
 
 ```sh
-DOCKER_HOST_IP=$(docker run --read-only --rm alpine ip route | awk 'NR==1 {print $3}') \
+DOCKER_HOST_IP=$(docker run --read-only --rm alpine ip route | awk 'NR==1 {print $3}'); \
```
ethanfrey: Nice trick

packages/sdk/README.md
```diff
@@ -0,0 +1,5 @@
+# @cosmwasm/sdk
+
+[![npm version](https://img.shields.io/npm/v/@cosmwasm/sdk.svg)](https://www.npmjs.com/package/@cosmwasm/sdk)
+
+An SDK to build CosmWasm clients.
```
ethanfrey: A good start. We will definitely need better docs, but let's get this working for us first

PR #58, opened 2020-02-06
Avoid rebuild in test job

webmaster128:
This copies the `build` results from one job to another. Naturally, this creates a dependency between the jobs, but reduces the overall resources required. Most importantly, we prepare a setup for a number of different tests that can run in parallel to the `test` job (e.g. browser tests, faucet cli tests, ...) and all rely on a single build.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,9 @@ workflows:
     jobs:
       - build
       - lint
-      - test
+      - test:
+          requires:
+            - build
 
 jobs:
   build:
@@ -30,6 +32,10 @@ jobs:
             - ~/.cache/yarn
       - run:
           command: yarn build
+      - persist_to_workspace:
+          root: .
+          paths:
+            - packages/*/build/*
   test:
     machine:
       # We can't use a containerized environment since it requires remote docker to start custom containers.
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -43,6 +49,11 @@ jobs:
       - run: # start early for less wait time below
           command: ./scripts/cosm/start.sh
           background: true
+      - attach_workspace:
+          at: /tmp/builds
+      - run:
+          name: Merge build folders into project (merge with hardlinks)
+          command: cp --recursive --link /tmp/builds/* .
```
ethanfrey: interesting command, something i wished to do before but didn't know how

.circleci/config.yml
```diff
@@ -5,7 +5,9 @@ workflows:
     jobs:
       - build
       - lint
-      - test
+      - test:
+          requires:
+            - build
```
ethanfrey: Dependency is explict and enforced in CI. Seems good to me.

.circleci/config.yml
webmaster128: I wonder if it is really possible to hardlink from the "mounted" workspace into the guest's main system. Depending on the implementation, those might be two different file systems. Anyways, it works and does the copy step in less than 1 second.

PR #59, opened 2020-02-06
Add @cosmwasm/cli

webmaster128:
~Based on #60~

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -1,0 +1,50 @@
+# @cosmwasm/cli
+
+[![npm version](https://img.shields.io/npm/v/@cosmwasm/cli.svg)](https://www.npmjs.com/package/@cosmwasm/cli)
+
+## Installation and first run
+
+The `cosmwasm-cli` executable is available via npm. We recommend local
+installations to your demo project. If you don't have one yet, just
+`mkdir cosmwasm-cli-installation && cd cosmwasm-cli-installation && yarn init --yes`.
+
+### locally with yarn
+
+```
+$ yarn add @cosmwasm/cli --dev
+$ ./node_modules/.bin/cosmwasm-cli
+```
+
+### locally with npm
+
+```
+$ npm install @cosmwasm/cli --save-dev
+$ ./node_modules/.bin/cosmwasm-cli
+```
+
+### globally with yarn
+
+```
+$ yarn global add @cosmwasm/cli
+$ cosmwasm-cli
+```
+
+### globally with npm
+
+```
+$ npm install -g @cosmwasm/cli
+$ cosmwasm-cli
+```
+
+## Getting started
+
+1. Install `@cosmwasm/cli` and run `cosmwasm-cli` as shown above
+2. TODO: write README inspired by
+   https://github.com/iov-one/iov-core/blob/master/packages/iov-cli/README.md
+
+## License
+
+This package is part of the cosmwasm-js repository, licensed under the Apache
+License 2.0 (see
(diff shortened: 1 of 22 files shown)


packages/cli/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "files": [
+    "build/",
+    "types/",
+    "tsconfig_repl.json",
+    "*.md",
+    "!*.spec.*",
+    "!**/testdata/"
+  ],
+  "dependencies": {
+    "@cosmwasm/sdk": "^0.0.3",
```
ethanfrey: Interesting, just our sdk lib, no bcp. I think this is a good idea.

However, I think the current sdk is a bit low-level sometimes (more or less a clean wrapper around the RestServer without too many deps). I would suggest adding (not this PR but soon) another package with a bit higher level helpers around sdk - adding imports for key management and signing transactions, also constructing messages.

This is another issue, but yeah my main thought here

packages/cli/src/cli.ts
```diff
@@ -0,0 +1,146 @@
...
+  if (args.version) {
+    const version = require(join(__dirname, "..", "package.json")).version;
+    console.info(version);
+    return;
+  }
+
+  const imports = new Map<string, readonly string[]>([
+    ["@cosmwasm/sdk", ["types", "RestClient"]],
+    [
+      "@iov/crypto",
```
ethanfrey: These three packages are not part of the dependencies. (crypto, encoding, keycontrol). This is related to my questions above. How do we pull all this in?

packages/cli/src/cli.ts
```diff
@@ -0,0 +1,146 @@
...
+        "Wallet",
+        "WalletId",
+        "WalletImplementationIdString",
+        "WalletSerializationString",
+      ],
+    ],
+  ]);
+
+  console.info(colors.green("Initializing session for you. Have fun!"));
+  console.info(colors.yellow("Available imports:"));
```
ethanfrey: Doesn't match above imports

webmaster128: Well spotted. This was a bug. Corrected.

packages/cli/package.json
webmaster128: One thing that really bothers me is that we can't do any kind of signing without BCP types. I imported all the BCP stuff now and added an example from the rest client tests.

webmaster128:
Added some example code. This can be merged I think.

packages/cli/README.md
```diff
@@ -0,0 +1,84 @@
...
+```
+$ npm install -g @cosmwasm/cli
+$ cosmwasm-cli
+```
+
+## Getting started
+
+1. Install `@cosmwasm/cli` and run `cosmwasm-cli` as shown above
+2. Start a local wasmd blockchain
+3. Start with `./bin/cosmwasm-cli --init examples/local_faucet.ts`
```
webmaster128: Ah damn, this can only work from within the development repo. Otherwise the "helper" file is not very helpful. This whole CLI idea is a mess .... 🤯

webmaster128: It feels like you can only lose when you try to balance between the necessary boilerplate to achive anything on the one side and flexibility of an API on the other side.

ethanfrey: Okay, I got it. This is a good place to play for now. 

For later, we can:

a. provide an example downloading the init file via `curl` with some sample setups
b. provide a second package (built into the binary) that provides a higher-level, opinionated API for common demo tasks, but doesn't create the objects

ethanfrey:
I need some time to play around with it more, but this is a great start

PR #61, opened 2020-02-06
Add a Pen for signing

webmaster128:
Utilize the power of @iov/crypto without polling in @iov/bcp and @iov/keycontrol.

If this is :+1:, I can port this to the cli, where this also removed the level* dependencies.

### DIFF
--- a/packages/sdk/src/pen.ts
+++ b/packages/sdk/src/pen.ts
@@ -1,0 +1,87 @@
+import {
+  Bip39,
+  EnglishMnemonic,
+  Secp256k1,
+  Sha256,
+  Sha512,
+  Slip10,
+  Slip10Curve,
+  Slip10RawIndex,
+} from "@iov/crypto";
+
+export type PrehashType = "sha256" | "sha512" | null;
+
+/**
+ * A pen is the most basic tool you can think of for signing. It works
+ * everywhere and can be used intuitively by everyone. However, it does not
+ * come with a great amount of features. End of semi suitable metaphor.
+ *
+ * This wraps a single keypair and allows for signing.
+ *
+ * Non-goals of this types are: multi account support, persistency, data migrations,
+ * obfuscation of sensitive data.
+ */
+export interface Pen {
+  readonly getPubkey: () => Promise<Uint8Array>;
+  readonly createSignature: (signBytes: Uint8Array, prehashType?: PrehashType) => Promise<Uint8Array>;
+}
+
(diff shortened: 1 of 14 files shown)


packages/sdk/src/pen.ts
```diff
@@ -0,0 +1,86 @@
...
+    Slip10RawIndex.hardened(118),
+    Slip10RawIndex.hardened(0),
+    Slip10RawIndex.normal(0),
+    Slip10RawIndex.normal(a),
+  ];
+}
+
+export class Secp256k1Pen implements Pen {
+  public static async fromMnemonic(
+    mnemonic: string,
```
ethanfrey: I like this interface, and for all single signer examples works well.

Default path with override is very nice

packages/sdk/src/pen.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  private constructor(privkey: Uint8Array, pubkey: Uint8Array) {
+    this.privkey = privkey;
+    this.pubkey = pubkey;
+  }
+
+  /**
+   * Creates a fixed length encoding of the signature parameters r (32 bytes) and s (32 bytes).
+   */
+  public async createSignature(
```
ethanfrey: This is the same output as the wallet, right?

webmaster128: Almost. IOV Core uses a custom format which we call "fixed length encoding":

```ts
  /**
   * A simple custom encoding that encodes the extended signature as
   * r (32 bytes) | s (32 bytes) | recovery param (1 byte)
   * where | denotes concatenation of bonary data.
   */
  public toFixedLength(): Uint8Array {
    return new Uint8Array([...this.r(32), ...this.s(32), this.recovery]);
  }
```

Cosmos also has a fixed length encoding, but only wants the first 64 bytes. This is normalized inside of 

```ts
export function encodeSecp256k1Signature(pubkey: Uint8Array, signature: Uint8Array): StdSignature {
  return {
    // eslint-disable-next-line @typescript-eslint/camelcase
    pub_key: {
      type: pubkeyType.secp256k1,
      value: toBase64(Secp256k1.compressPubkey(pubkey)),
    },
    // Recovery seems to be unused
    signature: toBase64(Secp256k1.trimRecoveryByte(signature)),
  };
}
```

where `Secp256k1.trimRecoveryByte` cuts the first 64 bytes from the 64 bytes or 65 bytes format.

PR #62, opened 2020-02-07
Add script to deploy ERC20 contract for local development

webmaster128:
This checks-in contract b26861a6aa9858585ed905a590272735bd4fe8177c708940236224e8c9ff73ca from https://github.com/confio/cosmwasm-examples/pull/36


scripts/cosm/deploy_erc20.js
```diff
@@ -0,0 +1,122 @@
...
+const faucetAddress = "cosmos1pkptre7fdkl6gfrzlesjjvhxhlc3r4gmmk8rs6";
+
+async function uploadContract(client, pen, wasm) {
+  const memo = "My first contract on chain";
+  const storeCodeMsg = {
+    type: "wasm/store-code",
+    value: {
+      sender: faucetAddress,
+      wasm_byte_code: Encoding.toBase64(wasm),
+      source: "https://github.com/confio/cosmwasm/raw/0.7/lib/vm/testdata/contract_0.6.wasm",
```
ethanfrey: This is not the proper source. Better leave it blank, or take it as an argument.

(It confused me until I saw line 81)

scripts/cosm/deploy_erc20.js
```diff
@@ -0,0 +1,122 @@
...
+  console.info(`Upload succeeded. Code ID is ${codeId}`);
+
+  const initMsg = {
+    decimals: 5,
+    name: "Ash token",
+    symbol: "ASH",
+    initial_balances: [
+      {
+        address: faucetAddress,
+        amount: "11",
```
ethanfrey: Maybe we can give the faucet a bit more than 0.00011 ASH to hand out.

Otherwise, this looks good

scripts/cosm/deploy_erc20.js
```diff
@@ -0,0 +1,122 @@
...
+  return client.postTx(marshalTx(signedTx));
+}
+
+async function main() {
+  const pen = await Secp256k1Pen.fromMnemonic(faucetMnemonic);
+  const client = new RestClient(httpUrl);
+
+  const wasm = fs.readFileSync(__dirname + "/contracts/cw-erc20.wasm");
+  const uploadResult = await uploadContract(client, pen, wasm);
+  if (uploadResult.code) {
```
ethanfrey: I wonder sometimes if it wouldn't be safer for `client.postTx` to throw an error if `code` is set, so this doesn't get forgotten when people just work on the happy path. Basically, full these three lines into RestClient

scripts/cosm/deploy_erc20.js
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env node
```
ethanfrey: This is a super example script.
Not just for the CI, but to demo.

Great initiative!

packages/sdk/types/index.d.ts
```diff
@@ -1,7 +1,8 @@
+import * as logs from "./logs";
```
ethanfrey: :+1: some nice helpers there

packages/sdk/src/logs.ts
```diff
@@ -59,3 +59,26 @@ export function parseLogs(input: unknown): readonly Log[] {
...
   return input.map(parseLog);
 }
+
+/**
+ * Searches in logs for the first event of the given event type and in that event
+ * for the first first attribute with the given attribute key.
+ *
+ * Throws if the attribute was not found.
+ */
+export function findAttribute(
```
ethanfrey: Yeah, a few lines, but super nice to expose as a function

scripts/cosm/deploy_erc20.js
webmaster128: :)

packages/sdk/src/logs.ts
webmaster128: It was not possible to copy it over (which I tried first) since nodejs 12 does not support JavaScript's new optional chaining syntax. This way, it gets processed and prebuilt by TS.

scripts/cosm/deploy_erc20.js
webmaster128: Yeah, I removed the hardcoded values now to avoid blowing up the script

scripts/cosm/deploy_erc20.js
webmaster128: I prefer to wrap the RestClient in a higher level client that does this kind of application logic. I think this will evolve naturally from the helper functions we use at the moment.

webmaster128:
> No tests run against this, right? (which is fine)
> This is a great helper script

Thanks!

The script is not yet used, right. This is the base for ERC20 token support in CosmWasmConnection. As soon as we want to have those, we need a prepared token with a fixed address on the development chain.

ethanfrey:
Loving it!

PR #63, opened 2020-02-08
Add some address derivation fun to CLI

webmaster128:


packages/cli/src/cli.ts
```diff
@@ -121,9 +128,13 @@ export function main(originalArgs: readonly string[]): void {
       const hexHash = toHex(hash);
       export class NewDummyClass {};
 
-      const pen = await Secp256k1Pen.fromMnemonic(
-        "zebra slush diet army arrest purpose hawk source west glimpse custom record",
-      );
+      const encoded = toHex(toUtf8(toBase64(toAscii("hello world"))));
```
ethanfrey: Hahaha...funny to see the output.
Also interesting 2 of the `toX` convert bytes to string, and the other two string to bytes

packages/sdk/src/pubkey.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+  describe("encodeSecp256k1Pubkey", () => {
+    it("encodes a full signature", () => {
+      const pubkey = fromBase64("AtQaCqFnshaZQp6rIkvAPyzThvCvXSDO+9AzbxVErqJP");
+      expect(encodeSecp256k1Pubkey(pubkey)).toEqual({
+        type: "tendermint/PubKeySecp256k1",
+        value: "AtQaCqFnshaZQp6rIkvAPyzThvCvXSDO+9AzbxVErqJP",
+      });
+    });
+
+    it("compresses uncompressed public keys", () => {
```
ethanfrey: nice check... this was to explain to me, right?

webmaster128: Last time we talked about signature representation with and without recovery byte. This is about public key representations. Secp256k1 can be expressed as 2x 32 bytes with a `0x04` prefix (uncompressed), e.g.

`041d4c015b00cbd914e280b871d3c6ae2a047ca650d3ecea4b5246bb3036d4d74960b7feb09068164d2b82f1c7df9e95839b29ae38e90d60578b2318a54e108cf8`

Then there is a fancy way to compress this to 32 bytes plus a `0x02` or `0x03` prefix. This compression I never bothered trying to understand. It uses some properties of the curve.

The prefixes indicate the format that is used, which is why the secp256k1 pubkeys are either 65 or 33 bytes long. Ethereum always uses uncompressed, Cosmos always uses compressed.

PR #64, opened 2020-02-09
Add unused account for more explicit testing of authAccounts

webmaster128:
And all the other things I found along the way


packages/sdk/src/restclient.spec.ts
```diff
@@ -188,13 +198,40 @@ describe("RestClient", () => {
...
-      const { result } = await client.authAccounts(faucetAddress);
-      const account = result.value;
-      expect(account.account_number).toEqual(4);
-      expect(account.sequence).toBeGreaterThanOrEqual(0);
+      const { result } = await client.authAccounts(faucet.address);
+      expect(result.value).toEqual(
+        jasmine.objectContaining({
+          public_key: encodeBech32Pubkey(faucet.pubkey, "cosmospub"),
+        }),
+      );
```
webmaster128: Those two tests are the hear of this PR

scripts/cosm/template/.wasmd/config/genesis.json
```diff
@@ -19,18 +19,25 @@
   },
   "app_hash": "",
   "app_state": {
-    "slashing": {
+    "distribution": {
```
webmaster128: Unfortunately, all the order of all those genesis keys changes for every generate. Any idea how to fix them?

packages/bcp/src/address.spec.ts
```diff
@@ -17,23 +17,6 @@ describe("address", () => {
     });
   });
 
-  describe("isValidAddress", () => {
```
ethanfrey: Yeah, thanks for cleaning this up, there were some loose ends to simplify when I pulled out this logic

packages/bcp/src/cosmwasmcodec.ts
```diff
@@ -15,18 +22,18 @@ import {
...
 import { nonceToAccountNumber, nonceToSequence, TokenInfos } from "./types";
 
 export class CosmWasmCodec implements TxCodec {
-  private readonly prefix: CosmosBech32Prefix;
+  private readonly addressPrefix: CosmosAddressBech32Prefix;
   private readonly tokens: TokenInfos;
 
-  public constructor(prefix: CosmosBech32Prefix, tokens: TokenInfos) {
-    this.prefix = prefix;
+  public constructor(prefix: CosmosAddressBech32Prefix, tokens: TokenInfos) {
```
ethanfrey: I wonder about not just making this a string.
It is true for the hub, but most zones make their own prefixes. Eg. Xrn:... for regen

packages/sdk/src/pubkey.ts
```diff
@@ -1,11 +1,79 @@
...
 
 export function encodeSecp256k1Pubkey(pubkey: Uint8Array): PubKey {
   return {
     type: pubkeyType.secp256k1,
     value: Encoding.toBase64(Secp256k1.compressPubkey(pubkey)),
   };
 }
+
+export type CosmosPubkeyBech32Prefix = "cosmospub" | "cosmosvalconspub" | "cosmosvaloperpub";
+const validPubkeyPrefixes = ["cosmospub", "cosmosvalconspub", "cosmosvaloperpub"];
```
ethanfrey: Didn't even see this file. Definitely better place for the logic

packages/sdk/src/restclient.spec.ts
```diff
@@ -188,13 +198,40 @@ describe("RestClient", () => {
...
-    it("works", async () => {
+    it("works for unused account without pubkey", async () => {
+      pendingWithoutCosmos();
+      const client = new RestClient(httpUrl);
+      const { result } = await client.authAccounts(unusedAccount.address);
+      expect(result).toEqual({
+        type: "cosmos-sdk/Account",
+        value: {
+          address: unusedAccount.address,
+          public_key: "", // not known to the chain
```
ethanfrey: Question

I think i did a parse step here, so pubkey is always the decoded format, not the bech32 format. Which is much easier to use/check in all other code.

And now I think that code was in a pr I closed without merge. Anyway, I would like this to be an object or undefined. But I can do that separate

packages/sdk/src/restclient.spec.ts
ethanfrey: As above, I would adjust authAccounts to do the decode step, so we can compare directly with faucet.pubkey

scripts/cosm/template/.wasmd/config/genesis.json
ethanfrey: I see why it is annoying. This can't really be fixed sdk side. Best solution is to pipe this through a command that sorts the json.

This is how they compare exports in cosmos hub

packages/sdk/src/restclient.spec.ts
webmaster128: Such a parsing step exists in BCP CosmWasmConnection: `const pubkey = !account.public_key ? undefined : decodeCosmosPubkey(account.public_key);`

At this level I tend to pass through the REST response 1:1. However, I think there should be a higher CosmWasmClient around RestClient without BCP types that exposes the perfect TypeScript types.

packages/sdk/src/pubkey.ts
webmaster128: I only created it yesterday when extracting `encodeSecp256k1Pubkey`

packages/sdk/src/restclient.spec.ts
ethanfrey: Okay, I agree with this.

Rest server 1:1, bcp adapter for compatibility with tooling. And a to.be created higher level sdk for using one cosmwasm chain.

ethanfrey: A part of my inspiration is that the rest server types keep changing (more will be bechified, etc.) And want to give a relatively stable api to apps. That can be done in another class.

PR #66, opened 2020-02-09
Add read support for multiple ERC20 tokens

webmaster128:


webmaster128:
The order of the results in GET `/wasm/contract` seems random. After initializing tokens at `cosmos18vd8fpwxzck93qlwghaj6arh4p7c5n89uzcee5` and `cosmos1hqrdl6wstt8qzshwc6mrumpjk9338k0lr4dqxd` and then running tests, I get the following order

```js
[
  'cosmos18vd8fpwxzck93qlwghaj6arh4p7c5n89uzcee5',
  'cosmos1wgh6adn8geywx0v78zs9azrqtqdegufulvfwmp',
  'cosmos1sshdl5qajv0q0k6shlk8m9sd4lplpn6gwvnr96',
  'cosmos1hqrdl6wstt8qzshwc6mrumpjk9338k0lr4dqxd'
]
```

The two contracts in the middle positions have been initialized after the first and the last one.

It would be good to sort by initialization order.

One other thing that would help for the tests: list all instances for a given code ID. That would ensure the transmitted data is reasonably small.

ethanfrey:
Both very good points.

I have fixef other issues in the rest server. Can you make an issue on wasmd about this?

packages/sdk/src/restclient.ts
```diff
@@ -284,14 +284,14 @@ export class RestClient {
 
   // Makes a "smart query" on the contract, returns response verbatim (json.RawMessage)
   // Throws error if no such contract or invalid query format
-  public async queryContractSmart(address: string, query: object): Promise<unknown> {
+  public async queryContractSmart(address: string, query: object): Promise<string> {
```
ethanfrey: So. Would it be better if it returned a string like other calls, that you need to JSON.parse?

Cuz right now it parses it server side. Same as init_msg on the contract - the original object, not the string version

ethanfrey:
> It would be good to sort by initialization order.

I agree, but there are no secondary indexes in the sdk. Nor do we even store when initialized. We store them under contract address.

From the code, it seems like they should always be returned in sorted order (by address), but even this doesn't seem to be the case. I will add that to make them less random, but we need a bigger overhaul in storage in the next version

ethanfrey:
Addressed some fixes here: https://github.com/cosmwasm/wasmd/pull/68

packages/sdk/src/restclient.ts
webmaster128: There are multiple levels of parsing. The Rust `Result` level is parsed, but the binary data inside of `Ok(data)` is converted to a string. This diff reflect the fact that we always get a string here. At some point in the stack there must be a `Vec<u8>` to string conversion, that probably breaks when a contract decodes to send non-UTF8 binary data.

webmaster128:
Okay, thank, good to know. I'll unblock this PR by not relying on other tests's instances. So no pressure from my side.

webmaster128:
8a0e09241fc9202c7384e81ad4767bed1cb61f85 ensure the correct contract address is available for the tests, independent of the contract order from wasmd. So this is ready.

packages/bcp/src/cosmwasmcodec.ts
```diff
@@ -26,13 +26,13 @@ import { pubkeyToAddress } from "./address";
 import { Caip5 } from "./caip5";
 import { parseTx } from "./decode";
 import { buildSignedTx, buildUnsignedTx } from "./encode";
-import { nonceToAccountNumber, nonceToSequence, TokenInfos } from "./types";
+import { BankTokens, nonceToAccountNumber, nonceToSequence } from "./types";
 
 export class CosmWasmCodec implements TxCodec {
   private readonly addressPrefix: CosmosAddressBech32Prefix;
-  private readonly tokens: TokenInfos;
+  private readonly tokens: BankTokens;
```
ethanfrey: I like this name

packages/bcp/src/cosmwasmconnection.spec.ts
```diff
@@ -33,38 +34,63 @@ describe("CosmWasmConnection", () => {
...
-      denom: "ucosm",
-    },
-    {
-      fractionalDigits: 6,
-      name: "Staking Token",
-      ticker: "STAKE",
-      denom: "ustake",
-    },
-  ];
+  const defaultConfig: TokenConfiguration = {
```
ethanfrey: Very nice

packages/bcp/src/types.ts
```diff
@@ -16,7 +16,22 @@ export interface TokenInfo {
   readonly fractionalDigits: number;
 }
 
-export type TokenInfos = ReadonlyArray<TokenInfo>;
+export type BankTokens = ReadonlyArray<BankToken>;
+
+export interface Erc20Token {
+  readonly contractAddress: string;
```
ethanfrey: Besides this address, the rest we can query from the chain, right? Ticker and fractionalDigits 

It would be nice to eventually make the  cosmwasmconnection constructor just take the addresses and query the rest

packages/sdk/src/restclient.spec.ts
```diff
@@ -441,25 +446,28 @@ describe("RestClient", () => {
...
-      async function getContractAddress(): Promise<string> {
-        const contracts = Array.from(await client.listContractAddresses());
-        const last = contracts.reverse().find(() => true);
-        if (!last) throw new Error("No contract found");
-        return last;
-      }
+      let contractAddress: string | undefined;
+
+      beforeAll(async () => {
+        if (cosmosEnabled()) {
```
ethanfrey: Ah, good way to do setup without worrying about fit or xit.

packages/sdk/src/restclient.ts
ethanfrey: Let's make a todo to fix/test that on wasmd side

packages/bcp/src/types.ts
webmaster128: Ideally yes. However, I am not sure what information is needed to configure the codec. Let's see ...

ethanfrey: This is more an idea for the future when we load some config file with erc20 tokens, or auto-config. Explicit is very nice for tests

PR #67, opened 2020-02-10
Rest server fixes

ethanfrey:
Update the rest client to use/test the new fields in the rest server.

In particular, this tests: 

* `getCode(id)` to return full bytecode (now working)
* returns full error message when we get a 500 error (major debug help)
* added `listContractsByCodeId(id)` gets contract info for all associated with given bytecode

### DIFF
--- a/packages/sdk/src/restclient.spec.ts
+++ b/packages/sdk/src/restclient.spec.ts
@@ -385,6 +385,12 @@ describe("RestClient", () => {
       expect(lastInfo.id).toEqual(codeId);
       expect(lastInfo.creator).toEqual(faucet.address);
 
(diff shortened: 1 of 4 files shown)


packages/sdk/src/restclient.spec.ts
```diff
@@ -223,6 +231,7 @@ describe("RestClient", () => {
       });
     });
 
+    // this is failing for me on first run (faucet has not signed anything)
```
webmaster128: True. I considered this unproblematic as deployment of contracts should be part of the default chain start flow soon.

ethanfrey:
Thanks for the merge

PR #69, opened 2020-02-10
Debug start of blockchain in CI

webmaster128:


webmaster128:
This debugs the internal error in case it comes again. But seems like an unreliable issue.

PR #71, opened 2020-02-10
Add ERC20 send support for faucet

webmaster128:


scripts/cosm/deploy_erc20.js
```diff
@@ -118,7 +118,18 @@ async function main() {
       },
     ],
   };
-  for (const initMsg of [initMsgAsh, initMsgBash]) {
+  const initMsgCash = {
```
ethanfrey: Ah, here the famous 18 decimals... trying to overflow :wink:

packages/bcp/src/cosmwasmcodec.spec.ts
```diff
@@ -1,64 +1,128 @@
...
+        chainId: chainId,
+        sender: "cosmos1txqfn5jmcts0x0q7krdxj8tgf98tj0965vqlmq" as Address,
+        recipient: "cosmos1dddd" as Address,
+        memo: "My first BASH payment",
+        amount: {
+          fractionalDigits: 0,
+          quantity: "345",
+          tokenTicker: "BASH" as TokenTicker,
+        },
+        fee: {
```
ethanfrey: wow, almost 1/3 of this tx is just setting the fee and memo, which we would like to make automatic. Maybe use `defaultFee` for the tests. I definitely need to think about how to configure this better in production.

(I'm really looking at the API and what we can simplify in a basic API, without removing the power of the full API - I really think we need some wrapper class which makes lots of assumptions for intro devs, then they can pull off the wrapper and grab all the fields when they need it)

packages/bcp/src/cosmwasmconnection.spec.ts
```diff
@@ -433,5 +458,47 @@ describe("CosmWasmConnection", () => {
 
       connection.disconnect();
     });
+
+    it("can send ERC20 tokens", async () => {
```
ethanfrey: :smile:

packages/bcp/src/cosmwasmconnection.ts
```diff
@@ -172,26 +171,31 @@ export class CosmWasmConnection implements BlockchainConnection {
...
           return {
             fractionalDigits: erc20.fractionalDigits,
-            quantity: response.balance,
+            quantity: normalizedBalance,
             tokenTicker: erc20.ticker as TokenTicker,
           };
         },
       ),
     );
+    const nonZeroErc20Amounts = erc20Amounts.filter(amount => amount.quantity !== "0");
```
ethanfrey: This is both cool, and expensive if we have more than say 5-10 erc20 defined. (Have you seen the list on myetherwallet... it is like 100+). Anyway, if the problem we have is too many tokens on the chain, then we are doing good :joy:

packages/bcp/src/encode.ts
```diff
@@ -79,37 +77,83 @@ export function encodeFullSignature(fullSignature: FullSignature): types.StdSign
...
-    value: {
-      msg: [
-        {
-          type: "cosmos-sdk/MsgSend",
-          value: {
-            from_address: tx.sender,
-            to_address: tx.recipient,
-            amount: [encodeAmount(tx.amount, tokens)],
+
+  const matchingBankToken = bankTokens.find(t => t.ticker === tx.amount.tokenTicker);
```
ethanfrey: Very nice!

packages/faucet/src/faucet.spec.ts
```diff
@@ -81,7 +75,34 @@ describe("Faucet", () => {
   });
 
   describe("send", () => {
-    it("can send", async () => {
+    it("can send bank token", async () => {
```
ethanfrey: So nice..

scripts/cosm/deploy_erc20.js
webmaster128: Yeah I used Uint64 before but then had to change to to arbitrary length BN

packages/bcp/src/cosmwasmcodec.spec.ts
webmaster128: For those kind of encoder tests I prefer to be specific to allow seeing what is going on and to reduce dependencies as much as possible. The example in "can send ERC20 tokens" at least removes the fee noise.

packages/bcp/src/cosmwasmconnection.ts
webmaster128: The bigger problem I see there is the lack of a proper token identifier. Optimizing working code is relatively easy compared to writing a cross chain spec.

packages/faucet/src/faucet.spec.ts
webmaster128: Oh, I mixed up the titles. This is the contract test, not the bank test

packages/bcp/src/cosmwasmcodec.spec.ts
ethanfrey: Good call. This is encoding. Totally right, j was thinking usage, and it did seem less noisy

PR #72, opened 2020-02-10
Added some cli helpers

ethanfrey:
Based on #73 (as that changed the query API a fair bit)

A guide for running a demo and a bit more helpers for the cli.
For what it is worth


scripts/guide.md
```diff
@@ -0,0 +1,64 @@
...
+
+const execMsg = { transfer: {recipient: rcpt, amount: "808"}}
+const exec = await executeContract(client, pen, foo, execMsg);
+exec
+client.queryContractSmart(foo, { balance: { address: rcpt } })
+
+
+// TODO: unused account
+// TODO: execute and send tokens
+```
```
webmaster128: We have a very similar guide in packages/cli/README.md. What about just moving this file in the same direcory?

ethanfrey: Good point

packages/cli/examples/helpers.ts
```diff
@@ -0,0 +1,101 @@
...
+  if (result.code) {
+    throw new Error(`Failed tx: (${result.code}): ${result.raw_log}`)
+  }
+  const instantiationLogs = logs.parseLogs(result.logs);
+  const contractAddress = logs.findAttribute(instantiationLogs, "message", "contract_address").value;
+  return contractAddress;
+}
+
+// helper functions
+const executeContract = async (execClient: RestClient, execPen: Secp256k1Pen, contractAddr: string, msg: object, transferAmount?: types.Coin[]): Promise<readonly logs.Log[]> => {
```
webmaster128: 👍 For some reason, defining `function`s does not work in the CLI. I don't know why. Maybe because the definition is wrapped in an async block and then not exported.

ethanfrey: yeah, i had tried that first, then tried closures.

rebasing now

ethanfrey:
Rebased on master, please merge when green

PR #73, opened 2020-02-10
Update return types for queries

ethanfrey:
Related to work in https://github.com/cosmwasm/wasmd/pull/71

Early commits were doing local testing aganst the branch to update query client.

Now, prepared against docker image from merged wasmd PR.

### DIFF
--- a/packages/sdk/src/restclient.spec.ts
+++ b/packages/sdk/src/restclient.spec.ts
@@ -511,8 +511,8 @@ describe("RestClient", () => {
 
         // query by one key
         const raw = await client.queryContractRaw(contractAddress!, expectedKey);
-        expect(raw).not.toBeNull();
-        const model = JSON.parse(fromAscii(raw!));
+        assert(raw, "must get result");
+        const model = JSON.parse(fromAscii(raw));
         expect(model.verifier).toBeDefined();
         expect(model.beneficiary).toBeDefined();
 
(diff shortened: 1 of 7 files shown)


ethanfrey:
Okay, I tagged the wasmd release and this tests properly against it now.

packages/sdk/src/restclient.spec.ts
```diff
@@ -509,10 +510,11 @@ describe("RestClient", () => {
         pendingWithoutCosmos();
 
         // query by one key
-        const model = await client.queryContractRaw(contractAddress!, expectedKey);
-        expect(model).not.toBeNull();
-        expect((model as any).verifier).toBeDefined();
-        expect((model as any).beneficiary).toBeDefined();
+        const raw = await client.queryContractRaw(contractAddress!, expectedKey);
+        expect(raw).not.toBeNull();
+        const model = JSON.parse(fromAscii(raw!));
```
webmaster128: You can replace the `expect`ation here to `assert(raw, "Query result must not be falsy")` with `import { assert } from "@iov/utils";`. This helper throws when the expression is falsy and has an amazing new TypeScript signature that makes the static analyzier know that. `raw!` will then become `raw` since it is narrowed down to `Uint8Array`.

ethanfrey: `asserts condition` ?
https://github.com/iov-one/iov-core/blob/master/packages/iov-utils/src/assert.ts

Time for me to go back to the manual.

I'll make the update

webmaster128: This is super helpful in test code: https://github.com/iov-one/iov-core/pull/1352/files

PR #74, opened 2020-02-11
Add CosmWasmClient to SDK

webmaster128:


ethanfrey:
Looks big, I will review this afternoon.

Quick question - there are now two CosmWasmConnections?
A `bcp` version, and an `sdk` version?

I like having a higher level over the RestClient, but maybe we want different names?

webmaster128:
Thanks, no rush. 

> Quick question (I have not read the code yet) - there are now two CosmWasmConnections?

One is called "*Client" and the other one is called "*Connection". Not great but at least different. I'm open for better names.

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -0,0 +1,228 @@
...
+}
+
+export interface PostTxResult {
+  readonly logs: readonly Log[];
+  readonly rawLog: string;
+  readonly transactionHash: string;
+}
+
+export class CosmWasmClient {
+  public static makeReadOnly(url: string): CosmWasmClient {
```
ethanfrey: I like this separation of readable and writable.

I assume the readable one has methods that would throw errors rather than a different interface.
Good for first version, maybe we can make `clas CosmWasmSigningClient extends CosmWasmClient` for write vs read.

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -0,0 +1,228 @@
...
+  public async getNonce(address?: string): Promise<GetNonceResult> {
+    const account = (await this.restClient.authAccounts(address || this.senderAddress)).result.value;
+    return {
+      accountNumber: account.account_number,
+      sequence: account.sequence,
+    };
+  }
+
+  public async postTx(tx: Uint8Array): Promise<PostTxResult> {
+    const result = await this.restClient.postTx(tx);
```
ethanfrey: nice simple processing of result

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -0,0 +1,228 @@
...
+    if (result.code) {
+      throw new Error(`Error when posting tx. Code: ${result.code}; Raw log: ${result.raw_log}`);
+    }
+    return {
+      logs: parseLogs(result.logs) || [],
+      rawLog: result.raw_log || "",
+      transactionHash: result.txhash,
```
ethanfrey: this is a hex string, I believe, how about parsing into Uint8Array?

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -0,0 +1,228 @@
...
+    return this.signingData.signCallback;
+  }
+
+  private constructor(url: string, signingData?: SigningData) {
+    this.restClient = new RestClient(url);
+    this.signingData = signingData;
+  }
+
+  public async chainId(): Promise<string> {
+    const response = await this.restClient.nodeInfo();
```
ethanfrey: We can cache this on first call, it is still a Promise, but one less rest call when we use the client for multiple txs

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -0,0 +1,228 @@
...
+      type: "wasm/store-code",
+      value: {
+        sender: this.senderAddress,
+        // eslint-disable-next-line @typescript-eslint/camelcase
+        wasm_byte_code: Encoding.toBase64(wasmCode),
+        source: "",
+        builder: "",
+      },
+    };
+    const fee = defaultUploadFee;
```
ethanfrey: This is a very nice, simple interface.
I would (in another PR) make the fees configurable in the constructor, like an Options argument, and use default for any that aren't set. So we get the same ease of use, but allow a one-time configuration in the constructor

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -0,0 +1,228 @@
...
+      msg: [storeCodeMsg],
+      fee: fee,
+      memo: memo,
+      signatures: [signature],
+    };
+
+    const result = await this.postTx(marshalTx(signedTx));
+    const codeIdAttr = findAttribute(result.logs, "message", "code_id");
+    const codeId = Number.parseInt(codeIdAttr.value, 10);
+    return codeId;
```
ethanfrey: :+1:

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -0,0 +1,228 @@
...
+        wasm_byte_code: Encoding.toBase64(wasmCode),
+        source: "",
+        builder: "",
+      },
+    };
+    const fee = defaultUploadFee;
+    const { accountNumber, sequence } = await this.getNonce();
+    const chainId = await this.chainId();
+    const signBytes = makeSignBytes([storeCodeMsg], fee, chainId, memo, accountNumber, sequence);
+    const signature = await this.signCallback(signBytes);
```
ethanfrey: I'm assuming a Pen fits in here? Will look at test code next to check.

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -0,0 +1,228 @@
...
+    const contractAddressAttr = findAttribute(result.logs, "message", "contract_address");
+    return contractAddressAttr.value;
+  }
+
+  public async execute(
+    contractAddress: string,
+    handleMsg: object,
+    memo = "",
+    transferAmount?: readonly Coin[],
+  ): Promise<{ readonly logs: readonly Log[] }> {
```
ethanfrey: Yeah more flexible than just `readonly Log[]`. One day I will figure out how to get the data value here

packages/sdk/src/testutils.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  const random = Math.floor(Math.random() * (max - min)) + min;
+  const bytes = leb128Encode(random);
+
+  data[0x00e67d + 1] = bytes[0];
+  data[0x00e67d + 2] = bytes[1];
+
+  return data;
+}
+
+export function makeRandomAddress(): string {
```
ethanfrey: Nice to pull this out. And this is totally easier than generating a real key

packages/sdk/src/cosmwasmclient.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+      const codeId = await client.upload(getRandomizedHackatom());
+      expect(codeId).toBeGreaterThanOrEqual(1);
+    });
+  });
+
+  describe("instantiate", () => {
+    it("works with transfer amount", async () => {
+      pendingWithoutCosmos();
+      const pen = await Secp256k1Pen.fromMnemonic(faucet.mnemonic);
+      const client = CosmWasmClient.makeWritable(httpUrl, faucet.address, async signBytes => {
```
ethanfrey: This is nice, but can we extend Pen to provide this closure?

eg. `makeWritable(httpUrl, facuet.address, pen.signer())` where

`signer(): SigningCallback`?

Or if that violates encapsulation, a helper like 
`CosmWasmClient.makeWritableFromPen(httpUrl, pen)`
and it can use `pen.pubkey` to generate the signing address as well

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -0,0 +1,228 @@
...
+import {
+  Coin,
+  MsgExecuteContract,
+  MsgInstantiateContract,
+  MsgStoreCode,
+  StdFee,
+  StdSignature,
+} from "./types";
+
+const defaultUploadFee: StdFee = {
```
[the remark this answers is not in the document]
webmaster128: Do you pay the fee for the gas limit or the gas used?

packages/sdk/src/cosmwasmclient.ts
webmaster128: "read only" might be misleading as it actually means _no signing support_. You can use a readonly client to post transactions signed elsewhere (which is what we do in the BCP implementation).

packages/sdk/src/cosmwasmclient.ts
webmaster128: Uppercase hex is the primary printable transaction ID in the Tendermint ecosystem, which is why I kept it a string. Maybe we can model that better by renaming `transactionHash` -> `transactionId`.

packages/sdk/src/cosmwasmclient.ts
webmaster128: I like how you think :)

packages/sdk/src/cosmwasmclient.ts
webmaster128: Ther idea is: The class does not care which tool you use to create the signature. Just conform to `SigningCallback`

packages/sdk/src/testutils.spec.ts
webmaster128: Note that the stuff in `testutils.spec.ts ` is inly available in tests. We can make it a runtime helper but I'd be careful with introducing unnecessary public symbols.

packages/sdk/src/cosmwasmclient.ts
webmaster128: Good point. I create the type `ExecuteResult` for easier extensibility in 90e76b7a41996c459a9665dda049d22918bc4565

packages/sdk/src/cosmwasmclient.spec.ts
webmaster128: Both ways make sense to me. I'll think about how to implement this feature without creating unwanted dependencies.

packages/sdk/src/cosmwasmclient.ts
webmaster128: 4c7124d12c9da060654ae1ab118451b705a8e745 adds format checks and format guarantees in the API. I think this is more convenient than exposing binary data.

webmaster128:
Thank you for thorough review @ethanfrey! Some of your points are very valid but I don't see super low hanging fruits right now. I'm sure there is much potential for improvement ahead. Will try to create some tickets. Merging this now as it clearly improves the situation and creates a framework for future changes.

packages/sdk/src/cosmwasmclient.ts
ethanfrey: Fair enough

ethanfrey:
Great.
This is a huge advance and I think we need to use it a bit in scripts and cli to determine the best next steps.

PR #79, opened 2020-02-11
Let Pen.sign return full StdSignature

### LINKED ISSUE #8: Add x/wasm tx types
There are three new tx types we should support from [x/wasmd](https://github.com/cosmwasm/wasmd/blob/master/x/wasm/internal/types/msg.go):

- [x] `MsgStoreCode`
- [x] `MsgInstantiateContract`
- [ ] `MsgExecuteContract`

Let's add typescript definitions for them and support in postTx

### LINKED ISSUE #78: Improve compatibility between Pen and CosmWasmClient
Right now, `CosmWasmClient` does not care how transactions are signed. Any callback in the form `(signBytes: Uint8Array): Promise<StdSignature>` can do the job. This is any async interface to even allow user interaction (e.g. typing a password). This flexibility is nice when replacing `Pen` with a production-ready key manager.

The goal of this ticket is to replace the 4 lines

```ts
      const pen = await Secp256k1Pen.fromMnemonic(faucet.mnemonic);
      const client = CosmWasmClient.makeWritable(httpUrl, faucet.address, async signBytes => {
        return encodeSecp256k1Signature(pen.pubkey, await pen.createSignature(signBytes));
      });
```

into two lines.

Since `CosmWasmClient` is supposed to be a production-ready class and `Secp256k1Pen` is just a dummy throw away signer, we probably should adapt the later to expose the required interface.

webmaster128:
Closes #78


packages/sdk/src/encoding.ts
```diff
@@ -59,12 +57,3 @@ export function makeSignBytes(
   const signMsg = sortJson(signJson);
   return toUtf8(JSON.stringify(signMsg));
 }
-
-export function encodeSecp256k1Signature(pubkey: Uint8Array, signature: Uint8Array): StdSignature {
```
webmaster128: This method (and tests) was just moves to module `./signature`

packages/sdk/src/signature.ts
```diff
@@ -0,0 +1,29 @@
...
+export function encodeSecp256k1Signature(pubkey: Uint8Array, signature: Uint8Array): StdSignature {
+  return {
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    pub_key: encodeSecp256k1Pubkey(pubkey),
+    // Recovery seems to be unused
+    signature: Encoding.toBase64(Secp256k1.trimRecoveryByte(signature)),
+  };
+}
+
+export function decodeSignature(
```
webmaster128: This one is new and supposed to be more or less the inverse of `encodeSecp256k1Signature`

packages/cli/README.md
```diff
@@ -65,7 +65,7 @@ const sendTokensMsg: types.MsgSend = {
 };
 
 const signBytes = makeSignBytes([sendTokensMsg], defaultFee, defaultNetworkId, memo, account_number, sequence);
-const signature = encodeSecp256k1Signature(pen.pubkey, await pen.createSignature(signBytes));
+const signature = await pen.sign(signBytes);
```
ethanfrey: One method and so much cleaner. Nice

packages/sdk/src/cosmwasmclient.spec.ts
```diff
@@ -113,9 +113,7 @@ describe("CosmWasmClient", () => {
     it("works", async () => {
       pendingWithoutCosmos();
       const pen = await Secp256k1Pen.fromMnemonic(faucet.mnemonic);
-      const client = CosmWasmClient.makeWritable(httpUrl, faucet.address, async signBytes => {
-        return encodeSecp256k1Signature(pen.pubkey, await pen.createSignature(signBytes));
-      });
+      const client = CosmWasmClient.makeWritable(httpUrl, faucet.address, signBytes => pen.sign(signBytes));
```
ethanfrey: Nice.

`..., pen.sign.bind(pen))` would also work but this looks cleaner

ethanfrey: And definitely simpler than the old way

packages/sdk/src/signature.ts
```diff
@@ -0,0 +1,29 @@
...
+  signature: StdSignature,
+): { readonly pubkey: Uint8Array; readonly signature: Uint8Array } {
+  switch (signature.pub_key.type) {
+    // Note: please don't add cases here without writing additional unit tests
```
ethanfrey: :+1:

packages/sdk/src/signature.ts
```diff
@@ -0,0 +1,29 @@
...
+    case pubkeyType.secp256k1:
+      return {
+        pubkey: Encoding.fromBase64(signature.pub_key.value),
```
ethanfrey: Minor point, but I want to bring it up as I will unlikely review this code again for a long while:

This decodes from base64 to uint8array. I am wondering about the other inverse...

I assume there is no way to reverse `Secp256k1.trimRecoveryByte(signature)`, and it doesn't matter?

`encodeSecp256k1Pubkey` moves to the compressed form. Do we want to uncompress on decode?

packages/sdk/src/cosmwasmclient.spec.ts
webmaster128: I hardly understand why `this` does not work for plain `pen.sign` and never understood the bind method. I tried to find a linter rule to disallow .bind but did not find any.

packages/cli/README.md
webmaster128: The simplification in the usage with the RestServer made me confident that this is the better interface in general and we're not improving one case while worsen another.

packages/sdk/src/signature.ts
webmaster128: Both very good points. Symmetry is such a great tool to verify the design of software.

I think everything that happens inside of @cosmwasm/sdk should be 1. compressed pubkeys and 2. fixed length 64 byte signatures (i.e. no recovery byte). That means all this normalization should happen when an external system (like BCP) wants to connect to the SDK. So the proper fix in my opinion is to remove both conversions and just ensure we receive the correct format.

webmaster128: Done in 41ede3605e4f5cc3b11d58be2f2e514a03c7af7d and a11b756ea63ac464db4a0c0d0e858fdc2d27e586

packages/sdk/src/cosmwasmclient.spec.ts
ethanfrey: I think cuz classes aren't really such by syntax sugar over an object with function members. The fact that you can't just pass `pen.sign` and have it call the method of pen is definitely one of the annoying gotchas of modern js

PR #81, opened 2020-02-11
Calculate Nonce from account number and sequence

webmaster128:
Those two commits are unrelated and easier to review individually.

### DIFF
--- a/packages/bcp/src/cosmwasmconnection.ts
+++ b/packages/bcp/src/cosmwasmconnection.ts
@@ -372,9 +372,12 @@ export class CosmWasmConnection implements BlockchainConnection {
 
     // tslint:disable-next-line: deprecation
     const accountForHeight = await this.restClient.authAccounts(senderAddress, response.height);
-    // this is technically not the proper nonce. maybe this causes issues for sig validation?
+    const accountNumber = accountForHeight.result.value.account_number;
+    // this is technically not the proper sequence. maybe this causes issues for sig validation?
     // leaving for now unless it causes issues
-    const sequence = (accountForHeight.result.value.sequence - 1) as Nonce;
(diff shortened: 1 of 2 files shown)


packages/bcp/src/cosmwasmconnection.ts
```diff
@@ -372,9 +372,12 @@ export class CosmWasmConnection implements BlockchainConnection {
 
     // tslint:disable-next-line: deprecation
     const accountForHeight = await this.restClient.authAccounts(senderAddress, response.height);
-    // this is technically not the proper nonce. maybe this causes issues for sig validation?
+    const accountNumber = accountForHeight.result.value.account_number;
```
ethanfrey: Nice. I didn't realize there was a query here already, totally makes sense to handle this properly if we have access to the info

packages/bcp/src/cosmwasmconnection.ts
```diff
@@ -372,9 +372,12 @@ export class CosmWasmConnection implements BlockchainConnection {
 
     // tslint:disable-next-line: deprecation
     const accountForHeight = await this.restClient.authAccounts(senderAddress, response.height);
```
ethanfrey: What is with the deprecation linter issue?

packages/bcp/src/cosmwasmconnection.ts
webmaster128: I still don't know how reliable the sequence is. But we don't do verification, so it is not critical. At least the code is more expressive now.

packages/bcp/src/cosmwasmconnection.ts
webmaster128: The field `this.restClient` is marked as depreated in order to encourace using the CosmWasmClient, which is missing a few features right now.

PR #82, opened 2020-02-12
Add CosmWasmClient.getIdentifier

webmaster128:
This replaces one direct usage of the deprecated field `this.restServer` in CosmWasmConnection with usage of CosmWasmClient.

### DIFF
--- a/packages/bcp/src/cosmwasmconnection.ts
+++ b/packages/bcp/src/cosmwasmconnection.ts
@@ -1,12 +1,5 @@
 /* eslint-disable @typescript-eslint/camelcase */
-import {
-  CosmosAddressBech32Prefix,
-  CosmWasmClient,
-  RestClient,
-  TxsResponse,
-  types,
-  unmarshalTx,
-} from "@cosmwasm/sdk";
+import { CosmosAddressBech32Prefix, CosmWasmClient, RestClient, TxsResponse, types } from "@cosmwasm/sdk";
 import {
   Account,
   AccountQuery,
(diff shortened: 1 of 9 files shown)


packages/sdk/src/cosmwasmclient.ts
```diff
@@ -106,6 +108,16 @@ export class CosmWasmClient {
     return response.node_info.network;
   }
 
+  /**
+   * Returns a 32 byte upper-case hex transaction hash (typically used as the transaction ID)
+   */
+  public async getIdentifier(tx: CosmosSdkTx): Promise<string> {
```
ethanfrey: Yes, better here

packages/bcp/src/cosmwasmconnection.ts
```diff
@@ -162,12 +156,14 @@ export class CosmWasmConnection implements BlockchainConnection {
...
-    const tx = unmarshalTx(signed);
-    // tslint:disable-next-line: deprecation
-    const bytes = await this.restClient.encodeTx(tx);
-    const hash = new Sha256(bytes).digest();
-    return toHex(hash).toUpperCase() as TransactionId;
+  /**
+   * This is a replacement for the unimplemented CosmWasmCodec.identifier. Here we have more
+   * context and network available, which we might use to implement the API in an async way.
+   */
+  public async identifier(signed: SignedTransaction): Promise<TransactionId> {
```
ethanfrey: And nicer to use the signed transaction argument here, fits in better

PR #83, opened 2020-02-12
Improve search implementation

webmaster128:
Add more clarity, testing and search by recipient support

### DIFF
--- a/packages/bcp/src/cosmwasmconnection.ts
+++ b/packages/bcp/src/cosmwasmconnection.ts
@@ -288,8 +288,14 @@ export class CosmWasmConnection implements BlockchainConnection {
     signedBy,
     tags,
   }: TransactionQuery): Promise<readonly (ConfirmedTransaction<UnsignedTransaction> | FailedTransaction)[]> {
-    if ([maxHeight, minHeight, signedBy, tags].some(component => component !== undefined)) {
-      throw new Error("Transaction query by maxHeight, minHeight, signedBy or tags not yet supported");
+    if ([signedBy, tags].some(component => component !== undefined)) {
+      throw new Error("Transaction query by signedBy or tags not yet supported");
+    }
+
+    if ([maxHeight, minHeight].some(component => component !== undefined)) {
+      throw new Error(
+        "Transaction query by minHeight/maxHeight not yet supported. This is due to missing flexibility of the Gaia REST API, see https://github.com/cosmos/gaia/issues/75",
+      );
     }
 
     if ([id, height, sentFromOrTo].filter(component => component !== undefined).length !== 1) {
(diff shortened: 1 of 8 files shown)


packages/bcp/src/cosmwasmconnection.ts
```diff
@@ -304,15 +283,37 @@ export class CosmWasmConnection implements BlockchainConnection {
...
+    signedBy,
+    tags,
+  }: TransactionQuery): Promise<readonly (ConfirmedTransaction<UnsignedTransaction> | FailedTransaction)[]> {
+    if ([maxHeight, minHeight, signedBy, tags].some(component => component !== undefined)) {
+      throw new Error("Transaction query by maxHeight, minHeight, signedBy or tags not yet supported");
+    }
+
+    if ([id, height, sentFromOrTo].filter(component => component !== undefined).length !== 1) {
+      throw new Error(
+        "Transaction query by id, height and sentFromOrTo is mutually exclusive. Exactly one must be set.",
```
ethanfrey: Height can be combined with others. Id and sentfromorto are indeed exclusive.

At least in wasmd node, I am not sure about in the cosmwasmclient

packages/sdk/src/cosmwasmclient.spec.ts
```diff
@@ -109,6 +111,134 @@ describe("CosmWasmClient", () => {
...
+      expect(result[0]).toEqual(
+        jasmine.objectContaining({
+          height: posted.height.toString(),
+          txhash: posted.hash,
+          tx: posted.tx,
+        }),
+      );
+    });
+
+    it("can search by sender", async () => {
```
ethanfrey: These are good cases.

I would add support for height (or min/max height) along with the sender/recipient queries.

It's not critical but I figure if you are fixing up the search implementation, it fits in.

Reason. I last queried at block 200000 and have all tx before then. I want to query for new TX, and don't want the whole list again. I can query for {sender:me, minHeight:200001}

webmaster128: Unfortunately, min/max height does not work via the REST API (https://github.com/cosmos/gaia/issues/75). I made this information available in the exception text in 18aa7e33dddab20e870be18ad1a50e7e0db6afba.

I not try if height plus sender/recipient works. I guess it does. However, it would make the API much more complex and I don't see a usecase for that.

packages/bcp/src/cosmwasmconnection.ts
webmaster128: CosmWasmClient defines distinct search cases (wich BCP should have as well) to better allow for proper implementation of specific use cases:

```ts
export interface SearchByIdQuery {
  readonly id: string;
}
export interface SearchByHeightQuery {
  readonly height: number;
}
export interface SearchBySentFromOrToQuery {
  readonly sentFromOrTo: string;
}
export declare type SearchTxQuery = SearchByIdQuery | SearchByHeightQuery | SearchBySentFromOrToQuery;
```

We can add the ability to combine height (not minHeight/maxHeight) with other queries but I don't see the use case for that right now.

PR #84, opened 2020-02-12
Don't query nonces in searchTx

### LINKED ISSUE #80: CosmWasmConnection.searchTx: don't query nonces
In contrast to getTx, we don't need signed transactions for searchTx:

```ts
export interface BlockchainConnection {
  // ...

  readonly getTx: (
    id: TransactionId,
  ) => Promise<ConfirmedAndSignedTransaction<UnsignedTransaction> | FailedTransaction>;

  // ...

  readonly searchTx: (
    query: TransactionQuery,
  ) => Promise<readonly (ConfirmedTransaction<UnsignedTransaction> | FailedTransaction)[]>;
}
```

This is on purpose to allow scraper backends for loading long lists while at the same time allowing the application to get all information for a single transaction on demand.

If we refactor the code a bit, we don't need to query `const accountForHeight = await this.restClient.authAccounts(senderAddress, response.height);` for every search result to get the nonce.

webmaster128:
Closes #80

### DIFF
--- a/packages/bcp/src/cosmwasmconnection.ts
+++ b/packages/bcp/src/cosmwasmconnection.ts
@@ -368,21 +368,22 @@ export class CosmWasmConnection implements BlockchainConnection {
     const firstMsg = response.tx.value.msg.find(() => true);
     if (!firstMsg) throw new Error("Got transaction without a first message. What is going on here?");
 
-    let senderAddress: string;
+    // needed to get the (account_number, sequence) for the primary signature
+    let primarySignerAddress: string;
     if (types.isMsgSend(firstMsg)) {
-      senderAddress = firstMsg.value.from_address;
+      primarySignerAddress = firstMsg.value.from_address;
     } else if (
       types.isMsgStoreCode(firstMsg) ||
       types.isMsgInstantiateContract(firstMsg) ||
       types.isMsgExecuteContract(firstMsg)
     ) {
-      senderAddress = firstMsg.value.sender;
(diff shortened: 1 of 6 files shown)


packages/bcp/src/cosmwasmconnection.ts
```diff
@@ -357,34 +355,42 @@ export class CosmWasmConnection implements BlockchainConnection {
     };
   }
 
-  private async parseAndPopulateTxResponse(
+  private parseAndPopulateTxResponseUnsigned(
```
ethanfrey: Nice, so if we don't care about the sigs, this becomes a LOT simpler

packages/bcp/src/cosmwasmconnection.ts
```diff
@@ -234,8 +234,7 @@ export class CosmWasmConnection implements BlockchainConnection {
     try {
       // tslint:disable-next-line: deprecation
       const response = await this.restClient.txsById(id);
-      const chainId = this.chainId();
-      return this.parseAndPopulateTxResponse(response, chainId);
+      return this.parseAndPopulateTxResponseSigned(response);
```
ethanfrey: Ahhh... so we only do the complex transform when getting exactly one tx, not for search.
This seems quite smart

webmaster128: Jap. This interface is like this because of the Ethereum implementation where you search via scraper and get a single tx via RPC.

PR #91, opened 2020-02-15
Add CosmWasmClient.getBlock

webmaster128:


packages/bcp/src/cosmwasmconnection.ts
```diff
@@ -130,9 +130,8 @@ export class CosmWasmConnection implements BlockchainConnection {
   }
 
   public async height(): Promise<number> {
-    // tslint:disable-next-line: deprecation
-    const { block } = await this.restClient.blocksLatest();
-    return block.header.height;
+    const { block } = await this.cosmWasmClient.getBlock();
```
ethanfrey: Note that `getBlock` can be a lot heavier than `getHeader`.

Or at least in some theoretical world, when blocks have many transactions (or if they have a code upload tx)

I think these don't include the signatures (which are HUGE on cosmos hub.... something like 10kB per block), but that is only from the `/commit` endpoint.

In any case, I prefer to getHeader (a small, constant-size return value) when we only need info from the header. And only `getBlock` when we actually want the transactions.

packages/sdk/src/restclient.ts
```diff
@@ -22,25 +22,47 @@ interface NodeInfoResponse {
...
-interface BlockMeta {
-  readonly header: {
-    readonly height: number;
-    readonly time: string;
-    readonly num_txs: number;
-  };
-  readonly block_id: {
-    readonly hash: string;
-  };
+export interface BlockHeader {
```
ethanfrey: Adding full info is good

ethanfrey: It seems like this changed since cosmos hub (gaia 2)?

It would be nice to make sure the test hit all the fields to make sure they are there (and quickly break if we upgrade to an incompatible version)

packages/bcp/src/cosmwasmconnection.ts
webmaster128: I was thinking about this too but our RestClient does not have a function to get headers. And it wurns out, the API does not have an enpoint for headers: https://cosmos.network/rpc/#/

packages/sdk/src/restclient.ts
webmaster128: Yeah seems like that. I added tests for all the fields in the types.

packages/bcp/src/cosmwasmconnection.ts
ethanfrey: Oh man this RestServer misses a lot from raw tendermint..

Good find there, this seems like the best solution, until the refactor everything with protobuf and we can connect directly to tendermint

webmaster128: API endpoint [GET /validatorsets/latest (Get the latest validator set)](https://cosmos.network/rpc/#/Tendermint_RPC/get_validatorsets_latest) would be an alternative to get the hight. It might be smaller, depending in the number of validators and transactions in blocks.

PR #93, opened 2020-02-16
Test block responses at RestClient level

webmaster128:
Testing this functionality at CosmWasmClient level does not make much sense, since it does not test the client but the REST API. This duplicates those tests for the lower layer and adds the mission fields.

### DIFF
--- a/packages/sdk/src/restclient.spec.ts
+++ b/packages/sdk/src/restclient.spec.ts
@@ -2,6 +2,7 @@
 import { Sha256 } from "@iov/crypto";
 import { Encoding } from "@iov/encoding";
 import { assert } from "@iov/utils";
+import { ReadonlyDate } from "readonly-date";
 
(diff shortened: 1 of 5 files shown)


packages/sdk/src/restclient.spec.ts
```diff
@@ -179,6 +186,73 @@ describe("RestClient", () => {
     });
   });
 
+  describe("blocksLatest", () => {
+    it("works", async () => {
```
ethanfrey: Nice test coverage

Any idea the meaning of those version fields?
I guess we could check them (at least block==10) in the RestClient and then raise an error/warning if the block height is different (meaning different header format)

webmaster128:
Here you find one more version (`p2p`): https://github.com/confio/cosmwasm-js/issues/92 Looks like integer versioning. I have no idea how this works. We can read and analyze it later once we understand it.

PR #94, opened 2020-02-16
Add getAccount to CosmWasmClient

ethanfrey:
This is only available on the RestClient now, which was a bit annoying when writing the frontend app. Expose it in CosmWasmClient with a cleaner return value (undefined for no such account)

### DIFF
--- a/packages/sdk/src/cosmwasmclient.spec.ts
+++ b/packages/sdk/src/cosmwasmclient.spec.ts
@@ -74,14 +74,14 @@ describe("CosmWasmClient", () => {
       });
     });
 
-    it("returns zeros for missing accounts", async () => {
+    it("throws for missing accounts", async () => {
       pendingWithoutCosmos();
       const client = CosmWasmClient.makeReadOnly(httpUrl);
       const missing = makeRandomAddress();
-      expect(await client.getNonce(missing)).toEqual({
-        accountNumber: 0,
-        sequence: 0,
-      });
+      await client.getNonce(missing).then(
+        () => fail("this must not succeed"),
+        error => expect(error).toMatch(/account does not exist on chain/i),
+      );
     });
   });
 
--- a/packages/sdk/src/cosmwasmclient.spec.ts
+++ b/packages/sdk/src/cosmwasmclient.spec.ts
@@ -74,7 +74,15 @@ describe("CosmWasmClient", () => {
       });
     });
 
-
+    it("returns zeros for missing accounts", async () => {
+      pendingWithoutCosmos();
+      const client = CosmWasmClient.makeReadOnly(httpUrl);
+      const missing = makeRandomAddress();
+      expect(await client.getNonce(missing)).toEqual({
+        accountNumber: 0,
+        sequence: 0,
+      });
+    });
   });
 
   describe("getAccount", () => {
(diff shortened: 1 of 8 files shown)


packages/sdk/src/cosmwasmclient.spec.ts
```diff
@@ -73,6 +73,40 @@ describe("CosmWasmClient", () => {
         sequence: 0,
       });
     });
+
+    it("returns zeros for missing accounts", async () => {
+      pendingWithoutCosmos();
+      const client = CosmWasmClient.makeReadOnly(httpUrl);
+      const missing = makeRandomAddress();
+      expect(await client.getNonce(missing)).toEqual({
+        accountNumber: 0,
```
webmaster128: This is a strange behaviour. `accountNumber` is an on-chain value for the index of the account, right? So `accountNumber: 0` is of no use and even misleading. I think we should return `undefined` in this case

packages/sdk/src/cosmwasmclient.spec.ts
```diff
@@ -73,6 +73,40 @@ describe("CosmWasmClient", () => {
...
+        ],
+      });
+    });
+
+    it("returns undefined for missing accounts", async () => {
+      pendingWithoutCosmos();
+      const client = CosmWasmClient.makeReadOnly(httpUrl);
+      const missing = makeRandomAddress();
+      expect(await client.getAccount(missing)).toBeUndefined();
+    });
```
webmaster128: Nice tests

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -150,13 +151,19 @@ export class CosmWasmClient {
...
+    const account = await this.getAccount(address);
     return {
-      accountNumber: account.account_number,
-      sequence: account.sequence,
+      accountNumber: account ? account.account_number : 0,
+      sequence: account ? account.sequence : 0,
     };
   }
 
+  public async getAccount(address?: string): Promise<CosmosSdkAccount | undefined> {
```
webmaster128: At some point I'd like to have custom return types for all the CosmWasmClient methods instead of returning the REST response directly (as it was done for `GetNonceResult`). But this is good for now. For blocks we don't have custom types as well.

packages/sdk/src/cosmwasmclient.spec.ts
webmaster128: If we want to avoid the undefined checks for the happy path, we should throw an error if the account does not exist.

ethanfrey: I made it undefined first, and it broke like 10 places, so I wondered about that. Throwing an error makes sense, as we only get the nonce when preparing to sign, and you cannot sign with a missing account

ethanfrey:
Ah, saw you did that in https://github.com/confio/cosmwasm-js/pull/94/commits/28ee35217e9b6fa0cc4eae6968a5264220add09c

Very nice

packages/sdk/src/cosmwasmclient.spec.ts
webmaster128: I agree. Changed it to throw instead of returning invalid data.

PR #95, opened 2020-02-16
Expose more types

ethanfrey:
A few types (and one method) that can be safely exposed and would be nice to use in the frontend.


webmaster128:
Makes sense overall. However, at this point I think we should have separate types for signing and non-signing clients

PR #96, opened 2020-02-16
Expose sendToken() on CosmWasmClient

ethanfrey:
Based on #94 (merge that first)

This let's us easily make token transfers with this API, as we sometimes need to move some cash before we can call contracts.

### DIFF
--- a/packages/sdk/src/cosmwasmclient.spec.ts
+++ b/packages/sdk/src/cosmwasmclient.spec.ts
@@ -222,50 +222,23 @@ describe("CosmWasmClient", () => {
     beforeAll(async () => {
       if (cosmosEnabled()) {
         const pen = await Secp256k1Pen.fromMnemonic(faucet.mnemonic);
-        const client = CosmWasmClient.makeReadOnly(httpUrl);
+        const client = CosmWasmClient.makeWritable(httpUrl, faucet.address, signBytes => pen.sign(signBytes));
 
         const memo = "My first contract on chain";
-        const sendMsg: MsgSend = {
-          type: "cosmos-sdk/MsgSend",
-          value: {
-            from_address: faucet.address,
-            to_address: makeRandomAddress(),
-            amount: [
-              {
-                denom: "ucosm",
-                amount: "1234567",
-              },
-            ],
+        const recipient = makeRandomAddress();
+        const transferAmount = [
+          {
+            denom: "ucosm",
+            amount: "1234567",
           },
-        };
+        ];
+        const result = await client.sendToken(recipient, transferAmount, memo);
 
-        const fee: StdFee = {
-          amount: [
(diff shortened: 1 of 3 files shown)


packages/sdk/src/cosmwasmclient.spec.ts
```diff
@@ -412,6 +419,37 @@ describe("CosmWasmClient", () => {
...
+      const result = await client.sendToken(beneficiaryAddress, transferAmount, "for dinner");
+      const [firstLog] = result.logs;
+      expect(firstLog).toBeTruthy();
+
+      // got tokens
+      const after = await client.getAccount(beneficiaryAddress);
+      assert(after);
+      expect(after.coins).toEqual(transferAmount);
+    });
+  });
```
webmaster128: 👍

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -96,23 +98,29 @@ export interface ExecuteResult {
 }
 
 export class CosmWasmClient {
-  public static makeReadOnly(url: string): CosmWasmClient {
-    return new CosmWasmClient(url);
+  public static makeReadOnly(url: string, feeTable?: Partial<FeeTable>): CosmWasmClient {
```
webmaster128: Why does this constructor take the argument? A `CosmWasmClient` from here cannot sign, which probably means it does not need fees.

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -124,9 +132,10 @@ export class CosmWasmClient {
     return this.signingData.signCallback;
   }
 
-  private constructor(url: string, signingData?: SigningData) {
+  private constructor(url: string, signingData?: SigningData, feeTable?: Partial<FeeTable>) {
```
webmaster128: You can make the `feeTable` non optional here. Since this constructor is private, only this file cares. Than the assignment expression below gets a little less crazy.

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -308,6 +323,36 @@ export class CosmWasmClient {
...
+    const signature = await this.signCallback(signBytes);
+    const signedTx = {
+      msg: [sendMsg],
+      fee: fee,
+      memo: memo,
+      signatures: [signature],
+    };
+
+    return this.postTx(marshalTx(signedTx));
+  }
```
webmaster128: Nice function. Fits nicely next to upload/instantiate/execute.

Can we rename this to plural `sendTokens`?

packages/sdk/src/cosmwasmclient.ts
ethanfrey: Good point

PR #97, opened 2020-02-17
Add read support for ERC20 executions

webmaster128:


packages/bcp/src/decode.spec.ts
```diff
@@ -140,7 +158,40 @@ describe("decode", () => {
...
         },
       };
-      expect(parseMsg(msg, defaultMemo, testdata.chainId, defaultTokens)).toEqual(defaultSendTransaction);
+      expect(parseMsg(msg, defaultMemo, testdata.chainId, defaultTokens, defaultErc20Tokens)).toEqual(
+        defaultSendTransaction,
+      );
+    });
+
+    it("works for ERC20 send transaction", () => {
+      const msg: types.MsgExecuteContract = {
```
ethanfrey: wow. this is very cool

packages/bcp/src/cosmwasmconnection.spec.ts
```diff
@@ -318,6 +319,56 @@ describe("CosmWasmConnection", () => {
       connection.disconnect();
     });
 
+    it("can get a recently posted ERC20 send transaction", async () => {
```
ethanfrey: :+1:

Issue #49, opened 2020-02-05
Add support for subscriptions


ethanfrey:
We should expose some sort of subscription listener to the RestClient and use it in the CosmWasmConnection. In particular, for full bcp-compatibility, we need to implement:

* `watchBlockHeaders`
* `watchAccount`

And `postTx` could be done nicer with a subscription that with a poll.

I believe these are needed for dropping in `CosmWasmConnection` into the iov-wallet

webmaster128:
What would also work is a hybrid, where you use a tendermint socker subscribtion to get notified about new data. When you receive a new transaction, you ignore the amino data and pull via REST.

ethanfrey:
Yes, I like this. I think we did some of this at iov. Just have a stream of "update happened" from the subscription service. And those events trigger a rest call to get the data.

I can look into proper tendermint subscription queries,  but that would also require a second url and connection (maybe optional?)

webmaster128:
For wallet integration we need `watchAccount` and `liveTx`. The first one is easy. The second one can be implemented [like Lisk](https://github.com/iov-one/iov-core/blob/v2.0.2/packages/iov-lisk/src/liskconnection.ts#L382-L414). This is horribly inefficient without min/max height search filters, but those [will be available soon](https://github.com/cosmos/cosmos-sdk/pull/5648).

webmaster128:
Closing since the BCP implementation is moving out of here and subscriptions will get their comeback in a couple of weeks when we successfully communicate with Tendermint.

PR #99, opened 2020-02-17
Implement CosmWasmConnection.watchAccount

webmaster128:
First part of #49

### DIFF
--- a/packages/bcp/src/cosmwasmconnection.spec.ts
+++ b/packages/bcp/src/cosmwasmconnection.spec.ts
@@ -1,7 +1,9 @@
 import { CosmosAddressBech32Prefix, decodeSignature } from "@cosmwasm/sdk";
 import {
+  Account,
   Address,
   Algorithm,
+  Amount,
   ChainId,
   isBlockInfoPending,
   isBlockInfoSucceeded,
@@ -17,6 +19,7 @@ import { Random, Secp256k1, Secp256k1Signature, Sha256 } from "@iov/crypto";
 import { Bech32, Encoding } from "@iov/encoding";
 import { HdPaths, Secp256k1HdWallet, UserProfile } from "@iov/keycontrol";
 import { assert } from "@iov/utils";
+import BN from "bn.js";
 
 import { CosmWasmConnection, TokenConfiguration } from "./cosmwasmconnection";
 import { encodeFullSignature } from "./encode";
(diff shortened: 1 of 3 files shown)


packages/bcp/src/cosmwasmconnection.spec.ts
```diff
@@ -267,6 +275,70 @@ describe("CosmWasmConnection", () => {
...
+      const events = new Array<Account | undefined>();
+
+      (async () => {
+        const connection = await CosmWasmConnection.establish(httpUrl, defaultPrefix, defaultConfig);
+        const subscription = connection.watchAccount({ address: recipient }).subscribe({
+          next: event => {
+            events.push(event);
+
+            if (events.length === 3) {
+              const [event1, event2, event3] = events;
```
ethanfrey: Looks solid test

packages/bcp/src/cosmwasmconnection.ts
```diff
@@ -185,8 +188,34 @@ export class CosmWasmConnection implements BlockchainConnection {
     }
   }
 
-  public watchAccount(_account: AccountQuery): Stream<Account | undefined> {
-    throw new Error("not implemented");
+  public watchAccount(query: AccountQuery): Stream<Account | undefined> {
```
ethanfrey: This is good.

I have some questions for when I try to integrate with the frontend, but not pressing.

Basically, if I create the stream and subscribe on when mounting a component, how do I close it properly on leaving? Looking at https://github.com/staltz/xstream#subscribe I would guess something like:

```js
const sub = client.watchAccount({address: address}).subscribe({ /* callbacks */});
// when done
sub.cancel() // ???? this is not on the README
```

ethanfrey: Also, this is only on bcp level.
I think it would be nice to have on sdk level as well (or done there and then the bcp is a wrapper around it).
I query balance in simple react app to put in the header, but it never updates unless you reload. This `watchAccount` function would be a perfect match to the `getAccount` just added to `@cosmwasm/sdk`

webmaster128: First point: right. And then there is `Subscription.unsubscribe()`.

Second point: All this streaming stuff barely works. It is a mess. You can choose between HTTP polling and websockets that break and don't reconnect. I think it is safe to call this road a failure. There is nothing the library can offer to the app which the app can't do on its own.

ethanfrey:
> Second point: All this streaming stuff barely works. It is a mess. You can choose between HTTP polling and websockets that break and don't reconnect. I think it is safe to call this road a failure.

Wow. That is harsh.

More and more single-page webapps are going the "reactive" route. This is what we want to provide as tooling. There must be a good way to do this. Or at least an established "good enough" practice. We should pull that into a new issue.

webmaster128:
Well, I'm a bit frustated that we put months of work into this functionality and at the end of the day end up with UX issues like https://github.com/iov-one/ponferrada/issues/964. It would be much more stable if the app did its reloads manually in some proper events (_user did x_ => _a reload makes sense_; _y time elapsed_ => _do a reload, why not_).

I am happy to discuss this separately. But I gave up the idea of magical data availablility for apps.

ethanfrey:
That's a tough one. I made an issue to discuss this whole topic about reactive apps. It is a topic for later, but yeah, good to build on experience and not depend on flaky stuff

PR #103, opened 2020-02-18
Unpad integers r, s in Secp256k1Signature constructor

### LINKED ISSUE #100: Fix exception "Unsigned integer r must be encoded as unpadded big endian."
In

```
  const { pubkey, signature } = decodeSignature(firstSignature);
  const secp256keSignature = new Secp256k1Signature(signature.slice(0, 32), signature.slice(32, 64));
```

the Secp256k1Signature constructor throws "Unsigned integer r must be encoded as unpadded big endian." when r (or s) is padded, i.e. the integer is smaller than 32 bytes but a fixed-length encoding is given.

webmaster128:
Closes #100

### DIFF
--- a/packages/sdk/src/signature.ts
+++ b/packages/sdk/src/signature.ts
@@ -1,4 +1,6 @@
+import { Secp256k1Signature } from "@iov/crypto";
 import { Encoding } from "@iov/encoding";
+import BN from "bn.js";
 
 import { encodeSecp256k1Pubkey } from "./pubkey";
 import { pubkeyType, StdSignature } from "./types";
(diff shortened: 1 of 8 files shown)


packages/sdk/src/signature.ts
```diff
@@ -37,3 +39,10 @@ export function decodeSignature(
       throw new Error("Unsupported pubkey type");
   }
 }
+
+// TODO: use Secp256k1Signature.fromFixedLength once this is published https://github.com/iov-one/iov-core/pull/1401
+export function makeSecp256k1SignatureFromFixedLength(signature: Uint8Array): Secp256k1Signature {
+  const unpaddedR = Uint8Array.from(new BN(signature.slice(0, 32)).toArray());
```
ethanfrey: What difference is there from `r = signature.slice(0, 32)`.
Passing the slice into BN and out does some normalization I supposed.
It would be nice to comment why (as this seems to be needed for correctness)

ethanfrey: An no tests for a failure case (which would prevent someone from "simplifying" this is a future)

ethanfrey: Ah, looking at #100 and
`Secp256k1Signature constructor throws "Unsigned integer r must be encoded as unpadded big endian." when r (or s) is padded, i.e. the integer is smaller than 32 bytes but a fixed-length encoding is given.`

I assume this happens when there are leading zeros.
Oof... go to brute force 256 of them to find a failure.

webmaster128: Exactly, r and s are integers that sometimes have leading zero(s). They are not uniformly distributed but a result of an ellitic curve operation (which I don't understand). I don't know the odds for leading zeros. The `Secp256k1Signature` constructor that disallows leading zeros comes from the Ethereum implementation, where r and s are DER encoded, which uses variant length integers.

Since this was an invalid usage of the API and IOV Core 2.1.0 will bring a direct, safe, tested API for us, I don't want to put much effort into this temporary helper function.

PR #104, opened 2020-02-18
Update test setup

webmaster128:
This is done to avoid conflicts with other BCP demo tokens (ASH, BASH, CASH, MESH …) to simplify multichain demos.


webmaster128:
> interesting names

Well, it is hard to come up with new names all the time. The names are basically random but alphabetically ordered (H, I, J).

PR #105, opened 2020-02-18
Bump ts target to es2017

webmaster128:
This gives us smaller, faster and easier-to-read JS output, specially using native `async`/`await`.

Same as we did in IOV Core recently: https://github.com/iov-one/iov-core/pull/1386


ethanfrey:
Nice.
As we are only supporting modern browers, this is very safe upgrade

PR #111, opened 2020-02-19
Upload gzipped contract (nicer when not localhost)

ethanfrey:
Makes deploy a bit faster/stabler (and cheaper!)


webmaster128:
The library should do that for you: https://github.com/confio/cosmwasm-js/issues/113

ethanfrey:
Wasmd looks at the first 3-4 bytes in the binary (magic bytes) and requires either gzip or wasm before possibly unzipping and passing through

ethanfrey:
Closing in favor of#114

PR #112, opened 2020-02-19
Add some useful restclient queries to cosmwasmclient

ethanfrey:
Maybe we should expose them all?
This was nice for some quick sanity checks, we can extend it if you agree.


webmaster128:
If there is no long term production ready usage of those getters, I'd void adding them to the client and use them via `RestClient` directly.

I see some use in `listCodeInfo`, which lists all existing contracts, right? But `listContractAddresses` without a code ID filter does not seem very useful in the long run.

webmaster128:
Thank you for your input! This has been implemented as much as possible in #121 and will be extended in https://github.com/confio/cosmwasm-js/issues/123. Closing this now.

PR #114, opened 2020-02-19
Auto-compress contracts in upload

### LINKED ISSUE #113: Implement gzip compression in SigningCosmWasmClient.upload
See
- https://github.com/cosmwasm/wasmd/pull/9
- https://github.com/cosmwasm/wasmd/issues/18

webmaster128:
Closes #111
Closes #113

### DIFF
--- a/packages/sdk/src/signingcosmwasmclient.ts
+++ b/packages/sdk/src/signingcosmwasmclient.ts
@@ -1,4 +1,6 @@
+import { Sha256 } from "@iov/crypto";
 import { Encoding } from "@iov/encoding";
+import pako from "pako";
 
 import { CosmWasmClient, GetNonceResult, PostTxResult } from "./cosmwasmclient";
 import { makeSignBytes, marshalTx } from "./encoding";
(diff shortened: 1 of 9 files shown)


webmaster128:
The deployment script now logs


> Waiting for blockchain and REST server to be available ...
Okay, thank you for your patience.
Upload succeeded. Receipt: {"originalSize":84463,"originalChecksum":"b26861a6aa9858585ed905a590272735bd4fe8177c708940236224e8c9ff73ca","compressedSize":29494,"compressedChecksum":"2e4562cdcab7c10a0ff44051b4e126687438a12710591cecf0b9002c886c8cda","codeId":1}
Contract instantiated for HASH at cosmos18vd8fpwxzck93qlwghaj6arh4p7c5n89uzcee5
Contract instantiated for ISA at cosmos1hqrdl6wstt8qzshwc6mrumpjk9338k0lr4dqxd
Contract instantiated for JADE at cosmos18r5szma8hm93pvx6lwpjwyxruw27e0k5uw835c
All done, let the coins flow.

packages/sdk/src/signingcosmwasmclient.ts
```diff
@@ -80,14 +95,15 @@ export class SigningCosmWasmClient extends CosmWasmClient {
     return super.getAccount(address || this.senderAddress);
   }
 
-  /** Uploads code and returns a code ID */
-  public async upload(wasmCode: Uint8Array, memo = ""): Promise<number> {
+  /** Uploads code and returns a receipt, including the code ID */
+  public async upload(wasmCode: Uint8Array, memo = ""): Promise<UploadReceipt> {
+    const compressed = pako.gzip(wasmCode, { level: 9 });
```
ethanfrey: This is good. We can merge like this, but I would check the magic byte first, so this works for both wasm and gzip wasm.

I will link the go code. But that can be a follow up pr

webmaster128: > so this works for both wasm and gzip wasm.

Why would the user ever want to create the gzip version manually? There are already multiple different representations of the contract that the user needs to handle (source, build, optimized built), and I suggest to leave the compression up to clients like CosmWasmClient and wasm-cli. As a consequence `wasmCode` is always the original wasm.

PR #120, opened 2020-02-21
Add nameservice scripts

ethanfrey:
Add some sample startup scripts to register name service contracts, just like the erc20 one.

This is meant to help bootstrap for my react app, but should be generally useful. Maybe we can start pulling out some helper functions now that we have two copies, but really, there is not too much boilerplate.


scripts/wasmd/deploy_nameservice.js
```diff
@@ -0,0 +1,59 @@
...
+  for (const initMsg of [initFree, initLuxury]) {
+    const memo = `Create an ERC20 instance for ${initMsg.name}`;
+    const contractAddress = await client.instantiate(uploadReceipt.codeId, initMsg, memo);
+    console.info(`Contract instantiated for ${initMsg.name} at ${contractAddress}`);
+  }
+}
+
+main().then(
+  () => {
+    console.info("All done, let the coins flow.");
```
webmaster128: This text was specifically for the token contracts

scripts/wasmd/deploy_nameservice.js
```diff
@@ -0,0 +1,59 @@
...
+async function main() {
+  const pen = await Secp256k1Pen.fromMnemonic(faucet.mnemonic);
+  const client = new SigningCosmWasmClient(httpUrl, faucet.address, signBytes => pen.sign(signBytes));
+
+  const wasm = fs.readFileSync(__dirname + "/contracts/cw-nameservice.wasm");
+  const uploadReceipt = await client.upload(wasm, "Upload Name Service contract");
+  console.info(`Upload succeeded. Receipt: ${JSON.stringify(uploadReceipt)}`);
+
+  for (const initMsg of [initFree, initLuxury]) {
+    const memo = `Create an ERC20 instance for ${initMsg.name}`;
```
webmaster128: This text should be updated

scripts/wasmd/deploy_nameservice.js
```diff
@@ -0,0 +1,59 @@
...
+};
+const guest = {
+  address: "cosmos17d0jcz59jf68g52vq38tuuncmwwjk42u6mcxej",
+};
+
+const initFree = {
+  name: "Free",
+};
+const initLuxury = {
+  name: "Luxury",
```
webmaster128: I wonder if we should have instance titles/labels for all types of contracts as standarized meta data

PR #121, opened 2020-02-21
Add code and contract getters to CosmWasmClient

webmaster128:
This is an alternative to #112, with some differences:
* use our own types `Code`, `CodeDetails`, `Contract` exposed from @cosmwasm/sdk that don't need to match the REST response directly
* Prepare for https://github.com/cosmwasm/wasmd/issues/75, avoiding the `/wasm/contract` getter, which has no long term usage
* Try to come up with the best possible names instead of mirroring REST backend or client names

### DIFF
--- a/packages/sdk/src/cosmwasmclient.ts
+++ b/packages/sdk/src/cosmwasmclient.ts
@@ -48,6 +48,30 @@ export interface SearchTxFilter {
   readonly maxHeight?: number;
 }
 
+export interface Code {
+  readonly id: number;
+  /** Bech32 account address */
+  readonly creator: string;
+  /** Hex-encoded sha256 hash of the code stored here */
+  readonly checksum: string;
+  readonly source?: string;
+  readonly builder?: string;
+}
+
+export interface CodeDetails {
+  /** The original wasm bytes */
+  readonly wasm: Uint8Array;
+}
+
+export interface Contract {
+  // TODO: add contract address (https://github.com/cosmwasm/wasmd/issues/75)
+  readonly codeId: number;
+  /** Bech32 account address */
+  readonly creator: string;
+  /** Argument passed on initialization of the contract */
+  readonly initMsg: object;
+}
+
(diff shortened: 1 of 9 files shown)


packages/sdk/src/cosmwasmclient.ts
```diff
@@ -166,6 +190,42 @@ export class CosmWasmClient {
...
+  }
+
+  public async getCodeDetails(codeId: number): Promise<CodeDetails> {
+    const result = await this.restClient.getCode(codeId);
+    return {
+      wasm: result,
+    };
+  }
+
+  public async getContracts(codeId: number): Promise<readonly Contract[]> {
```
ethanfrey: Fine for current rest api, but we do want to add the contract address to all of these soon. I guess it is blocked on wasmd changes, but very important for usa ility.

For the name app, I added address to the api, pulling the argument from get contact and just brute force downloading all contracts and filtering locally for get contracts. This gives me the proper api now, and I can clean up the implementation once wasmd is enhanced.

webmaster128: Sounds like a good approach and I agree we want the address here ASAP (make a comment in the `Contract` interface already). Can you show me that code? I don't know how address generation works for contract instances and find it hard to follow your explanation.

webmaster128:
> Also make the contract generic T over init_msg, default object, as the caller often has more info.

When I worked on this, I thought `init_msg` should not be part of the list at all (see https://github.com/cosmwasm/wasmd/issues/80)

packages/sdk/src/cosmwasmclient.ts
webmaster128: I found the code/implementation in name-app you're refering to. This is .... wow, creative. And it does the job 🎩. But even in our small short living testing world this would easily send 100 requests to get the addresses of a single `codeId`. If there is no very good reason we need the interface now, I'd wait until the backend is ready.

PR #124, opened 2020-02-27
Test wasmd 0.7

ethanfrey:
Setting up for running tests against pre-release 0.7


webmaster128:
Please avoid any kind of rebases in this branch, since it includes release tags (of alpha versions)

ethanfrey:
We can always make another small pr if we find anything between rc1 alpha2 and full release.

PR #126, opened 2020-02-27
A lot of 0.7.0 adaptions

webmaster128:
Tests might pass with this

### DIFF
--- a/packages/sdk/src/restclient.spec.ts
+++ b/packages/sdk/src/restclient.spec.ts
@@ -12,6 +12,7 @@ import { PostTxsResponse, RestClient, TxsResponse } from "./restclient";
 import { SigningCosmWasmClient } from "./signingcosmwasmclient";
 import cosmoshub from "./testdata/cosmoshub.json";
 import {
+  bech32AddressMatcher,
   getRandomizedHackatom,
   makeRandomAddress,
   pendingWithoutWasmd,
@@ -670,7 +671,12 @@ describe("RestClient", () => {
 
       // create new instance and compare before and after
       const existingContractsByCode = await client.listContractsByCodeId(codeId);
-      existingContractsByCode.forEach(ctc => expect(ctc.code_id).toEqual(codeId));
+      for (const contract of existingContractsByCode) {
+        expect(contract.address).toMatch(bech32AddressMatcher);
+        expect(contract.code_id).toEqual(codeId);
+        expect(contract.creator).toMatch(bech32AddressMatcher);
+        expect(contract.label).toMatch(/^.+$/);
+      }
 
       const result = await instantiateContract(client, pen, codeId, beneficiaryAddress, transferAmount);
       expect(result.code).toBeFalsy();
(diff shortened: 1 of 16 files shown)


webmaster128:
🎉

packages/sdk/src/restclient.ts
```diff
@@ -362,11 +358,26 @@ export class RestClient {
     return unwrapWasmResponse(responseData) || [];
   }
 
-  // throws error if no contract at this address
-  public async getContractInfo(address: string): Promise<ContractDetails> {
+  /**
+   * Returns null when contract was not found at this address.
+   */
```
ethanfrey: If this is desired behavior, i can chango the server to return null not error. Please make an issue on wasmd

packages/sdk/src/signingcosmwasmclient.spec.ts
```diff
@@ -143,9 +143,14 @@ describe("SigningCosmWasmClient", () => {
       );
 
       // execute
-      const result = await client.execute(contractAddress, {release:{}}, undefined);
-      const [firstLog] = result.logs;
-      expect(firstLog.log).toEqual(`released funds to ${beneficiaryAddress}`);
+      const result = await client.execute(contractAddress, { release: {} }, undefined);
+      const wasmEvent = result.logs.find(() => true)?.events.find(e => e.type === "wasm");
+      assert(wasmEvent, "Event of type wasm expected");
+      expect(wasmEvent.attributes).toContain({ key: "action", value: "release" });
```
ethanfrey: Nice

packages/sdk/src/testutils.spec.ts
```diff
@@ -27,28 +27,27 @@ export function leb128Encode(uint: number): Uint8Array {
...
 
   // The return value of the export function cosmwasm_api_0_6 is unused and
   // can be randomized for testing.
   //
   // Find position of mutable bytes as follows:
   // $ wasm-objdump -d contract.wasm | grep -F "cosmwasm_api_0_6" -A 1
-  // 00e67c func[149] <cosmwasm_api_0_6>:
-  // 00e67d: 41 83 0c                   | i32.const 1539
+  // 0136d2 func[198] <cosmwasm_api_0_6>:
+  //  0136d3: 41 83 0c                   | i32.const 1539
```
ethanfrey: Magic here

packages/sdk/src/restclient.spec.ts
```diff
@@ -581,9 +582,15 @@ describe("RestClient", () => {
       {
         const result = await executeContract(client, pen, contractAddress);
         expect(result.code).toBeFalsy();
-        // console.log("Raw log:", result.raw_log);
-        const [firstLog] = parseLogs(result.logs);
-        expect(firstLog.log).toEqual(`released funds to ${beneficiaryAddress}`);
+        // console.log("Raw log:", result.logs);
+        const logs = parseLogs(result.logs);
```
ethanfrey: I'd love some tests on transfer, especially when we send to the contract and contract sends to the beneficiary in the same transaction. I think it all ends up mixed up.

Something happens in the higher levels outside the module and good to check that, so I can try to adjust the server before final repease

packages/sdk/src/restclient.ts
webmaster128: The reason I want to handle the _not found_ case explicitely is to allow setting a nice custom error message in the `CosmWasmClient`, since I suspect this case to be relatively common.

I think the cleanest solution for a missing resource in `GET /wasm/contract/${address}` is to return HTTP 404 with whatever content. That would avoid the need to look into error texts alltogether. See https://github.com/cosmwasm/wasmd/issues/89

PR #129, opened 2020-03-01
Search transactions by events

webmaster128:
- [x] Basic tx query
- [x] Review `SearchBySentFromOrToQuery` implementation (`message.sender=${query.sentFromOrTo}` does not limit to bank module and probably includes things we don't want)
- [x] Allow search by tags in `CosmWasmClient.searchTx`
- [x] Use in BCP


packages/sdk/src/cosmwasmclient.ts
```diff
@@ -159,11 +175,16 @@ export class CosmWasmClient {
       }
     } else if (isSearchBySentFromOrToQuery(query)) {
       // We cannot get both in one request (see https://github.com/cosmos/gaia/issues/75)
-      const sent = await this.txsQuery(withFilters(`message.sender=${query.sentFromOrTo}`));
-      const received = await this.txsQuery(withFilters(`transfer.recipient=${query.sentFromOrTo}`));
+      const sentQuery = withFilters(`message.module=bank&message.sender=${query.sentFromOrTo}`);
+      const receivedQuery = withFilters(`transfer.recipient=${query.sentFromOrTo}`);
```
ethanfrey: Very good queries. I saw the comment above, and this is exactly what I would have suggested.

But then, I wonder, do we want to capture just `MsgSend`, or all events that move funds (like when a contract sends tokens, or if you send to a contract)? `receivedQuery` works in the second case (got from contract), but `sentQuery` only works on the "plain" sends. This should be somehow symetric.|

Suggestion:
* Either add `message.module=bank&` to the received query
* Or just use `transfer.sender` on the first query (which only works on our patched sdk, but I think this is fine). 

You can always make two sentQueries - the current one which always matches on basic ones even on upstream:cosmos-sdk, and the "enhanced" `transfer.sender` one, and then de-duplicate them as is already done (to not double flag a self-send)

packages/bcp/src/cosmwasmconnection.spec.ts
```diff
@@ -589,6 +589,124 @@ describe("CosmWasmConnection", () => {
...
+      const byIdResult = byIdResults[0];
+      expect(byIdResult.transactionId).toEqual(transactionId);
+      assert(isConfirmedTransaction(byIdResult), "Expected transaction to succeed");
+      assert(byIdResult.log, "Log must be available");
+      const [firstByIdlog] = JSON.parse(byIdResult.log);
+      expect(firstByIdlog.events.length).toEqual(2);
+      expect(firstByIdlog.events[0].type).toEqual("message");
+      expect(firstByIdlog.events[1].type).toEqual("wasm");
+      // wasm event attributes added by contract
+      expect(firstByIdlog.events[1].attributes).toContain({ key: "action", value: "transfer" });
```
ethanfrey: This is very nice, good naming in the erc20 contract

packages/bcp/src/cosmwasmconnection.ts
```diff
@@ -324,7 +343,35 @@ export class CosmWasmConnection implements BlockchainConnection {
     } else if (height) {
       txs = await this.cosmWasmClient.searchTx({ height: height }, filter);
     } else if (sentFromOrTo) {
-      txs = await this.cosmWasmClient.searchTx({ sentFromOrTo: sentFromOrTo }, filter);
+      const pendingRequests = new Array<Promise<readonly TxsResponse[]>>();
+      pendingRequests.push(this.cosmWasmClient.searchTx({ sentFromOrTo: sentFromOrTo }, filter));
+      for (const contract of this.erc20Tokens.map(token => token.contractAddress)) {
+        const searchBySender = [
```
ethanfrey: I like this generic interface for event/tag queries.

The query looks very much like the result, very symetrical and easy to understand

packages/sdk/src/cosmwasmclient.searchtx.spec.ts
```diff
@@ -0,0 +1,361 @@
...
+      pendingWithoutWasmd();
+      assert(postedSend, "value must be set in beforeAll()");
+      const client = new CosmWasmClient(wasmdEndpoint);
+      const results = await client.searchTx({ sentFromOrTo: postedSend.recipient });
+      expect(results.length).toBeGreaterThanOrEqual(1);
+
+      // Check basic structure of all results
+      for (const result of results) {
+        const msg = fromOneElementArray(result.tx.value.msg);
+        assert(isMsgSend(msg), `${result.txhash} (height ${result.height}) is not a bank send transaction`);
```
ethanfrey: Yeah, this is the question I posed below - do we want to only match MsgSend, or other transactions that also affect the balance?

I am fine with just matching MsgSend if this is what we want, but then adjust the queries to be only that

packages/sdk/src/cosmwasmclient.searchtx.spec.ts
```diff
@@ -0,0 +1,361 @@
...
+    });
+  });
+
+  describe("with SearchByTagsQuery", () => {
+    it("can search by transfer.recipient", async () => {
+      pendingWithoutWasmd();
+      assert(postedSend, "value must be set in beforeAll()");
+      const client = new CosmWasmClient(wasmdEndpoint);
+      const results = await client.searchTx({
+        tags: [{ key: "transfer.recipient", value: postedSend.recipient }],
```
ethanfrey: Nice

packages/sdk/src/cosmwasmclient.searchtx.spec.ts
```diff
@@ -0,0 +1,361 @@
...
+      expect(results[results.length - 1]).toEqual(
+        jasmine.objectContaining({
+          height: postedSend.height.toString(),
+          txhash: postedSend.hash,
+          tx: postedSend.tx,
+        }),
+      );
+    });
+
+    it("can search by message.contract_address", async () => {
```
ethanfrey: nice one to test/demo

packages/sdk/src/restclient.spec.ts
```diff
@@ -479,12 +477,114 @@ describe("RestClient", () => {
         expect(count).toEqual("0");
       }
     });
+
+    it("can query by tags (module + code_id)", async () => {
```
ethanfrey: Nice that this already worked in rest... just a test away

packages/sdk/src/cosmwasmclient.ts
webmaster128: I think for simplicity we should stick with bank transfers and follow approach number 1. Otherwise 🤯 when trying to cover everything with a simple "sentFromOrTo".

PR #130, opened 2020-03-02
Add and test isValidBuilder

webmaster128:
This is the same regexp as in wasmd. `isValidBuilder` is unsed for now but could be used to prevent signing broken data.

Two things I noted here:
1. The regexp itself does not have a length limit. We should add have a limit independent of this.
2. Names without origanizations are supported. My gut feeling is that non of our wools will get a `_` organization on dockerhub and we should disallow those to prevent plain `cosmwasm-opt:0.6.2` from working

### DIFF
--- a/packages/sdk/src/builder.spec.ts
+++ b/packages/sdk/src/builder.spec.ts
@@ -1,0 +1,64 @@
+import { isValidBuilder } from "./builder";
+
+describe("builder", () => {
+  describe("isValidBuilder", () => {
+    // Valid cases
+
+    it("returns true for simple examples", () => {
+      expect(isValidBuilder("myorg/super-optimizer:0.1.2")).toEqual(true);
+      expect(isValidBuilder("myorg/super-optimizer:42")).toEqual(true);
+    });
+
+    it("supports images with no organization", () => {
+      // from https://hub.docker.com/_/ubuntu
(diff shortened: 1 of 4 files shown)


packages/sdk/src/builder.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    it("returns true for tags with lower and upper chars", () => {
+      expect(isValidBuilder("myorg/super-optimizer:0.1.2-alpha")).toEqual(true);
+      expect(isValidBuilder("myorg/super-optimizer:0.1.2-Alpha")).toEqual(true);
+    });
+
+    it("allows very long images", () => {
+      // This is > 2 KiB of data
+      expect(
+        isValidBuilder(
```
ethanfrey: should we prevent this?

webmaster128: Not as part of the regexp, but I'd set a max length limit on the string.

webmaster128:
I suggest setting a max string length of 128 bytes and a min number of name components to 2 (i.e. enforece the organization part).

ethanfrey:
Okay, want to update and test this regexp with those changes (max length for entire name, right, not any parts?)

Then I will update wasmd with those changes as well (just open an issue)

PR #131, opened 2020-03-02
CosmWasmClient types

webmaster128:
Create REST API independent higher level types for all CosmWasmClient interfaces.

~Based on #129~

### DIFF
--- a/packages/sdk/src/cosmwasmclient.searchtx.spec.ts
+++ b/packages/sdk/src/cosmwasmclient.searchtx.spec.ts
@@ -92,8 +92,8 @@ describe("CosmWasmClient.searchTx", () => {
       expect(result.length).toEqual(1);
       expect(result[0]).toEqual(
         jasmine.objectContaining({
-          height: postedSend.height.toString(),
-          txhash: postedSend.hash,
+          height: postedSend.height,
+          hash: postedSend.hash,
           tx: postedSend.tx,
         }),
       );
@@ -144,8 +144,8 @@ describe("CosmWasmClient.searchTx", () => {
       expect(result.length).toEqual(1);
       expect(result[0]).toEqual(
         jasmine.objectContaining({
-          height: postedSend.height.toString(),
-          txhash: postedSend.hash,
+          height: postedSend.height,
+          hash: postedSend.hash,
           tx: postedSend.tx,
         }),
       );
@@ -163,7 +163,7 @@ describe("CosmWasmClient.searchTx", () => {
       // Check basic structure of all results
       for (const result of results) {
         const msg = fromOneElementArray(result.tx.value.msg);
-        assert(isMsgSend(msg), `${result.txhash} (height ${result.height}) is not a bank send transaction`);
+        assert(isMsgSend(msg), `${result.hash} (height ${result.height}) is not a bank send transaction`);
         expect(
           msg.value.to_address === postedSend.sender || msg.value.from_address == postedSend.sender,
         ).toEqual(true);
@@ -172,8 +172,8 @@ describe("CosmWasmClient.searchTx", () => {
       // Check details of most recent result
       expect(results[results.length - 1]).toEqual(
         jasmine.objectContaining({
-          height: postedSend.height.toString(),
-          txhash: postedSend.hash,
+          height: postedSend.height,
+          hash: postedSend.hash,
           tx: postedSend.tx,
         }),
       );
@@ -189,7 +189,7 @@ describe("CosmWasmClient.searchTx", () => {
       // Check basic structure of all results
       for (const result of results) {
         const msg = fromOneElementArray(result.tx.value.msg);
-        assert(isMsgSend(msg), `${result.txhash} (height ${result.height}) is not a bank send transaction`);
+        assert(isMsgSend(msg), `${result.hash} (height ${result.height}) is not a bank send transaction`);
         expect(
           msg.value.to_address === postedSend.recipient || msg.value.from_address == postedSend.recipient,
         ).toEqual(true);
@@ -198,8 +198,8 @@ describe("CosmWasmClient.searchTx", () => {
       // Check details of most recent result
       expect(results[results.length - 1]).toEqual(
         jasmine.objectContaining({
-          height: postedSend.height.toString(),
-          txhash: postedSend.hash,
+          height: postedSend.height,
+          hash: postedSend.hash,
           tx: postedSend.tx,
         }),
       );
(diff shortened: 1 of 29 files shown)


packages/sdk/src/cosmwasmclient.searchtx.spec.ts
```diff
@@ -92,8 +92,8 @@ describe("CosmWasmClient.searchTx", () => {
       expect(result.length).toEqual(1);
       expect(result[0]).toEqual(
         jasmine.objectContaining({
-          height: postedSend.height.toString(),
-          txhash: postedSend.hash,
+          height: postedSend.height,
```
ethanfrey: height as number, and rename. This definitely looks nicer

packages/sdk/src/cosmwasmclient.spec.ts
```diff
@@ -83,10 +83,10 @@ describe("CosmWasmClient", () => {
       const client = new CosmWasmClient(wasmdEndpoint);
       expect(await client.getAccount(unused.address)).toEqual({
         address: unused.address,
-        account_number: 5,
+        accountNumber: 5,
```
ethanfrey: No more eslint:camel-case exceptions :+1:

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -92,6 +102,39 @@ export interface ContractDetails extends Contract {
...
+  readonly height: number;
+  readonly hash: string;
+  readonly rawLog: string;
+  readonly logs: readonly Log[];
+  readonly tx: CosmosSdkTx;
+  /** The gas limit as set by the user */
+  readonly gasWanted?: number;
+  /** The gas used by the execution */
+  readonly gasUsed?: number;
+  readonly timestamp: string;
```
ethanfrey: Please doc what the string format is (I assume same as BlockHeader.time?) There is an issue when parsing in go where you must know if it is supposed to be second, millisecond, microsecond, or nanosecond precision....

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -92,6 +102,39 @@ export interface ContractDetails extends Contract {
   readonly initMsg: object;
 }
 
+/** A transaction that is indexed as part of the transaction history */
+export interface IndexedTx {
+  readonly height: number;
+  readonly hash: string;
```
ethanfrey: This is upper-case hex, right?

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -2,14 +2,24 @@ import { Sha256 } from "@iov/crypto";
...
 export interface GetNonceResult {
   readonly accountNumber: number;
   readonly sequence: number;
 }
 
+export interface Account {
+  /** Bech32 account address */
+  readonly address: string;
+  readonly balance: ReadonlyArray<Coin>;
+  /** Bech32 encoded pubkey */
```
ethanfrey: I find the pubkey object (algorithm, data) format easier to handle outside of this package. Unless we decide bech32 pubkey is the only public interface everywhere (like from the Pen), I would prefer to use the object version for consistency, and just use the bech32 internally.

ethanfrey: eg, if I want to check the algorithm type... do I want to grab a few bytes and check for the amino-encoding in some file? Would anyone but us two even get that idea?

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -129,31 +172,47 @@ export class CosmWasmClient {
...
-      return this.restClient.blocks(height);
-    } else {
-      return this.restClient.blocksLatest();
-    }
+  public async getBlock(height?: number): Promise<Block> {
+    const response =
+      height !== undefined ? await this.restClient.blocks(height) : await this.restClient.blocksLatest();
+
+    return {
+      id: response.block_id.hash,
```
ethanfrey: This is very much nicer

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -129,31 +172,47 @@ export class CosmWasmClient {
...
+  public async getAccount(address: string): Promise<Account | undefined> {
     const account = await this.restClient.authAccounts(address);
     const value = account.result.value;
-    return value.address === "" ? undefined : value;
+    return value.address === ""
+      ? undefined
+      : {
+          address: value.address,
+          balance: value.coins,
+          pubkey: value.public_key || undefined,
```
ethanfrey: Nice cleanup. I would consider converting the public_key from bech32 into an object (as mentioned above). I am pretty sure the parse code exists

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -313,6 +369,15 @@ export class CosmWasmClient {
         `Found more results on the backend than we can process currently. Results: ${result.total_count}, supported: ${limit}`,
       );
     }
-    return result.txs;
+    return result.txs.map(
+      (restItem): IndexedTx => ({
```
ethanfrey: Nice and much more standardized

packages/sdk/types/signingcosmwasmclient.d.ts
```diff
@@ -28,6 +28,16 @@ export interface UploadReceipt {
   readonly compressedChecksum: string;
   /** The ID of the code asigned by the chain */
   readonly codeId: number;
+  readonly logs: readonly Log[];
+  /** Transaction hash (might be used as transaction ID). Guaranteed to be non-empty upper-case hex */
+  readonly transactionHash: string;
+}
+export interface InstantiateResult {
```
ethanfrey: Nice to add more into here as well

packages/sdk/src/signingcosmwasmclient.ts
```diff
@@ -188,7 +200,11 @@ export class SigningCosmWasmClient extends CosmWasmClient {
 
     const result = await this.postTx(signedTx);
     const contractAddressAttr = findAttribute(result.logs, "message", "contract_address");
-    return contractAddressAttr.value;
+    return {
```
ethanfrey: Very cool

packages/sdk/src/signingcosmwasmclient.ts
```diff
@@ -150,6 +160,8 @@ export class SigningCosmWasmClient extends CosmWasmClient {
       compressedSize: compressed.length,
       compressedChecksum: Encoding.toHex(new Sha256(compressed).digest()),
       codeId: Number.parseInt(codeIdAttr.value, 10),
+      logs: result.logs,
```
ethanfrey: Looks like all results get normalized with logs and transactionHash. I like that

packages/sdk/src/cosmwasmclient.ts
webmaster128: Good point

packages/sdk/src/cosmwasmclient.ts
webmaster128: Jap, documented in 23f88564e5df7ec092b9caf76708cb26d1cd12bc

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -185,7 +187,7 @@ export class CosmWasmClient {
       : {
           address: value.address,
           balance: value.coins,
-          pubkey: value.public_key || undefined,
+          pubkey: value.public_key ? decodeBech32Pubkey(value.public_key) : undefined,
```
ethanfrey: :+1:

PR #137, opened 2020-03-06
Support any prefix in decodeBech32Pubkey/encodeBech32Pubkey

webmaster128:


packages/sdk/src/pubkey.ts
```diff
@@ -29,10 +22,7 @@ const pubkeyAminoPrefixSr25519 = Encoding.fromHex("0dfb1005");
 const pubkeyAminoPrefixLength = pubkeyAminoPrefixSecp256k1.length;
 
 export function decodeBech32Pubkey(bechEncoded: string): PubKey {
-  const { prefix, data } = Bech32.decode(bechEncoded);
-  if (!isCosmosPubkeyBech32Prefix(prefix)) {
-    throw new Error(`Invalid bech32 prefix. Must be one of ${validPubkeyPrefixes.join(", ")}.`);
-  }
+  const { data } = Bech32.decode(bechEncoded);
```
ethanfrey: That is fine for now.

For later, maybe we could somehow configure the expected prefix per chain? Or does it not matter at all?

webmaster128: The only place I know where pubkeys are bech32 encoded is this REST response:

```
>> await restClient.authAccounts("enigma1vch5dkz89a7hp3nn0ycrlr9y798nlk7qs0encd")
{
  height: '18144',
  result: {
    type: 'cosmos-sdk/Account',
    value: {
      address: 'enigma1vch5dkz89a7hp3nn0ycrlr9y798nlk7qs0encd',
      coins: [Array],
      public_key: 'enigmapub1addwnpepqw5k9p439nw0zpg2aundx4umwx4nw233z5prpjqjv5anl5grmnchzp2xwvv',
      account_number: 20,
      sequence: 1
    }
  }
}
```

In the previous Cosmos SDK version, this did not have a prefix at all and was just

```
{
  type: 'tendermint/PubKeySecp256k1',
  value: 'A6lihrEs3PEFCu8m01ebcas3KjEVAjDIEmU7P9ED3PFx'
}
```

https://github.com/cosmos/amino-js/blob/v0.7.0-alpha.1/src/types/crypto.ts#L2-L11

Given this always comes from chain and is not a user input, I think it is alright to just drop the unused prefix.

PR #139, opened 2020-03-07
Add remaining fields to NodeInfoResponse

### LINKED ISSUE #92: Add missing fields for node info
Right now we only have the network ID [in the type](https://github.com/confio/cosmwasm-js/blob/v0.0.4/packages/sdk/src/restclient.ts#L17-L23) and tests. There is much more:

```js
{
  node_info: {
    protocol_version: { p2p: '7', block: '10', app: '0' },
    id: '67289558533ad6df13f96bfb7c8ae44b457a919a',
    listen_addr: 'tcp://0.0.0.0:26656',
    network: 'testing',
    version: '0.33.0',
    channels: '4020212223303800',
    moniker: 'testing',
    other: { tx_index: 'on', rpc_address: 'tcp://0.0.0.0:26657' }
  },
  application_version: {
    name: 'wasm',
    server_name: 'wasmd',
    client_name: 'wasmcli',
    version: '0.6.0',
    commit: 'e1a2cc3337f785a9109e011570537502cdbd38e2',
    build_tags: 'netgo,ledger',
    go: 'go version go1.13.7 linux/amd64'
  }
}
```

I was looking for a node height, which is not available here. So this has no priority as far as I can see.

webmaster128:
Closes #92

### DIFF
--- a/packages/sdk/src/restclient.spec.ts
+++ b/packages/sdk/src/restclient.spec.ts
@@ -22,6 +22,7 @@ import {
   tendermintAddressMatcher,
   tendermintIdMatcher,
   tendermintOptionalIdMatcher,
+  tendermintShortHashMatcher,
   wasmdEnabled,
   wasmdEndpoint,
 } from "./testutils.spec";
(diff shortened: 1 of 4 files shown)


webmaster128:
> Now to see what we do with these fields in cosmwasmclient

🤷‍♂️

PR #140, opened 2020-03-07
Add efficient CosmWasmClient.getHeight

webmaster128:


packages/sdk/src/cosmwasmclient.ts
```diff
@@ -149,6 +156,18 @@ export class CosmWasmClient {
     return response.node_info.network;
   }
 
+  public async getHeight(): Promise<number> {
+    if (this.anyValidAddress) {
```
ethanfrey: Nice trick here to optionally use any seen address

PR #141, opened 2020-03-08
Cache code details

webmaster128:
Code details are immutable and large. Let's not download and process them multiple times in one client.

~Based on #140~

### DIFF
--- a/packages/sdk/src/cosmwasmclient.spec.ts
+++ b/packages/sdk/src/cosmwasmclient.spec.ts
@@ -276,6 +276,19 @@ describe("CosmWasmClient", () => {
       // check data
       expect(new Sha256(result.data).digest()).toEqual(fromHex(expectedInfo.checksum));
     });
+
+    it("caches downloads", async () => {
+      pendingWithoutWasmd();
+      const client = new CosmWasmClient(wasmdEndpoint);
+      const openedClient = (client as unknown) as PrivateCosmWasmClient;
+      const getCodeSpy = spyOn(openedClient.restClient, "getCode").and.callThrough();
+
+      const result1 = await client.getCodeDetails(deployedErc20.codeId); // from network
+      const result2 = await client.getCodeDetails(deployedErc20.codeId); // from cache
+      expect(result2).toEqual(result1);
+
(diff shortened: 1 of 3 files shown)


packages/sdk/src/cosmwasmclient.spec.ts
```diff
@@ -236,6 +276,19 @@ describe("CosmWasmClient", () => {
...
+      pendingWithoutWasmd();
+      const client = new CosmWasmClient(wasmdEndpoint);
+      const openedClient = (client as unknown) as PrivateCosmWasmClient;
+      const getCodeSpy = spyOn(openedClient.restClient, "getCode").and.callThrough();
+
+      const result1 = await client.getCodeDetails(deployedErc20.codeId); // from network
+      const result2 = await client.getCodeDetails(deployedErc20.codeId); // from cache
+      expect(result2).toEqual(result1);
+
+      expect(getCodeSpy).toHaveBeenCalledTimes(1);
```
ethanfrey: Nice one

packages/sdk/src/cosmwasmclient.ts
```diff
@@ -276,27 +300,34 @@ export class CosmWasmClient {
...
+          builder: entry.builder || undefined,
+        };
+      },
     );
   }
 
   public async getCodeDetails(codeId: number): Promise<CodeDetails> {
-    const getCodeResult = await this.restClient.getCode(codeId);
+    const cached = this.codesCache.get(codeId);
+    if (cached) return cached;
```
ethanfrey: Style point

I thought linter wanted if body to be inside brackets.

webmaster128: Back in the tslint days that was the case. Now we have a more sophisticated `curly` configuration in eslint with the following requirements:
1. any multiline statement must have bracket ("multi-line")
2. if any of the cases in an if/elseif/else constructrion has brackets, all of them need them ("consistent")

This protects especially from this dangerous case:

```js
if (foo)
  bar1();
```

gets refactored to

```js
if (foo)
  bar1();
  bar2();
```

But for many cases where we have a short test and a single expression, we can write it in one line.

PR #142, opened 2020-03-11
Cache chainID in CosmWasmClient

webmaster128:


packages/sdk/src/cosmwasmclient.ts
```diff
@@ -148,14 +148,21 @@ export class CosmWasmClient {
...
+  private chainId: string | undefined;
 
   public constructor(url: string, broadcastMode = BroadcastMode.Block) {
     this.restClient = new RestClient(url, broadcastMode);
   }
 
-  public async chainId(): Promise<string> {
-    const response = await this.restClient.nodeInfo();
-    return response.node_info.network;
+  public async getChainId(): Promise<string> {
```
ethanfrey: +1 for caching.

The rest of this PR is just renaming and a bit such cleanup, right?

webmaster128: Right, we have caching, renaming to verb-prefixed method for consistency and extraction of the chain ID into `wasm.expectedChainId`.

PR #143, opened 2020-03-12
Update wasmd to 0.7.1

webmaster128:


packages/sdk/src/testutils.spec.ts
```diff
@@ -3,53 +3,8 @@ import { Bech32, Encoding } from "@iov/encoding";
...
-  const min = 128;
-  const max = 16383;
-  const random = Math.floor(Math.random() * (max - min)) + min;
-  const bytes = leb128Encode(random);
-
-  data[0x0136d3 + 1] = bytes[0];
-  data[0x0136d3 + 2] = bytes[1];
-
-  return data;
+export function getHackatom(): Uint8Array {
```
ethanfrey: Much simpler.
Though I am still a bit sad to see that code go away - cool stuff but not really maintainable.
It will live forever in git though.

webmaster128: Yeah I was feeling the same. Knowledge of those wasm internals will suvive and can easily be re-applied later on.

webmaster128:
>  It only affects tests, right? So no problem running this code against a v0.7.0 backend (same behavior as now, right?)

Right

PR #146, opened 2020-03-18
Various library refinements

webmaster128:
Originally intended to support a new article, but got a little off-topic.

- More documentation on main classes
- More testing, especially on the REST side
- Rename to `RestClient.txById` (singular)
- Make transaction execution error code available in transaction search

### DIFF
--- a/packages/sdk/src/restclient.ts
+++ b/packages/sdk/src/restclient.ts
@@ -1,4 +1,4 @@
-import { Encoding } from "@iov/encoding";
+import { Encoding, isNonNullObject } from "@iov/encoding";
 import axios, { AxiosError, AxiosInstance } from "axios";
 
 import { Coin, CosmosSdkTx, Model, parseWasmData, StdTx, WasmData } from "./types";
@@ -137,8 +137,6 @@ interface SearchTxsResponse {
   readonly limit: string;
   readonly txs: readonly TxsResponse[];
 }
-
-interface PostTxsParams {}
 
 export interface PostTxsResponse {
   readonly height: string;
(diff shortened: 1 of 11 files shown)


packages/sdk/src/restclient.ts
```diff
@@ -119,6 +119,10 @@ interface WasmError {
 export interface TxsResponse {
   readonly height: string;
   readonly txhash: string;
+  /** 🤷‍♂️ */
```
ethanfrey: This is vaguely equal to the module that defined the error.
So there is one for all the generic errors (invalid address), but then if wasmd defines a specific error type, and slashing another, these should have those codespaces on them.

`(Codespace, code)` needs to be a universally unique identifier

packages/sdk/src/signingcosmwasmclient.ts
```diff
@@ -103,14 +103,26 @@ export class SigningCosmWasmClient extends CosmWasmClient {
   private readonly signCallback: SigningCallback;
   private readonly fees: FeeTable;
 
+  /**
```
ethanfrey: Nice docs

packages/sdk/src/signingcosmwasmclient.ts
```diff
@@ -103,14 +103,26 @@ export class SigningCosmWasmClient extends CosmWasmClient {
...
+   * Creates a new client with signing capability to interact with a CosmWasm blockchain. This is the bigger brother of CosmWasmClient.
+   *
+   * This instance does a lot of caching. In order to benefit from that you should try to use one instance
+   * for the lifetime of your application. When switching backends, a new instance must be created.
+   *
+   * @param apiUrl The URL of a Cosmos SDK light client daemon API (sometimes called REST server or REST API)
+   * @param senderAddress The address that will sign and send transactions using this instance
+   * @param signCallback An asynchonous callback to create a signature for a given transaction. This can be implemented using secure key stores that require user interaction.
```
ethanfrey: Just checked this is async, which is very nice to call out to UI :+1:

Issue #148, opened 2020-03-18
Add CLI bindings for contracts in cosmwasm-examples 


ethanfrey:
This was inspired by the desire to make TS-CLI bindings for `mask` contract

It would be nice to have a CLI script for each one that would allow higher-level access (beyond `helpers.ts`). Have the proper types for the messages and the queries, and expose proper functions. Take in a Pen and SigningCosmWasmClient as args and do the rest.

* Upload code (this is generic, can be left in helpers)
* Instantiate code (with proper types)
* Execute code (one per method)
* Query (one per method)

Note: mask requires some double encoding (eg. the json message body for contract msg needs to be serialized, converted to bytes, and then base64-encoded). Making a helper here is a big advantage over doing it manually. Maybe for erc20 this doesn't make a big difference, but at least having the (auto-generated :smile: ) types defined here may make it easier for people ("how do I submit a transfer?")

PR #150, opened 2020-03-19
Update helpers to modern api, demo net

### LINKED ISSUE #147: Revisit examples for the CLI
I wanted to make some additions and realized `helpers.ts` is using RestClient and (almost) entirely superceded by `CosmWasmClient`. Let's give these an overhaul to the newest version.

ethanfrey:
Closes #147 

The readme works against the demo net. Please review the exisiting code in `helpers.ts`.

This is now updated README, showing go cli compatibility and connecting to other testnets.
I will address #148 based on top of this

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -81,10 +81,11 @@ The above code shows you the use of the API and various objects and is a great w
 how to embed cosmwasm-js into your project. However, if you just want a cli to perform some
 quick queries on a chain, you can use an extended set of helpers:
 
-1. Start a local wasmd blockchain, for example running the setup from `../../scripts/wasmd/start.sh`
-2. Start with `./bin/cosmwasm-cli --init examples/helpers.ts` (note the new init file)
-3. Deploy some erc20 contracts: `../../scripts/wasmd/init.sh`
-4. Play around as in the following example code
+Start with `./bin/cosmwasm-cli --init examples/helpers.ts`
+
+(This points to the Demo Net at https://lcd.demo-07.cosmwasm.com for ease of use. Other networks, look below)
+
+Setup Account:
 
 ```ts
 // you can hand-copy a mnemonic here, but this is easiest for reuse between sessions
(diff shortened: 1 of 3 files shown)


packages/cli/examples/helpers.ts
```diff
@@ -1,101 +1,100 @@
 /* eslint-disable @typescript-eslint/camelcase */
-import {
-  logs
-} from "@cosmwasm/sdk";
+import { FeeTable } from "@cosmwasm/sdk/types/signingcosmwasmclient";
```
webmaster128: Exported properly now (https://github.com/CosmWasm/cosmwasm-js/pull/151)

packages/cli/examples/helpers.ts
```diff
@@ -1,101 +1,100 @@
...
+import axios from "axios";
+import * as fs from "fs";
```
webmaster128: Both `axios` and `fs` are no runtime dependencies of `@cosmwasm/cli`, so they just work by luck. If we want to have them, I'd rather imports them by default and add to dependencies.

We already have

```
  let init = `
    import * as http from 'http';
    import * as https from 'https';
  `;
```

which I guess can be replaced with `axios`, since axios API can easily be used in browsers as well. Let me take care of this in a separate PR

packages/cli/README.md
```diff
@@ -81,55 +81,169 @@ The above code shows you the use of the API and various objects and is a great w
...
+  bech32prefix: "xrn:",
+}
+^D
+
+const mnemonic = loadOrCreateMnemonic("regen.key");
+const {address, client} = await connect(mnemonic, regenOptions);
+address
+
+// check some random genesis account
+client.getAccount("xrn:1pdfr7xuckj6lhdphdde6peres9ufwgpsv87mag")
```
webmaster128: `await` missing?

ethanfrey: Empty promises get resolved by the cli. 
Only needed when assigning

packages/cli/examples/helpers.ts
ethanfrey: Thanks. I will rebase and update

packages/cli/examples/helpers.ts
ethanfrey: Ditto on rebase and update

packages/cli/README.md
webmaster128: Interesting, thanks

PR #154, opened 2020-03-20
Support multiple init files

### LINKED ISSUE #149: Support multiple init files
If I have some generic setup code (helpers) and some contract-specific code (see #148) it would be nice to import both of them without merging the source files. This could be done if we support multiple instances of `--init` in `packages/cli/src/cli.ts`

```ts
export function main(originalArgs: readonly string[]): void {
  const parser = new ArgumentParser({ description: "The CosmWasm REPL" });
  parser.addArgument("--version", {
    action: "storeTrue",
    help: "Print version and exit",
  });
  parser.addArgument("--init", {
    metavar: "FILEPATH",
    help: "Read initial TypeScript code from file",
  });
  // ...
    if (args.init) {
    init += fs.readFileSync(args.init, "utf8") + "\n";
  }
```

webmaster128:
~Based on #152~

Migrates to yargs, which is more actively maintained.

You can now use
```
--init examples/foo.ts --init examples/bar.ts
```

or just

```
--init examples/foo.ts examples/bar.ts
```

Closes #149

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -11,8 +11,8 @@ export function main(originalArgs: readonly string[]): void {
     .options({
       // User options (we get --help and --version for free)
       init: {
-        describe: "Read initial TypeScript code from file",
-        type: "string",
+        describe: "Read initial TypeScript code from files",
+        type: "array",
       },
       // Maintainer options
       debug: {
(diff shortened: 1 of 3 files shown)


packages/cli/src/cli.ts
```diff
@@ -1,41 +1,32 @@
...
-  const args = parser.parseArgs([...originalArgs]);
-
-  if (args.version) {
-    const version = require(join(__dirname, "..", "package.json")).version;
-    console.info(version);
-    return;
-  }
+  const args = yargs
+    .options({
+      // User options (we get --help and --version for free)
```
ethanfrey: I was looking for the, nice to see that

PR #156, opened 2020-03-20
Add mask helpers

### LINKED ISSUE #148: Add CLI bindings for contracts in cosmwasm-examples

ethanfrey:
Closes #148 

Add docs and helper file for using the mask contract with `@cosmwasm/cli`

This walks through all execution paths of the mask contracts from sending tokens to staking positions and transfering it.


packages/cli/MASK.md
```diff
@@ -0,0 +1,179 @@
...
+```ts
+const hash = "1f50bbff503fd9c7bfe713bbf42b309cf88ef299fa76e0242051c9a7e25649a3"
+client.getCodes().then(codes => codes.filter(x => x.checksum === hash))
+```
+
+If it is not uploaded, we will upload it:
+
+```ts
+// Either download the code
+const wasmUrl = "https://github.com/CosmWasm/cosmwasm-examples/blob/mask-0.1.0/mask/contract.wasm?raw=true";
```
webmaster128: axios follows redirects – interesting

packages/cli/examples/mask.ts
```diff
@@ -0,0 +1,89 @@
+/*
+ * This file was generated with ❤️ by wasm.glass and is licensed
+ * for you under WTFPL OR 0BSD OR Unlicense OR MIT OR BSD-3-Clause.
+ * Note that different terms apply for the contract's source code and schema.
+ * Type generation powered by json-schema-to-typescript.
+ */
```
webmaster128: I'd remove this header, since you do much more in this file and it can be misleading this way.

packages/cli/examples/mask.ts
```diff
@@ -0,0 +1,89 @@
...
+                amount: types.Coin[];
+                from_address: string;
+                to_address: string;
+              };
+            }
+          | {
+              contract: {
+                contract_addr: string;
+                // this had to be changed - is Base64 encoded string
+                msg: string;
```
webmaster128: Hmm, in [handle_msg.json](https://api.wasm.glass/sources/62e43dc8eb111576d7721af6a4fc251564ceec007e0c895e7416ddb4de812cae/schema/handle_msg.json) it is

```
                "msg": {
                  "type": "array",
                  "items": {
                    "type": "integer",
                    "format": "uint8",
                    "minimum": 0
                  }
                },
```

because `Binary` schema is

```
    "Binary": {
      "type": "array",
      "items": {
        "type": "integer",
        "format": "uint8",
        "minimum": 0.0
      }
    },
```

which I guess is a bug

ethanfrey: It showed me `msg: number[]` which is a straight-forward encoding. Note that node typically encodes `Uint8Array` as intelligently as serde-json-wasm... as `[12, 54, 164, 210, 7, ...]`

We want base64 encoding, and this is out of scope of the json-schema generator. Maybe an enhancement there to detect this and replace with a `Base64` type, and add `type Base64 = string` above.

webmaster128: So the problem is `impl Serialize for Binary {` and `impl<'de> Deserialize<'de> for Binary {` is not known to the trait `JsonSchema`? This is interesting and bad at the same time. Then I think we don't get better type safety in messages than what we have in JSON. And we need to change back all `Binary` fields to just `String`

ethanfrey: I wish there was some way to mark it as "base64-encode string", but there is not.

Either it is "byte array" or "string". Since the standard way to encode byte arrays in JSON is base64. I would actually consider leave the JSON schema how it is (the Go struct also just uses `[]byte` to represent it), and modify the json-schema-to-typescript to annotate all such Uint8Array items as "base64 encoding".

PR #159, opened 2020-03-23
Use codeId from uploadResult

webmaster128:
Can be merged once demonet supports uploading the same code multiple times

### DIFF
--- a/packages/cli/MASK.md
+++ b/packages/cli/MASK.md
@@ -43,19 +43,16 @@ const wasmFile = ".../cosmwasm-examples/mask/contract.wasm"
 const wasm = fs.readFileSync(wasmFile);
 
 // Then upload it
-const up = await client.upload(wasm, { source: "https://crates.io/api/v1/crates/cw-mask/0.1.0/download", builder: "confio/cosmwasm-opt:0.7.3"});
-up
-up.logs[0].events[0]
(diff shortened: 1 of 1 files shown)


packages/cli/MASK.md
```diff
@@ -43,19 +43,16 @@ const wasmFile = ".../cosmwasm-examples/mask/contract.wasm"
...
-up.logs[0].events[0]
+const uploadResult = await client.upload(wasm, { source: "https://crates.io/api/v1/crates/cw-mask/0.1.0/download", builder: "confio/cosmwasm-opt:0.7.3"});
+uploadResult
+uploadResult.logs[0].events[0]
 ```
 
 Now, make an instance (as above):
 
 ```ts
-// get the proper codeId
```
ethanfrey: I do like showing how to get the codeId or an existing contract.

Even if we support re-uploading, we should document the prefered use-case of re-use.

webmaster128: Yeah, I agree. In this case I think we should restructure the doc to focus on one path (ether reuse or upload). Otherwise it is too much back and forth for someone unfamiliar with the flow. We could reuse existing code for the main tutorial and extract the upload into a new section at the bottom of the document.

ethanfrey: Sounds good.
Let me get another testnet up in the meantime.

ethanfrey:
Demonet-071.cosmwasm.com is live and using v0.7.1

dynst:
The `cli` package got removed in #1713

PR #165, opened 2020-05-07
Bugfix: sequence number is 0 for first tx

webmaster128:


webmaster128:
> How did this pass tests before?

This PR prepares #166. On current master, the first transaction from the faucet address is one of the init transaction (code upload). So sequence 0 is never a search result for the BCP searches. It worked by accident.

PR #166, opened 2020-05-07
Add some multi message signature tests

webmaster128:
Based on #165


webmaster128:
> Odd the errors there. I guess too many sigs makes sense, but wrong order???
> Did you raise issue on cosmos sdk to ask about this (or find docs that explain that)?

Oh yeah, it is real. The list of required signers is defined my the messages. The order is fixed and it is exhaustive. This basically makes it impossible to build n/m multisig using the native signing mechanism. All of this is described here: https://github.com/cosmos/cosmos-sdk/blob/v0.38.3/x/auth/types/stdtx.go#L89-L106 So I guess this works as intended.

PR #168, opened 2020-05-09
Create @cosmwasm/crypto

webmaster128:
Internalize the package and modernize it (especially added types of dependencies).

In contrast to @iov/crypto we don't use libsodium and bundled Wasm code anymmore (no Ed25519 right now).


webmaster128:
To be honest, the main motivation for this was flexibility (and allow thinks [like custom `k`s](https://github.com/CosmWasm/cosmwasm-js/pull/169) in the API, which probably does more harm than adding value). I've mixed feelings about this now. We probably should contribute to @iov/crypto until we have a very good reason to migrate away from that.

ethanfrey:
This PR is deprecated now after the current decision to rename and internalize a whole lot of iov packages, correct?

webmaster128:
Jap, it just exists because it is the base of #169, which I need for an article. Will cleanup when @cosmjs/crypto is in here.

PR #169, opened 2020-05-09
Demo usage of custom k

webmaster128:
This shows the usage of custom k to generate multiple signatures.

I don't think this should be merged since exposing such an API can create more danger than usage.


webmaster128:
Custom k is interesting to understand (see also https://medium.com/@simonwarta/signature-determinism-for-blockchain-developers-dbd84865a93e). However, this is can be dangerous when used incorrectly. So we don't add this to the codebase.

PR #180, opened 2020-05-30
Export symbols BankToken and Erc20Token

webmaster128:
The types BankToken and Erc20Token are needed to construct a codes without a connection.

BankTokens was removed to avoid maintaining an unnecessary type. Usage was unified with the Erc20Token.


webmaster128:
The relevant caller code is here: https://github.com/iov-one/token-finder/pull/13/files#diff-f953b3a2fb6d7cd16d698364c6a1dc45R7-R17

But in general, you want every type that is part of a public interface to be exported, such that the user can decide to either pass the value directly or create a variable of the correct type.

PR #183, opened 2020-06-02
Split Cosmos SDK and CosmWasm functionality

### LINKED ISSUE #182: Split Cosmos SDK (0.38) and CosmWasm functionality
Right now those are combined in the SDK package. The goal here is to make CosmWasm fully optional.

webmaster128:
Closes #182


packages/sdk38/src/cosmosclient.ts
```diff
@@ -0,0 +1,314 @@
...
+  /** Array of raw transactions */
+  readonly txs: ReadonlyArray<Uint8Array>;
+}
+
+/** Use for testing only */
+export interface PrivateCosmWasmClient {
+  readonly restClient: RestClient;
+}
+
+export class CosmosClient {
```
ethanfrey, line 120: Okay, to understand this:

1. We don't need to worry about tx parse codecs here as this is done in the RestClient/Server
2. This depends on the auth and bank modules of the cosmos-sdk but nothing else
3. It is meant to be easily extended with more functions by extending the class, correct?

packages/cosmwasm/src/cosmwasmclient.ts
```diff
@@ -103,24 +102,6 @@ export interface ContractDetails extends Contract {
   readonly initMsg: object;
 }
 
-/** A transaction that is indexed as part of the transaction history */
```
ethanfrey, line 106: This is all we remove here?

I was thinking that we get most of this for free by importing `CosmosClient from "@cosmwasm/sdk38"`.

It is not essential if this is going to be deprecated, but if this is a model for how we will do composition, adding more modules, it is nice to show the reuse.

packages/sdk38/src/cosmosclient.ts
webmaster128: 1. Yes, this connects to the REST server
2. Right now, yes. Before we had auth, bank, wasm and wasm is stripped off
3. Right now, there is no extensibility concept worked out. In @cosmwasm/cosmwasm we have a copy of that code. This is obviously no long term solution but we have to start somewhere.

packages/cosmwasm/src/cosmwasmclient.ts
webmaster128: I'll deduplicate that code here, no problem

webmaster128:
@ethanfrey I added two commits that deduplicate large chunks of the REST client.

Reusing the higher level clients is going to be tricky since
- you need different Rest clients
- you need multiple inheretance (SigningCosmWasmClient inherits SigningCosmosClient and CosmWasmClient)

I think we find a good solution once the basic sorting is done.

ethanfrey:
Okay, this is just about having two packages - one with cosmwasm support, one with basic bank support. And the way to adding custom (or extended standard like staking) modules is not yet done.

Fair enough. But good to have a tracking issue on that - and study kepler's approach

Issue #179, opened 2020-05-30
Remove BCP functionality from here


webmaster128:
The package @cosmwasm/bcp was [a successful experiment](https://medium.com/confio/adding-cosmwasm-to-the-neuma-multichain-wallet-ec657d893268) to show token contracts in a multichain wallet. There is no intend to add this functionality into a BCP based wallet like Neuma in production quality. In order to simplify our codebase, we plan the following steps:

1. Refactor @cosmwasm/faucet to work with @cosmwasm/sdk directly and remove the dependency on @cosmwasm/bcp. This implies the removal of smart contract token support from the faucet. It is simplifies to just support bank tokens, which is most important for fees.
2. Remove CosmWasm specifics (ERC20 smart contract token support) from @cosmwasm/bcp
3. Hand over @cosmwasm/bcp to IOV Core (under which target name?)

Mid-term additions

4. Split @cosmwasm/sdk into Cosmos SDK and CosmWasm functionality, such that CosmWasm becomes optional

@davepuchyr Number 2. and 3. is service work for IOV to express our gratitude for being able to build on top of IOV Core and maintain healty code for everyone in the ecosystem. Please folllow this ticket closely and chip in if I get something wrong.

ethanfrey:
This sounds good.

Happy to hand off the @cosmwasm/bcp experiment to iov-core if they want to maintain it, integrate it in their wallet or in any way make real use of it.

And yes, I also agree on refactoring the faucet to just use @cosmwasm/sdk (bank).

ethanfrey:
However, I am not sure about

> Split @cosmwasm/sdk into Cosmos SDK and CosmWasm functionality, such that CosmWasm becomes optional

We are talking about making a new repo to be modular and extensible, pulling in parts of iov-core and cosmwasm-js and others and making sure this is extensible for other modules. I would leave these as is here (working for all out 0.38-based cosmwasm releases) and do the refactoring on the new branch which will have plenty of redesign tasks as well. Unless you think there is a benefit to refactoring here first. (In any case, this is a separate ticket - bcp stuff is a full :+1:)

webmaster128:
As Ethan pointed out correctly, _Split @cosmwasm/sdk into Cosmos SDK and CosmWasm functionality, such that CosmWasm becomes optional_ is independent of this ticket. I created https://github.com/CosmWasm/cosmwasm-js/issues/182 for that. The order of #179 and #182 does not really matter technically.

[...]

PR #184, opened 2020-06-02
Remove CosmWasm functionality from BCP

webmaster128:
~Based on #183~

This is step 2 from #179

- [x] Make bcp package independent of cosmwasm
- [x] Rename CosmWasm* -> Cosmos*


ethanfrey:
I will review this when #179 is done so the diff is manageable

webmaster128:
@ethanfrey ping

packages/bcp/src/cosmwasmcodec.spec.ts
```diff
@@ -1,9 +1,9 @@
-import { Address, PostableBytes, PrehashType, SendTransaction, TokenTicker } from "@iov/bcp";
+import { PostableBytes, PrehashType } from "@iov/bcp";
 import { Encoding } from "@iov/encoding";
 
-import { CosmWasmCodec } from "./cosmwasmcodec";
+import { CosmosCodec } from "./cosmwasmcodec";
```
ethanfrey, line 4: why is the file still named `cosmwasmcodec` if it is now a more generic `CosmosCodec`?

packages/bcp/src/cosmwasmconnection.ts
```diff
@@ -85,50 +82,47 @@ function compareByHeightAndHash(a: IndexedTx, b: IndexedTx): number {
 /** Account and undefined are valid events. The third option means no event fired yet */
 type LastWatchAccountEvent = Account | undefined | "no_event_fired_yet";
 
-export class CosmWasmConnection implements BlockchainConnection {
+export class CosmosConnection implements BlockchainConnection {
```
ethanfrey, line 85: Rename and simplification is good.

I would rename the file as well.

packages/bcp/src/decode.spec.ts
```diff
@@ -159,40 +140,7 @@ describe("decode", () => {
           ],
         },
       };
-      expect(parseMsg(msg, defaultMemo, testdata.chainId, defaultTokens, defaultErc20Tokens)).toEqual(
-        defaultSendTransaction,
-      );
-    });
-
-    it("works for ERC20 send transaction", () => {
```
ethanfrey, line 167: that's easier

packages/bcp/src/cosmwasmcodec.spec.ts
webmaster128: God point, I forgot that, will update

PR #186, opened 2020-06-03
Add test accounts Alice/Bob 

### LINKED ISSUE #185: Let faucet and cosmwasm packages use different accounts
Starting with #184, faucet does not depend on cosmwasm anymore. As a consequence, lerna runs their tests in parallel. Unfortunately we need to block that in order to avoid sequence issues which occur because both packages use the same test account / mnemonic.

We should have a default test account, let's say `alice` with a handful of addresses and the faucet that is really made for the faucet. Then we can use alice my default as the test user.

webmaster128:
~Based on #184~

Closes #185

- [x] Add new accounts
- [x] Use new accounts
- [x] Reset lerna test concurrency to default


ethanfrey:
Please ping me when the others are done and this is ready to review so I don't forget

packages/cosmwasm/src/cosmwasmclient.searchtx.spec.ts
```diff
@@ -47,8 +47,8 @@ describe("CosmWasmClient.searchTx", () => {
 
   beforeAll(async () => {
     if (wasmdEnabled()) {
-      const pen = await Secp256k1Pen.fromMnemonic(faucet.mnemonic);
-      const client = new SigningCosmWasmClient(wasmd.endpoint, faucet.address, (signBytes) =>
+      const pen = await Secp256k1Pen.fromMnemonic(alice.mnemonic);
```
ethanfrey, line 50: Yes, using different accounts in different packages that target a live chain is much nicer.

scripts/wasmd/README.md
```diff
@@ -93,12 +93,36 @@ You should get output matching the following:
    Pubkey 1: AiDosfIbBi54XJ1QjCeApumcy/FjdtF+YhywPf3DKTx7<br>
    Pubkey 2: AzQg33JZqH7vSsm09esZY5bZvmzYwE/SY78cA0iLxpD7<br>
    Pubkey 3: A3gOAlB6aiRTCPvWMQg2+ZbGYNsLd8qlvV28m8p2UhY2<br>
-   Pubkey 4: Aum2063ub/ErUnIUB36sK55LktGUStgcbSiaAnL1wadu<br>
-2. Unused: for testing account state; this account never changes balances or nonces<br>
+   Pubkey 4: Aum2063ub/ErUnIUB36sK55LktGUStgcbSiaAnL1wadu
+2. **Alice**: Test account for the cosmwasm package that can run in parallel with faucet without sequence conflicts<br>
```
ethanfrey, line 97: Thanks for the explanation here.

PR #187, opened 2020-06-03
Remove BCP dependency from faucet

webmaster128:
Based on #186

Step 1 from #179


packages/faucet/src/faucet.spec.ts
```diff
@@ -15,175 +13,165 @@ function pendingWithoutWasmd(): void {
...
+      const readOnlyClient = new CosmosClient(httpUrl);
+      const account = await readOnlyClient.getAccount(recipient);
       assert(account);
       expect(account.balance).toEqual([
         {
-          quantity: "23456",
-          fractionalDigits: 6,
-          tokenTicker: "COSM" as TokenTicker,
+          amount: "23456",
+          denom: "ucosm",
```
ethanfrey, line 82: okay, so we use the raw denom from the blockchain.

I think this is fine actually. Let's just leave it like this, updates the amounts in config (10COSM -> 10_000_000 ucosm) and wait until there is some magic metadata on-chain to handle this translation.

packages/faucet/src/types.ts
```diff
@@ -1,7 +1,32 @@
...
+  /**
+   * The number of fractional digits the token supports.
+   *
+   * A quantity is expressed as atomic units. 10^fractionalDigits of those
+   * atomic units make up 1 token.
+   *
+   * E.g. in Ethereum 10^18 wei are 1 ETH and from the quantity 123000000000000000000
+   * the last 18 digits are the fractional part and the rest the wole part.
+   */
+  readonly fractionalDigits: number;
```
ethanfrey, line 24: Do we want this still? Or get rid of it for now?
We can have a working faucet that uses native tokens.

packages/faucet/src/faucet.spec.ts
webmaster128: This is a pretty low level function that takes a plain `Coin`. The higher level interface reads 10COSM from env variables and convers this internally

PR #188, opened 2020-06-04
Use new package names

webmaster128:
A renaming to @iov/cosmos-sdk to [prepare a move IOV Core](https://github.com/CosmWasm/cosmwasm-js/issues/179) and the remaining packages are now called @cosmjs/*

### DIFF
--- a/packages/bcp/package.json
+++ b/packages/bcp/package.json
@@ -39,7 +39,7 @@
     "pack-web": "yarn build-or-skip && webpack --mode development --config webpack.web.config.js"
   },
   "dependencies": {
-    "@cosmwasm/sdk38": "^0.8.0",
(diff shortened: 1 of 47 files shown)


packages/bcp/package.json
```diff
@@ -1,8 +1,11 @@
 {
-  "name": "@cosmwasm/bcp",
+  "name": "@iov/cosmos-sdk",
```
ethanfrey, line 2: I assume the next step is to move this package out of the monorepo, right?

webmaster128: Right. But before we can do that, we need to publish a version of @cosmjs/sdk38 to npm, which I'd do after #192

PR #190, opened 2020-06-04
Upgrade @iov/* dependencies to 2.3.2

webmaster128:
~Based on #188~


packages/bcp/src/address.ts
```diff
@@ -1,9 +1,7 @@
 import { PubKey, pubkeyToAddress as sdkPubkeyToAddress, pubkeyType } from "@cosmjs/sdk38";
 import { Address, Algorithm, PubkeyBundle } from "@iov/bcp";
 import { Secp256k1 } from "@iov/crypto";
-import { Encoding } from "@iov/encoding";
-
-const { toBase64 } = Encoding;
+import { toBase64 } from "@iov/encoding";
```
ethanfrey, line 4: That is a nicer API

packages/cli/src/cli.ts
```diff
@@ -122,17 +129,6 @@ export function main(originalArgs: readonly string[]): void {
   for (const [moduleName, symbols] of imports.entries()) {
     console.info(colors.yellow(`  * from ${moduleName}: ${symbols.join(", ")}`));
   }
-  const encodingHelpers = [
```
ethanfrey, line 125: and simplifies our code

PR #191, opened 2020-06-04
Move marshalTx/unmarshalTx to BCP

### LINKED ISSUE #189: Move marshalTx/unmarshalTx to BCP
Those are BCP specific binary representations

webmaster128:
Based on #190

Closes #189


packages/sdk38/src/encoding.ts
```diff
@@ -20,11 +20,6 @@ function sortJson(json: any): any {
   return result;
 }
 
-export function marshalTx(tx: StdTx): Uint8Array {
```
ethanfrey, line 25: This was never used internally. Interesting.

Nice to move simpler helpers to where they are used/needed

PR #193, opened 2020-06-05
Give blockchain more time to index transactions

webmaster128:
Usually not needed but this makes the CI more stable


webmaster128:
What changed recently that could make this fix necessary is that 3 tests run in parallel now: faucet, cosmwasm, bcp. All of them depend on sdk38 but the 3 don't depend on each other anymore.

PR #195, opened 2020-06-05
Setup browser testing in CI

webmaster128:
Many of the changes here are required since 3 packages run in parallel now: faucet, cosmwasm, bcp

### DIFF
--- a/packages/bcp/package.json
+++ b/packages/bcp/package.json
@@ -34,7 +34,7 @@
     "build-or-skip": "[ -n \"$SKIP_BUILD\" ] || yarn build",
     "test-node": "node jasmine-testrunner.js",
     "test-firefox": "yarn pack-web && karma start --single-run --browsers Firefox",
-    "test-chrome": "yarn pack-web && karma start --single-run --browsers ChromeHeadlessInsecure",
+    "test-chrome": "yarn pack-web && karma start --single-run --browsers ChromeHeadless",
     "test": "yarn build-or-skip && yarn test-node",
     "pack-web": "yarn build-or-skip && webpack --mode development --config webpack.web.config.js"
   },
--- a/packages/bcp/src/cosmosconnection.spec.ts
+++ b/packages/bcp/src/cosmosconnection.spec.ts
@@ -488,16 +488,13 @@ describe("CosmosConnection", () => {
 
       // search by height
       const heightResults = await connection.searchTx({ height: byIdResult.height });
-      expect(heightResults.length).toEqual(1);
-      const heightResult = heightResults[0];
-      expect(heightResult.transactionId).toEqual(transactionId);
-      assert(isConfirmedTransaction(heightResult), "Expected transaction to succeed");
-      assert(heightResult.log, "Log must be available");
-      const [firstHeightLog] = JSON.parse(heightResult.log);
-      expect(firstHeightLog.events.length).toEqual(2);
-      const heightTransaction = heightResult.transaction;
-      assert(isSendTransaction(heightTransaction), "Expected send transaction");
-      expect(heightTransaction).toEqual(unsigned);
+      expect(heightResults.length).toBeGreaterThanOrEqual(1);
+      expect(heightResults).toContain(
+        jasmine.objectContaining({
+          transactionId: transactionId,
+          transaction: unsigned,
+        }),
+      );
 
       connection.disconnect();
     });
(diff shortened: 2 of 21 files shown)


packages/bcp/package.json
```diff
@@ -34,7 +34,7 @@
     "build-or-skip": "[ -n \"$SKIP_BUILD\" ] || yarn build",
     "test-node": "node jasmine-testrunner.js",
     "test-firefox": "yarn pack-web && karma start --single-run --browsers Firefox",
-    "test-chrome": "yarn pack-web && karma start --single-run --browsers ChromeHeadlessInsecure",
+    "test-chrome": "yarn pack-web && karma start --single-run --browsers ChromeHeadless",
```
ethanfrey, line 37: Always nice to see the word `Insecure` disappear from the codebase :smile:

packages/bcp/src/cosmosconnection.spec.ts
```diff
@@ -41,30 +41,26 @@ function makeRandomAddress(): Address {
...
 
 /**
  * We use a different test account than the faucet, since BCP errors when transactions
  * with multiple messages are found.
  */
 const bob = {
-  mnemonic:
-    "economy stock theory fatal elder harbor betray wasp final emotion task crumble siren bottom lizard educate guess current outdoor pair theory focus wife stone",
-  path: HdPaths.cosmosHub(4),
+  mnemonic: "remain fragile remove stamp quiz bus country dress critic mammal office need",
```
ethanfrey, line 57: Why the change here?

Ah, this is to keep in sync with other iov/bcp code rather than in sync with other cosm(wasm)-js code, right?

packages/cosmwasm/src/cosmwasmclient.spec.ts
```diff
@@ -72,9 +74,10 @@ describe("CosmWasmClient", () => {
 
       const height1 = await client.getHeight();
       expect(height1).toBeGreaterThan(0);
-      await sleep(1_000);
+      await sleep(blockTime * 1.3); // tolerate chain being 30% slower than expected
```
ethanfrey, line 77: always good for CI

packages/sdk38/src/cosmosclient.searchtx.spec.ts
```diff
@@ -104,20 +105,22 @@ describe("CosmosClient.searchTx", () => {
           },
         };
         const transactionId = await client.getIdentifier(tx);
-        const heightBeforeThis = await client.getHeight();
         try {
           await client.postTx(tx.value);
         } catch (error) {
           // postTx() throws on execution failures, which is a questionable design. Ignore for now.
```
ethanfrey, line 111: What alternative would you suggest?

I think it is good to ensure people don't blindly ignore failures (if they don't read the exection result), but it should be easy to detect ("tx ran but failed in checktx/delivertx" along with the log info from the chain)

packages/bcp/package.json
webmaster128: Yeah this was never needed here. Will created that for IOV because of CORS issues in the original Gaia docker container.

packages/sdk38/src/cosmosclient.searchtx.spec.ts
ethanfrey: Do you want to create an issue for this approach and we can discuss this?

I would not change it here, but for sdk39 I think we will change lots of interfaces, so good to check all issues with the current design so we can improve on them,

packages/bcp/src/cosmosconnection.spec.ts
webmaster128: A few PRs ago, we decided to use the mnemonic "economy stock theory fatal elder harbor …" for the faucet package only. The usage here was forgotten. With browser testing and the new independence of faucet, bcp, cosmwasm, we get new parallel running conflicts that did not occur in the node testing by pure luck.

webmaster128:
> let's see what this does to CI times.

This is no big deal because we use the same `build` job's artifacts that we create for `test` and only run one fast headless browser.

packages/sdk38/src/cosmosclient.searchtx.spec.ts
webmaster128: Okay, I put it here: https://github.com/CosmWasm/cosmwasm-js/issues/199
There was https://github.com/CosmWasm/cosmwasm-js/issues/194 already which shows the current solution is unsufficient.

PR #196, opened 2020-06-05
Fork IOV dependencies

### LINKED ISSUE #192: Integrate util, encoding, and crypto package
After BCP is moved out, we have 3 @iov/* dependencies left: util, encoding, and crypto. Those should be included here to swap the dependency (IOV Core depends on Cosm JS).

util and crypto can be unchanged. encoding should be split:

- @cosmjs/encoding: fromAscii, fromBase64, fromHex, fromUtf8, toAscii, toBase64, toHex, toUtf8, Bech32, fromRfc3339, toRfc3339
- @cosmjs/math: Decimal, Int53, Uint32, Uint53, Uint64,
- @cosmjs/util: isUint8Array, isNonNullObject
- not copied: TransactionEncoder, JsonCompatible*, isJsonCompatible*

webmaster128:
This internalizes all our IOV dependencies and thus swaps the project dependency to: IOV Core depends on CosmJS.

Closes #192

Based on #191 and #195. Merge those first.


webmaster128:
Okay, this dependency graph makes me quite happy

cosmjs

ethanfrey:
Dep graph does look clean, and only the user-facing parts start bringing in significant non-primative (math/crypto) dependencies.

I would review this when the other PRs are merged and this is rebased (now that I just approved them), so I can see the diff easier.

Question: @iov/cosmos-sdk is moving to another repo later or no?

webmaster128:
Here we go, updated.

> Question: @iov/cosmos-sdk is moving to another repo later or no?

Yes. But before we can do that, we need to publish an initial version of @cosmjs/sdk38. Until then, it only works here. I set the package to private in the last commit, such that we never publish it ourselves and only IOV will publish it.

ethanfrey:
Sounds good.

PR #200, opened 2020-06-08
Improve hackatom test contract code

webmaster128:
- Reduce `uploadCustomContract`/`uploadContract` to just one function to make visible in the test which contract we is uploaded
- Add meta data to `getHackatom()` to avoid spreading it all over the place

### DIFF
--- a/packages/cosmwasm/src/restclient.spec.ts
+++ b/packages/cosmwasm/src/restclient.spec.ts
@@ -33,6 +33,7 @@ import cosmoshub from "./testdata/cosmoshub.json";
 import {
   alice,
   bech32AddressMatcher,
+  ContractUploadInstructions,
   deployedErc20,
   fromOneElementArray,
   getHackatom,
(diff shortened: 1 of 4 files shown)


packages/cosmwasm/src/restclient.spec.ts
```diff
@@ -60,19 +61,19 @@ function makeSignedTx(firstMsg: Msg, fee: StdFee, memo: string, firstSignature:
...
   const memo = "My first contract on chain";
   const theMsg: MsgStoreCode = {
     type: "wasm/store-code",
     value: {
       sender: alice.address0,
-      wasm_byte_code: toBase64(wasmCode),
-      source: "https://github.com/confio/cosmwasm/raw/0.7/lib/vm/testdata/contract_0.6.wasm",
-      builder: "confio/cosmwasm-opt:0.6.2",
+      wasm_byte_code: toBase64(contract.data),
+      source: contract.source || "",
```
ethanfrey, line 75: nice to make this variable

webmaster128:
> only changes test code if i understand

Exactly. I didn't like searching too long to know which contract was uploaded in a given test

Issue #199, opened 2020-06-08
Reconsider error handling in *Client


webmaster128:
We have ugly code like this a few times in the tests

```ts
        try {
          await client.postTx(tx.value);
        } catch (error) {
          // postTx() throws on execution failures, which is a questionable design. Ignore for now.
          // console.log(error);
        }
```

This is because we throw exceptions for failed transaction. My feeling is we do not want to do that but instead return a union like `TxSuccess | TxFailure` and only throw for lower level errors.

webmaster128:
The success case should have the current fields

```ts
  readonly logs: readonly Log[];
  readonly rawLog: string;
  /** Transaction hash (might be used as transaction ID). Guaranteed to be non-empty upper-case hex */
  readonly transactionHash: string;
```

plus a `readonly data?: Uint8Array` added in https://github.com/CosmWasm/cosmjs/pull/201

PR #201, opened 2020-06-09
Add data field to PostTxsResponse

### LINKED ISSUE #198: Add data field to PostTxsResponse

webmaster128:
The updated test contract turned out to be unnecessary for the testing, since the contract's `InitResponse.data` [is not available in the transaction](https://github.com/CosmWasm/wasmd/issues/134). But it does not hurt as well.

The result data is now tested using the code ID and the contract address.

Closes #198


webmaster128:
>  Do you want to expose this field in CosmosClient or CosmWasmClient?

Added to #199

PR #210, opened 2020-06-09
Reorganise error handling

### LINKED ISSUE #199: Reconsider error handling in *Client

willclarktech:
Closes #199

Currently errors are still thrown in some of the functions which wrap `postTx`.

### DIFF
--- a/packages/cosmwasm/src/cosmwasmclient.ts
+++ b/packages/cosmwasm/src/cosmwasmclient.ts
@@ -28,11 +28,26 @@ export interface Account {
   readonly sequence: number;
 }
 
-export interface PostTxResult {
+export interface PostTxFailureResult {
+  /** Transaction hash (might be used as transaction ID). Guaranteed to be non-empty upper-case hex */
+  readonly transactionHash: string;
+  readonly height: string;
+  readonly code: number;
+  readonly rawLog: string;
+}
+
+export interface PostTxSuccessResult {
   readonly logs: readonly Log[];
   readonly rawLog: string;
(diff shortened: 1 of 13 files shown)


packages/cosmwasm/src/cosmwasmclient.ts
```diff
@@ -28,11 +28,26 @@ export interface Account {
...
   readonly rawLog: string;
   /** Transaction hash (might be used as transaction ID). Guaranteed to be non-empty upper-case hex */
   readonly transactionHash: string;
+  readonly data?: Uint8Array;
+}
+
+export type PostTxResult = PostTxSuccessResult | PostTxFailureResult;
+
+export function isPostTxFailureResult(postTxResult: PostTxResult): postTxResult is PostTxFailureResult {
+  return (postTxResult as PostTxFailureResult).code !== undefined;
```
webmaster128, line 50: We should check this for truthiness because the undefined server response just comes from JSON serializing a 0 error code: https://github.com/cosmos/cosmos-sdk/blob/v0.38.4/types/result.go#L77

willclarktech: Ah great, I assumed it was a lazy check but should have known better!

PR #221, opened 2020-06-11
Add protobuf TxBody demo

### LINKED ISSUE #211: Create a proto encoded TxBody with a MsgSend
At https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/types/types.proto there is a protobuf definition for MsgSend. We need a demo package (just tests, no API design) that

- [x] Takes as JS object plus a `MsgSend` protobuf schema and turns it into a serialized binary
- [x] This binary needs to be embedded in a [google.protobuf.Any](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/any.proto)
- [x] Creates a `TxBody` protobuf document using the schema from https://github.com/cosmos/cosmos-sdk/blob/master/types/tx/types.proto and serialized it

This requires looking into possible protobuf encoding libraries.

The spec for that is [ADR-020](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-020-protobuf-transaction-encoding.md) and there is an overview available at https://warta.it/blog/cosmos-sdk-protobuf-signing

willclarktech:
Closes #211 and #222 

Here's the basic flow. From the documentation I'm not clear on how `type_url` or the interface registry are supposed to work, so will have to discuss that with @webmaster128 next week.


webmaster128:
> From the documentation I'm not clear on how `type_url` or the interface registry are supposed to work, so will have to discuss that with @webmaster128 next week.

If I understand [ADR-023](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-023-protobuf-naming.md#adoption-of-any) correctly, the usage of fake URLs are planned:

> The adoption of google.protobuf.Any as the recommended approach for encoding interface types (as opposed to oneof) makes package naming a central part of the encoding as fully-qualified message names now appear in encoded messages.


Here is how Google does the translation: 

> The pack methods provided by protobuf library will by default use 'type.googleapis.com/full.type.name' as the type URL and the unpack methods only use the fully qualified type name after the last '/' in the type URL, for example "foo.bar.com/x/y.z" will yield type name "y.z".

https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/any.proto#L87-L91

I raised that question in the Cosmos chat. Let's see.

packages/demo-protobuf/src/demo.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+    // fromAddress and toAddress are now Buffers
+    expect(Uint8Array.from(msgSendDecoded.fromAddress)).toEqual(msgSend.fromAddress);
+    expect(Uint8Array.from(msgSendDecoded.toAddress)).toEqual(msgSend.toAddress);
+    expect(msgSendDecoded.amount).toEqual(msgSend.amount);
+  });
+
+  it("works with dynamically loaded proto files", async () => {
+    pendingWithoutNode();
+    const protoPath = path.join(__dirname, "demo.proto");
+    const root = await protobuf.load(protoPath);
```
webmaster128, line 69: Here is a code snippet that shows how to load .proto files in a browser. I created that as part of a very very ealy demo locally:

```ts
const source = `
package awesomepackage;
syntax = "proto3";

import "google/protobuf/timestamp.proto";

message Person {
  string first_name = 1;
  string last_name = 2;
}

message AwesomeMessage {
  uint32 awesome_number = 2; // becomes awesomeNumber
  string title = 1;
  google.protobuf.Timestamp created = 3;
  google.protobuf.Any message = 5;
  Person author = 6;
  bytes checksum = 100;
  string description = 4;
}
`;

const root = proto
  .parse(source)
  .root.addJSON(proto.common.get("google/protobuf/timestamp.proto")!.nested!)
  .addJSON(proto.common.get("google/protobuf/any.proto")!.nested!)
  .resolveAll();
// const root = parsed.root;

// Obtain a message type
const AwesomeMessage = root.lookupType("awesomepackage.AwesomeMessage");
```

packages/demo-protobuf/src/demo.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      memo: "Some memo",
+      timeoutHeight: 9999,
+      extensionOptions: [],
+    });
+    const txBodyBytes = TxBody.encode(txBody).finish();
+
+    // Deserialization
+    const txBodyDecoded = TxBody.decode(txBodyBytes);
+    const msg = txBodyDecoded.messages[0];
+    assert(msg.value);
```
webmaster128, line 51: This is a bit of annoyance in protobuf: [default values](https://developers.google.com/protocol-buffers/docs/proto3#default_values) are ususally not encoded, such that undefined and empty become the same thing. For the well known types we use often (e.g. Any, TxBody), we could create some encoding/decoding wrappers that normalize undefined to an empty byte array.

packages/demo-protobuf/src/demo.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+    // fromAddress and toAddress are now Buffers
+    expect(Uint8Array.from(msgSendDecoded.fromAddress)).toEqual(msgSend.fromAddress);
+    expect(Uint8Array.from(msgSendDecoded.toAddress)).toEqual(msgSend.toAddress);
+    expect(msgSendDecoded.amount).toEqual(msgSend.amount);
+  });
+
+  it("works with dynamically loaded proto files", () => {
+    const { root } = protobuf.parse(demoProto);
+    const typeUrl = "/demo.MsgDemo";
+    const encoder = root.lookupType(getTypeName(typeUrl));
```
webmaster128, line 63: Oh, this works without the fully qualified name (just `MsgDemo` from `demo.MsgDemo`)

packages/demo-protobuf/src/registry.ts
```diff
@@ -0,0 +1,21 @@
+import protobuf from "protobufjs";
+
+import { cosmos_sdk as cosmosSdk, google } from "./generated/codecimpl";
+
+export class Registry {
+  private readonly types: Map<string, protobuf.Type>;
+
+  constructor(customTypes: readonly [string, protobuf.Type][] = []) {
+    this.types = new Map<string, protobuf.Type>([
+      ["/cosmos.Coin", (cosmosSdk.v1.Coin as unknown) as protobuf.Type],
```
webmaster128, line 10: This does not seem to be a safe cast, as `protobuf.Type` contains all the dynamic stuff like

```ts
    /** Message fields. */
    public fields: { [k: string]: Field };

    /** Oneofs declared within this namespace, if any. */
    public oneofs: { [k: string]: OneOf };

    /** Extension ranges, if any. */
    public extensions: number[][];

    /** Reserved ranges, if any. */
    public reserved: (number[]|string)[];

    /** Message fields by id. */
    public readonly fieldsById: { [k: number]: Field };

    /** Fields of this message as an array for iteration. */
    public readonly fieldsArray: Field[];

    /** Oneofs of this message as an array for iteration. */
    public readonly oneofsArray: OneOf[];
```

which does not exist in the generated classes

packages/demo-protobuf/types/demo.proto.d.ts
```diff
@@ -0,0 +1,2 @@
+declare const _default: '\nsyntax = "proto3";\npackage demo;\n\nmessage MsgDemo {\n  string example = 1;\n}\n';
```
webmaster128, line 1: Wow, all the string made it into the type. We should widen the type to `string`

packages/demo-protobuf/src/demo.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      amount: "1234567890",
+    });
+    const msgSend = MsgSend.create({
+      fromAddress: Uint8Array.from(Array.from({ length: 20 }, () => 1)),
+      toAddress: Uint8Array.from(Array.from({ length: 20 }, () => 2)),
+      amount: [coin],
+    });
+    const msgSendBytes = MsgSend.encode(msgSend).finish();
+    const msgSendWrapped = Any.create({
+      type_url: "/cosmos.bank.MsgSend",
```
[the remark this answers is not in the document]
willclarktech: I don't believe so.

packages/demo-protobuf/src/demo.spec.ts
willclarktech: Yes, I think because the root is attached to the namespace `demo`.

packages/demo-protobuf/types/demo.proto.d.ts
willclarktech: Seems to be a bug in either TS or ESLint, which claims the type `string` is unnecessary because it can be inferred.

packages/demo-protobuf/src/demo.proto.ts
```diff
@@ -0,0 +1,8 @@
+export default `
+syntax = "proto3";
+package demo;
+
+message MsgDemo {
+  string example = 1;
+}
+`;
```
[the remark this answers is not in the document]
willclarktech: I guess for a demo readability > speed.

packages/demo-protobuf/types/demo.proto.d.ts
webmaster128: But in ESLint I'd say. Ot an incompatibility between ESLint and TS since the widening rules of TS changed over time. `as string` works but

```ts
const proto: string = `
syntax = "proto3";
package demo;

message MsgDemo {
  string example = 1;
}
`;

export default proto;
```

should work as well.

packages/demo-protobuf/src/demo.proto.ts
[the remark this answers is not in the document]
webmaster128: Sure!

packages/demo-protobuf/src/registry.ts
```diff
@@ -0,0 +1,30 @@
...
+  ) => protobuf.Writer;
+  readonly decode: (reader: protobuf.Reader | Uint8Array, length?: number) => protobuf.Message<{}>;
+};
+
+export class Registry {
+  private readonly types: Map<string, GeneratedType>;
+
+  constructor(customTypes: Iterable<[string, GeneratedType]> = []) {
+    this.types = new Map<string, GeneratedType>([
+      ["/cosmos.Coin", (cosmosSdk.v1.Coin as unknown) as GeneratedType],
```
webmaster128, line 19: Nice! This should work without casting now, right?

willclarktech: It's not so easy. See https://github.com/CosmWasm/cosmjs/pull/221/commits/f45cc06a5c0781c72d36e509484cca02a1d482ff for a first attempt.

packages/demo-protobuf/src/registry.ts
```diff
@@ -0,0 +1,27 @@
+import protobuf from "protobufjs";
+
+import { cosmos_sdk as cosmosSdk, google } from "./generated/codecimpl";
+
+abstract class GeneratedType {
+  public abstract create(properties?: { [k: string]: any }): any;
+  public abstract encode(message: any | { [k: string]: any }, writer?: protobuf.Writer): protobuf.Writer;
+  public abstract decode(reader: protobuf.Reader | Uint8Array, length?: number): any;
+}
```
webmaster128, line 9:
```ts
interface GeneratedType {
  readonly create: (properties?: { [k: string]: any }) => any;
  readonly encode: (message: any | { [k: string]: any }, writer?: protobuf.Writer) => protobuf.Writer;
  readonly decode: (reader: protobuf.Reader | Uint8Array, length?: number) => any;
}
```

works as well. For the generated types like `cosmosSdk.v1.Coin` this is the static interface and for the dynamic types it applies to instances of `protobuf.Type`.

willclarktech: Huh, I tried something like this and it didn't like it. Not sure what was different.

packages/demo-protobuf/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "baseUrl": ".",
+    "outDir": "build",
+    "declarationDir": "build/types",
+    "experimentalDecorators": true,
```
webmaster128, line 7: Looking at `packages/demo-protobuf/build/decorator.js`, TypeScript seems to have its own compatibility implementation, such that core-js is not required.

packages/demo-protobuf/src/decorator.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    const Any = myRegistry.lookupType("/google.protobuf.Any")!;
+
+    const msgDemo = MsgDemoT.create({
+      example: "Some example text",
+    });
+    const msgDemoBytes = MsgDemoT.encode(msgDemo).finish();
+    const msgDemoWrapped = Any.create({
+      type_url: typeUrl,
+      value: msgDemoBytes,
+    });
```
webmaster128, line 31: This is great stuff. Now the user would not want to see `MsgDemoT` at all but pass an instance of `MsgDemo` into the registry and get an `Any` instance out. Can you make that happen?

I guess the name "registry" might not be perfect for the resulting structure, but we can just keep it until we see how it looks.

packages/demo-protobuf/src/decorator.ts
```diff
@@ -0,0 +1,20 @@
...
+
+function getTypeName(typeUrl: string): string {
+  const parts = typeUrl.split(".");
+  return parts[parts.length - 1];
+}
+
+export function CosmosMessage(registry: Registry, typeUrl: string): TypeDecorator<any> {
+  return (constructor: Constructor<Message<any>>) => {
+    const typeName = getTypeName(typeUrl);
+    const generatedType = util.decorateType(constructor, typeName);
```
webmaster128, line 13: This seems to be implemented via reflection (https://github.com/protobufjs/protobuf.js/blob/v6.9.0/src/util.js#L113-L143), which is great since we doon't need the .proto or JSON implementation at all then.

webmaster128:
The default values of the class seem to be an issue. Default values are not mentioned in the documentation here: https://github.com/protobufjs/protobuf.js#using-decorators. With the following diff the tests pass:

```diff
diff --git a/packages/demo-protobuf/src/decorator.spec.ts b/packages/demo-protobuf/src/decorator.spec.ts
index bc61e47..93d4966 100644
--- a/packages/demo-protobuf/src/decorator.spec.ts
+++ b/packages/demo-protobuf/src/decorator.spec.ts
@@ -17,7 +17,8 @@ describe("decorator demo", () => {
       public readonly booleanDemo: boolean = true;
 
       @CosmosField.String(2)
-      public readonly stringDemo: string = "example text";
+      // @ts-ignore
+      public readonly stringDemo: string;
 
       @CosmosField.Bytes(3)
       public readonly bytesDemo: Uint8Array = Uint8Array.from([1, 2, 3, 4, 5, 6, 7, 8]);
@@ -29,14 +30,21 @@ describe("decorator demo", () => {
       public readonly uint64Demo: number = 123;
 
       @CosmosField.Repeated(6)
-      public readonly listDemo: readonly string[] = ["this", "is", "a", "list"];
+      // @ts-ignore
+      public readonly listDemo: readonly string[];
     }
 
     const MsgDemoT = myRegistry.lookupType(typeUrl)!;
     const TxBody = myRegistry.lookupType("/cosmos.tx.TxBody")!;
     const Any = myRegistry.lookupType("/google.protobuf.Any")!;
 
-    const msgDemo = MsgDemoT.create();
+    console.log(MsgDemoT.create);
+
+    const msgDemo = MsgDemoT.create({
+      int64Demo: 20, // value overridden by `public readonly uint64Demo: number = 123;`
+      stringDemo: "foo",
+      listDemo: ["a", "b", "c"]
+    });
     const msgDemoBytes = MsgDemoT.encode(msgDemo).finish();
     const msgDemoWrapped = Any.create({
       type_url: typeUrl,
```

packages/demo-protobuf/src/decorator.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export const CosmosField = {
+  Boolean: (id: number) => Field.d<boolean>(id, "bool"),
+
+  String: (id: number) => Field.d<string>(id, "string"),
+  Bytes: (id: number) => Field.d<Uint8Array>(id, "bytes"),
+
+  Int64: (id: number) => Field.d<number>(id, "int64"),
+  UInt64: (id: number) => Field.d<number>(id, "uint64"),
```
webmaster128, line 25: I'd follow the JavaScript style where "uint" is an atomic word (e.g. Int8Array, Uint8Array)

packages/demo-protobuf/src/decorator.spec.ts
willclarktech: https://github.com/CosmWasm/cosmjs/issues/238

PR #224, opened 2020-06-16
Upgrade typescript and @typescript-eslint/*

webmaster128:


webmaster128:
Arg, this brings us in trouble with top-level await in the cli. Will investigate.

webmaster128:
Here we go

PR #225, opened 2020-06-16
Use --es6 to generate proto .js

webmaster128:


webmaster128:
Blocked by https://github.com/protobufjs/protobuf.js/issues/1308

webmaster128:
Will get back to this when my PR is merged upstream. But there is no real need for this right now.

PR #233, opened 2020-06-19
Stabilize unsuccessful tx search

webmaster128:
In https://app.circleci.com/pipelines/github/CosmWasm/cosmjs/639/workflows/d43ce96b-a51f-4131-810c-8ebcb0eebed0/jobs/1991/steps the test fails sometime because there is not enough time between block creation and indexing the unsuccessful transaction. There is no sleep statement for `sendUnsuccessful` and the test before the failing one is very fast:

```
@cosmjs/sdk38:   2 CosmosClient.searchTx
@cosmjs/sdk38:     2.1 with SearchByIdQuery
@cosmjs/sdk38:       ✓ can search successful tx by ID (0.012 sec) 
@cosmjs/sdk38:       ✗ can search unsuccessful tx by ID (0.027 sec)
@cosmjs/sdk38:         - Expected 0 to equal 1.
@cosmjs/sdk38:         - Expected undefined to equal <jasmine.objectContaining(Object(…)>.
@cosmjs/sdk38:       ✓ can search by ID (non existent) (0.003 sec)
@cosmjs/sdk38:       ✓ can search by ID and filter by minHeight (0.028 sec)
```

In order to avoid this issue, the unsuccessful tx is not posted first, allowing us to utilize the sleep in

```ts
        const recipient = makeRandomAddress();
        const transferAmount = coins(1234567, "ucosm");
        const result = await client.sendTokens(recipient, transferAmount);
        await sleep(75); // wait until tx is indexed
        const txDetails = await new RestClient(wasmd.endpoint).txById(result.transactionHash);
        sendSuccessful = {
          sender: faucet.address,
          recipient: recipient,
          hash: result.transactionHash,
          height: Number.parseInt(txDetails.height, 10),
          tx: txDetails.tx,
        };
```

for both transactions.

### DIFF
--- a/packages/cosmwasm/src/cosmwasmclient.searchtx.spec.ts
+++ b/packages/cosmwasm/src/cosmwasmclient.searchtx.spec.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/camelcase */
-import { Coin, CosmosSdkTx, isMsgSend, makeSignBytes, MsgSend, Secp256k1Pen } from "@cosmjs/sdk38";
+import { Coin, coins, CosmosSdkTx, isMsgSend, makeSignBytes, MsgSend, Secp256k1Pen } from "@cosmjs/sdk38";
 import { assert, sleep } from "@cosmjs/utils";
 
 import { CosmWasmClient, isPostTxFailure } from "./cosmwasmclient";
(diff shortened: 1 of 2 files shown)


willclarktech:
Does this make the test order important?

webmaster128:
> Does this make the test order important?

No. It changes the order within the `beforeAll` block, such that both broadcasts happen before `await sleep(75); // wait until tx is indexed`

PR #241, opened 2020-06-23
Decorate MsgSend

webmaster128:
This example shows how 9 lines of decorated code replace 96 lines of autogenerated code. It shows how easy it can be to register custom types.

### DIFF
--- a/packages/demo-protobuf/src/decorator.ts
+++ b/packages/demo-protobuf/src/decorator.ts
@@ -29,6 +29,9 @@ export const cosmosField = {
   int64: (id: number): FieldDecorator => Field.d<number>(id, "int64"),
   uint64: (id: number): FieldDecorator => Field.d<number>(id, "uint64"),
 
-  repeatedString: (id: number): FieldDecorator => Field.d<string[]>(id, "string", "repeated"),
   nested: (id: number, ctor: Constructor<Message<{}>>): FieldDecorator => Field.d(id, ctor),
+
+  repeatedString: (id: number): FieldDecorator => Field.d<string[]>(id, "string", "repeated"),
+  repeatedNested: (id: number, ctor: Constructor<Message<{}>>): FieldDecorator =>
+    Field.d(id, ctor, "repeated"),
 };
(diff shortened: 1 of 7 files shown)


packages/demo-protobuf/src/decorator.ts
```diff
@@ -29,6 +29,9 @@ export const cosmosField = {
   int64: (id: number): FieldDecorator => Field.d<number>(id, "int64"),
   uint64: (id: number): FieldDecorator => Field.d<number>(id, "uint64"),
 
-  repeatedString: (id: number): FieldDecorator => Field.d<string[]>(id, "string", "repeated"),
   nested: (id: number, ctor: Constructor<Message<{}>>): FieldDecorator => Field.d(id, ctor),
+
+  repeatedString: (id: number): FieldDecorator => Field.d<string[]>(id, "string", "repeated"),
+  repeatedNested: (id: number, ctor: Constructor<Message<{}>>): FieldDecorator =>
```
webmaster128, line 35: Should this be `repeatedMessage` instead?

willclarktech: If so, `nested` should probably be `message` too. Another option is `nestedMessage`. I don't have a strong opinion here.

webmaster128: Good point. Let's use `message`/`repeatedMessage`

webmaster128: Done

packages/demo-protobuf/src/msgs.ts
```diff
@@ -0,0 +1,27 @@
...
+@cosmosMessage(defaultRegistry, "/cosmos.Coin")
+export class Coin extends Message<{}> {
+  @cosmosField.string(1)
+  public readonly denom?: string;
+
+  @cosmosField.string(2)
+  public readonly amount?: string;
+}
+
+@cosmosMessage(defaultRegistry, "/cosmos.bank.MsgSend")
```
ethanfrey, line 17: Just saw this. This is amazing!

You gotta write this up somewhere when it is finished.

webmaster128: :D Let's see if a chain accepts it

PR #245, opened 2020-06-24
Upgrade CosmWasm to 0.9

### LINKED ISSUE #208: Add support for CosmWasm 0.9
CosmWasm 0.9 is planned to be ready by the end of this month. Not too much change is required here but we need to add support for the new contract upgrade message types.

- [ ] Upgrade local blockchain to 0.9
- [ ] Upgrade test contracts to 0.9
- [ ] Add support for new message types
- [ ] Create a demo test case that upgrades a contract

### LINKED ISSUE #213: Test response data from contract in handle
Requires #208 and an updated hackatom contract.

In https://github.com/CosmWasm/cosmwasm/pull/414, the hackatom testing contract got a dummy reponse `data` of F00BAA. We should test that we get this data in the client in the execute section of the test "can upload, instantiate and execute wasm".

webmaster128:
Closes #208
Closes #213


packages/cosmwasm/src/signingcosmwasmclient.ts
```diff
@@ -86,6 +104,20 @@ export interface UploadResult {
...
+/**
+ * The options of an .instantiate() call.
+ * All properties are optional.
+ */
+export interface InstantiateOptions {
+  readonly memo?: string;
+  readonly transferAmount?: readonly Coin[];
+  /**
+   * A bech32 encoded address of an admin account.
+   * Caution: an admin has the privillage to upgrade a contract. If this is not desired, do not set this value.
```
willclarktech, line 116: Typo: "privilege"

packages/cosmwasm/src/signingcosmwasmclient.ts
```diff
@@ -234,6 +282,108 @@ export class SigningCosmWasmClient extends CosmWasmClient {
     };
   }
 
+  public async updateAdmin(contractAddress: string, newAdmin: string, memo = ""): Promise<ChangeAdminResult> {
+    const updateAdminMsg: MsgUpdateAdmin = {
+      type: "wasm/update-contract-admin",
+      value: {
+        sender: this.senderAddress,
+        contract: contractAddress,
+        // eslint-disable-next-line @typescript-eslint/camelcase
```
willclarktech, line 291: I wish there was a way to tell this rule when it's allowed and when not.

webmaster128: eslint7 brings a new rule for naming conventions: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/naming-convention.md This allows you to set: _allow strictCamelCase for variables; allow strictCamelCase or snake_case for properties_.

What I find more annoying (and this won't change) is all the needed exceptions for cases when you don't actually decide how something is named (creating objects of a type defined elsewhere, destructing objects etc.).

willclarktech: Yeah, that's what I meant: what's allowed can't be expressed in terms of regexes even if the analyzer knows it's a property vs a variable.

PR #258, opened 2020-06-29
Fix: error TS2304: cannot find name 'window'

hskang9:
Fixes error that appears when running command `yarn build`
```
yarn run v1.22.4
$ tsc
../../../node_modules/@types/reach__router/index.d.ts:12:30 - error TS2304: Cannot find name 'Window'.

12 export type WindowLocation = Window['location'] & HLocation;
```

### DIFF
--- a/packages/cli/tsconfig.json
+++ b/packages/cli/tsconfig.json
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "baseUrl": ".",
+    "lib": ["dom", "es5", "es6"],
     "outDir": "build",
     "rootDir": "src"
   },


hskang9:
https://www.loom.com/share/e5f33f17e8e348c4b2dfa23abca6ce4d

webmaster128:
Welcome @hskang9!

Thank you for the report and the video. The dependency `../../../node_modules/@types/reach__router/index.d.ts:12:30` is not in our code. @cosmjs/cli should not depend on browser libs, like "dom". Can you show me the "temp-frontend" project to better understand were the problem comes from?

hskang9:
Ah tempt-frontend is the other irrelevent workspace that I am working on, I just opened a new window there. I agree that cli should not use dom, but I do not understand why other libraries need 'window' when the cosmwasm-js lib is built. In fact, this just does not happen in only on @cosmjs/cli. It happens on @cosmjs/utils and other packages in the monorepo 🧐. I use node version v12.17.0.

hskang9:
Maybe typescript acts differently between node versions. Is there a required node.js version to build this repo?

webmaster128:
I saw something different: When you execute `yarn build` in `packages/cli` (like in the video), then `../../../node_modules` is outside of the `cosmjs` repo. Did you forget to `yarn install` the cosmjs dependencies?

webmaster128:
Closing because of https://github.com/CosmWasm/cosmjs/pull/258#issuecomment-652220424. Feel free to re-open if you think this is a bug in this repo.

PR #260, opened 2020-06-30
Improve protobuf registry

### LINKED ISSUE #238: Develop registry in `demo-protobuf` to hide more functionality

willclarktech:
Closes #238 
Not sure what exactly the end state is, but this is at least cleaner than before.

### DIFF
--- a/packages/proto-signing/src/registry.ts
+++ b/packages/proto-signing/src/registry.ts
@@ -9,23 +9,23 @@ export interface GeneratedType {
   readonly decode: (reader: protobuf.Reader | Uint8Array, length?: number) => any;
 }
 
-export type EncodeObject = {
+export interface EncodeObject {
   readonly typeUrl: string;
   readonly value: any;
-};
+}
 
-export type DecodeObject = {
+export interface DecodeObject {
   readonly typeUrl: string;
   readonly value: Uint8Array;
-};
+}
 
-export type TxBodyValue = {
+export interface TxBodyValue {
   readonly messages: readonly EncodeObject[];
   readonly memo?: string;
   readonly timeoutHeight?: number;
   readonly extensionOptions?: google.protobuf.IAny[];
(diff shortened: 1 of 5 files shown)


packages/proto-signing/src/registry.ts
```diff
@@ -8,24 +9,111 @@ export interface GeneratedType {
...
+  readonly typeUrl: string;
+  readonly value: any;
+};
+
+export type DecodeObject = {
+  readonly typeUrl: string;
+  readonly value: Uint8Array;
+};
+
+export type TxBodyValue = {
```
webmaster128, line 22: Can we use interfaces for those types?

packages/proto-signing/src/registry.ts
```diff
@@ -8,24 +9,111 @@ export interface GeneratedType {
...
+    const { cosmosCoin, cosmosMsgSend, cosmosTxBody, googleAny } = defaultTypeUrls;
     this.types = new Map<string, GeneratedType>([
-      ["/cosmos.Coin", cosmosSdk.v1.Coin],
-      ["/cosmos.bank.MsgSend", cosmosSdk.x.bank.v1.MsgSend],
-      ["/cosmos.tx.TxBody", cosmosSdk.tx.v1.TxBody],
-      ["/google.protobuf.Any", google.protobuf.Any],
+      [cosmosCoin, cosmosSdk.v1.Coin],
+      [cosmosMsgSend, cosmosSdk.x.bank.v1.MsgSend],
+      [cosmosTxBody, cosmosSdk.tx.v1.TxBody],
+      [googleAny, google.protobuf.Any],
```
webmaster128, line 46: I think we should remove the `TxBody` and `Any` here, as we don't need to encode it as an `Any`. `TxBody` lives at a different level and is not as dynamic. The only thing we really need the registry for is message types and types used by messages (such as Coin).

PR #262, opened 2020-06-30
Decode protobuf Buffers as Uint8Arrays

### LINKED ISSUE #239: Wrap protobufjs `Field` to use `Uint8Array` instead of `Buffer` in Node.js

willclarktech:
Closes #239

I don't see a straightforward way to wrap `Field`. This is an 80/20 solution.﻿

### DIFF
--- a/packages/proto-signing/src/registry.ts
+++ b/packages/proto-signing/src/registry.ts
@@ -68,7 +68,7 @@ export class Registry {
     }
     const type = this.lookupTypeWithError(typeUrl);
     const created = type.create(value);
-    return type.encode(created).finish();
+    return Uint8Array.from(type.encode(created).finish());
   }
 
   public encodeTxBody(txBodyFields: TxBodyValue): Uint8Array {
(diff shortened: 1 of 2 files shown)


packages/proto-signing/src/registry.ts
```diff
@@ -94,7 +94,13 @@ export class Registry {
       return this.decodeTxBody(value);
     }
     const type = this.lookupTypeWithError(typeUrl);
-    return type.decode(value);
+    const decoded = type.decode(value);
+    Object.entries(decoded).forEach(([key, val]: [string, any]) => {
+      if (val.constructor.name === "Buffer") {
```
webmaster128, line 99: We can also use `Buffer.isBuffer`:

```ts
  if (typeof Buffer !== "undefined" && typeof Buffer.isBuffer !== "undefined") {
    // Buffer.isBuffer is available at runtime
    if (Buffer.isBuffer(data)) return false;
  }
```

See isUint8Array

webmaster128: Alternatively we can simply do

```ts
if (Object.prototype.toString.call(val) === "[object Uint8Array]") {
  // here val is `Uint8Array` or `Buffer`
  decoded[key] = Uint8Array.from(val);
}
```

willclarktech: I like the `Buffer.isBuffer` option as being the closest to what we actually mean.

willclarktech:
Code coverage loss is due to the `Buffer` check.

PR #273, opened 2020-07-02
Add typedoc setup

### LINKED ISSUE #234: Setup typedoc
We have typedoc configurations ready, but don't use them right now.

- [ ] Add typedoc build script ([like here](https://github.com/iov-one/iov-core/tree/master/scripts/typedoc/bin) or by using npx. Since typedoc is not always up-to-date, it'd like to keep it out of our main yarn.lock)
- [ ] Build docs in a separate CI step
- [ ] Push docs to `gh-pages` branch as part of the CI

webmaster128:
~Setp 1 and 2 from~ Closes #234


scripts/docs/copy_docs.sh
```diff
@@ -0,0 +1,47 @@
...
+# assets
+mkdir -p "$assets_dir"
+cp docs/cosmjs-tree.png "$assets_dir"
+
+{
+  echo '<h1>CosmJS documentation</h1>'
+  echo '<p>CosmJS is a modular library consisting of multiple packages that are organized like this:</p>'
+  echo '<p><a href="assets/cosmjs-tree.png"><img alt="CosmJS dependency graph" src="assets/cosmjs-tree.png" style="max-width: 100%; width: 950px" /></a></p>'
+  echo '<h2>Package documentation</h2>'
+  echo '<p>The following packages are can be used as a library.</p>'
```
willclarktech, line 22: "can be used as libraries"

willclarktech: Faucet ends up in this list.

webmaster128: Good point. The faucet could be used as a library as well if there was an index.ts exporting the classes. But I don't see a need for that right now. Will clean up.

Issue #276, opened 2020-07-06
sdk38: add support for more modules


webmaster128:
## TypeScript messages types

Arbitrary messages can be sent as JavaScript objects already, but it would be nice to have TypeScript definitions for the most common ones (see `packages/sdk38/src/msgs.ts`).

Cosmos SDK

- [ ] ~~[auth](https://docs.cosmos.network/master/modules/auth/03_messages.html)~~
- [x] [bank](https://docs.cosmos.network/master/modules/bank/03_messages.html)
- [x] [crisis](https://docs.cosmos.network/master/modules/crisis/02_messages.html)
- [x] [distribution](https://docs.cosmos.network/master/modules/distribution/04_messages.html)
- [x] [evidence](https://docs.cosmos.network/master/modules/evidence/03_messages.html)
- [x] [gov](https://docs.cosmos.network/master/modules/gov/03_messages.html)
- [ ] ~~mint~~
- [ ] ~~params~~
- [x] [slashing](https://docs.cosmos.network/master/modules/slashing/03_messages.html)
- [x] [staking](https://docs.cosmos.network/master/modules/staking/03_messages.html)
- [ ] ~~upgrade~~

External

- [x] wasm

## LCD endpoints

[Common endpoints](https://cosmos.network/rpc)

Base endpoints

- [x] /blocks
- [x] /node_info
- [x] /txs (query, broadcast, encode)

Cosmos SDK modules

- [x] /auth (sequence, balance)
- [x] /bank
- [x] /staking (#288)
- [x] /gov (#292)
- [x] /slashing
- [x] /distribution (#295)
- [x] /minting (#287)
- [x] /supply (#280)

External

- [x] /wasm

ethanfrey:
I think there is not ibc in the launchpad release. So maybe that list can be trimmed down a bit. Otherwise looks good.

It would be good to copy all type definitions, but I don't think they all need extensive unit tests. Bank and staking only. 

Maybe later Gov voting (but this has almost no integration tests even in the sdk as it takes so long to run an election with time periods). Just creating a proposal successfully should be enough there to make sure the proposal types are properly formed

willclarktech:
@ethanfrey All the msg types except for IBC were already merged here: https://github.com/CosmWasm/cosmjs/pull/304 (but with low confidence about accuracy because there are no tests).

ethanfrey:
Thanks @willclarktech 

[...]

PR #280, opened 2020-07-07
Create extensible API client

### LINKED ISSUE #277: Ban the term "REST server/client" in this codebase
In Cosmos SDK there is a Light Client Daemon (LCD), i.e. a client to the blockchains that runs on a server providing access to the chain via a JSON over HTTP interface. This interface is often referred to as "REST", probably because of some kind of misconception like REST = JSON over HTTP.

### What REST is

The acronym REST stands for Representational state transfer and is a praradim independent of the document and coding and the transport layer. The combination JSON/HTTP is a typical one but not the only one. The [6 main constraints are](https://en.wikipedia.org/wiki/Representational_state_transfer#Architectural_constraints):
1. Client-server architecture
2. Statelessness
3. Cacheability
4. Layered system
5. Code on demand (optional)
6. Uniform interface

On the web, REST is implemented via HTTP verbs that map to [CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete) in order to manipulate resources identified by URIs.

### Why the light client daemon is not a REST interface

1. Many of the endpoint (such as `POST /staking/delegators/{delegatorAddr}/unbonding_delegations`) assume there is a keypair of the user available on the server. This conflicts with separation of concerns of a client-server architecture, statelessness and a layered system.
2. In RESTful APIs, the URI is a location of a resouce, not an action. Endpoints like `POST /txs/encode` clearly identify an operation, not a resource.
3. Parts of the Tendermint RPC interface is wrapped, but which represets procedure calls and not resources
4. Every writing operation is done by a transaction submission, not by PUTing or PATCHing resources in state. Transactions contain messages, which represent actions, not resources. If you wanted to do a RESTful token transfer, you had to submit an updated source account and an updated destination account instead of submitting the change information.

### REST not needed

Looking at the operations that are done over the current interface, it would make more sense to refer to it as a generic HTTP interface, or _the LCD API_. There is no point in trying to make it a bad REST API if it can be a good HTTP API.

-----------

Given that, _LCD_, _LCD API_ and _LCD client_ are better alternatives.

### LINKED ISSUE #278: sdk38: Allow adding custom modules to LCD client
It should be easy to adding more modules to the LCD client

### LINKED ISSUE #279: Remove defaults from WasmSuccess/WasmResponse
The generic types in

```ts
// Currently all wasm query responses return json-encoded strings...
// later deprecate this and use the specific types for result
// (assuming it is inlined, no second parse needed)
type WasmResponse<T = string> = WasmSuccess<T> | WasmError;

interface WasmSuccess<T = string> {
  readonly height: string;
  readonly result: T;
}
```

should have no defaults. Instead, the type should always be set. The comment is probably outdated and should be removed or replaces with a proper TSDoc comment.

webmaster128:
Closes #277
Closes #278
Closes #279

- [x] Proof of concept
- [x] Use in CosmosClient/SigningCosmosClient
- [x] Export outside of cosmjs/sdk38
- [x] Use in CosmWasmClient/SigningCosmWasmClient


packages/sdk38/src/lcdclient.ts
```diff
@@ -0,0 +1,350 @@
...
+  /** Constructs an LCD client with 3 modules */
+  public static withModules<A extends LcdModule, B extends LcdModule, C extends LcdModule>(
+    options: LcdClientBaseOptions,
+    setupModuleA: LcdModuleSetup<A>,
+    setupModuleB: LcdModuleSetup<B>,
+    setupModuleC: LcdModuleSetup<C>,
+  ): LcdClient & A & B & C;
+
+  /** Constructs an LCD client with 4 modules */
+  public static withModules<
```
ethanfrey, line 93: It wouldn't be easier to just use `withModules(options: LcdModule, modules: ReadonlyArray<LcdModule>)`?

You need to keep the specific type info?

packages/sdk38/src/lcdclient.ts
```diff
@@ -0,0 +1,350 @@
...
+    setupModuleC?: LcdModuleSetup<C>,
+    setupModuleD?: LcdModuleSetup<D>,
+    setupModuleE?: LcdModuleSetup<E>,
+    setupModuleF?: LcdModuleSetup<F>,
+    setupModuleG?: LcdModuleSetup<G>,
+    setupModuleH?: LcdModuleSetup<H>,
+  ): any {
+    const client = new LcdClient(options.apiUrl, options.broadcastMode);
+
+    const modules = new Array<LcdModule>();
```
ethanfrey, line 204: Huh, wouldn't it be simpler to just take the `Array<LcdModule>`? I mean maybe this is nicer to caller, but why so complex?

packages/sdk38/src/lcdclient.ts
```diff
@@ -0,0 +1,350 @@
...
+    const { data } = await this.client.post(path, params).catch(parseAxiosError);
+    if (data === null) {
+      throw new Error("Received null response from server");
+    }
+    return data;
+  }
+
+  // The /auth endpoints
+
+  public async authAccounts(address: string): Promise<AuthAccountsResponse> {
```
ethanfrey, line 268: This is actually a module. Sure, everyone needs it, but interesting to see if this could be unified with the other modules.

packages/sdk38/src/lcdclient.ts
```diff
@@ -0,0 +1,350 @@
...
+    if (setupModuleD) modules.push(setupModuleD(client));
+    if (setupModuleE) modules.push(setupModuleE(client));
+    if (setupModuleF) modules.push(setupModuleF(client));
+    if (setupModuleG) modules.push(setupModuleG(client));
+    if (setupModuleH) modules.push(setupModuleH(client));
+    for (const module of modules) {
+      assert(isNonNullObject(module), `Module must be a non-null object`);
+      for (const key in module) {
+        assert(typeof key == "string", `Found non-string module key: ${key}`);
+        (client as any)[key] = module[key];
```
ethanfrey, line 217: Ah, nice for easy lookup.

Is there an example of how to use this?

`(client as any).wasmd.listCodes()`

ethanfrey: Oh, wait, you basically copy all the functions to the top level. Are they bound to the module still? So I should be able to do

`(client as any).listCodes()` to call the method from the `wasmd` module, or any. Maybe it would be easier to subspace or prefix those to make it more clear. We do use `authAccounts` here not `accounts` to clarify which module is being called.

packages/sdk38/src/lcdclient.ts
webmaster128: The difference is the type safety of the result. This way, the client from `const client = LcdClient.withModules({ apiUrl: wasmd.endpoint }, registerWasmModule, setupSupplyModule);` is a `LcdClient & WasmModule & SupplyModule`, providing all methods for all of them.

packages/sdk38/src/lcdclient.ts
webmaster128: Overloading with different generic types for each argument is required for type safety as mentioned above.

packages/sdk38/src/lcdclient.ts
webmaster128: Yeah good point. It better shows how things are supposed to work

packages/sdk38/src/lcdclient.spec.ts
```diff
@@ -0,0 +1,857 @@
...
+    }
+
+    it("works for no modules", async () => {
+      const client = LcdClient.withModules({ apiUrl: wasmd.endpoint });
+      expect(client).toBeTruthy();
+    });
+
+    it("works for one module", async () => {
+      pendingWithoutWasmd();
+      function wasmClientRegisterer(base: LcdClient): WasmModule {
```
ethanfrey, line 92: This is a good example, and also a demo of how to bind the `LcdClient` so it works when we copy the function to another `this`. I guess we should document to avoid classes in the `LcdModuleSetup`

packages/sdk38/src/lcdclient.spec.ts
```diff
@@ -0,0 +1,857 @@
...
+      expect(codes.length).toBeGreaterThanOrEqual(3);
+      expect(codes[0].id).toEqual(deployedErc20.codeId);
+      expect(codes[0].data_hash).toEqual(deployedErc20.checksum.toUpperCase());
+      expect(codes[0].builder).toEqual(deployedErc20.builder);
+      expect(codes[0].source).toEqual(deployedErc20.source);
+    });
+
+    it("works for two modules", async () => {
+      pendingWithoutWasmd();
+      function registerWasmModule(base: LcdClient): WasmModule {
```
ethanfrey, line 113: Maybe pull this outside of the `it` to avoid repeating?

packages/sdk38/src/lcdclient.spec.ts
```diff
@@ -0,0 +1,857 @@
...
+        registerWasmModule,
+        registerSupplyModule,
+      );
+      const codes = await client.listCodeInfo();
+      expect(codes.length).toBeGreaterThanOrEqual(3);
+      expect(codes[0].id).toEqual(deployedErc20.codeId);
+      expect(codes[0].data_hash).toEqual(deployedErc20.checksum.toUpperCase());
+      expect(codes[0].builder).toEqual(deployedErc20.builder);
+      expect(codes[0].source).toEqual(deployedErc20.source);
+      const supply = await client.totalSupply();
```
ethanfrey, line 149: very nice.

Question - how does the type system handle this?
I assumed it couldn't even know `listCodeInfo`, `totalSupply`, etc were present. Much less that they were types. Looks like you somehow managed that.

packages/sdk38/src/lcdclient.spec.ts
```diff
@@ -0,0 +1,857 @@
...
+      const signBytes2 = makeSignBytes([msg2, msg1], fee, wasmd.chainId, memo, an2, sequence2);
+      const signature1 = await account1.sign(signBytes1);
+      const signature2 = await account2.sign(signBytes2);
+      const signedTx = {
+        msg: [msg2, msg1],
+        fee: fee,
+        memo: memo,
+        signatures: [signature2, signature1],
+      };
+      const postResult = await client.postTx(signedTx);
```
ethanfrey, line 722: Okay, so you can technically post with LcdClient (like rest client), but then we wrap it with the `SigningClient` to make this all much more easy to handle?

packages/sdk38/src/lcdclient.spec.ts
webmaster128: I wanted to support classes but type system disallows it because a class has no index signature. Modules need to conform to `export type LcdModule = Record<string, (...args: any[]) => any>;`, i.e. a map with string keys and function members.

packages/sdk38/src/lcdclient.ts
ethanfrey: Ahhhh... that is your trick to get the types proper. I was wondering what was going on.

packages/sdk38/src/lcdclient.ts
ethanfrey: The examples in tests clarified this for me. THanks

packages/sdk38/src/lcdclient.spec.ts
webmaster128: This is what the overload

```ts
  /** Constructs an LCD client with 2 modules */
  public static withModules<A extends LcdModule, B extends LcdModule>(
    options: LcdClientBaseOptions,
    setupModuleA: LcdModuleSetup<A>,
    setupModuleB: LcdModuleSetup<B>,
  ): LcdClient & A & B;
```

does. For `const client = LcdClient.withModules({ apiUrl: wasmd.endpoint }, registerWasmModule, setupSupplyModule);` we get the type `LcdClient & A & B`, where `A` is the return type of `registerWasmModule` and `B` is the return type of `setupSupplyModule`.

ethanfrey: Answer resolved above. Very nice solution.

packages/sdk38/src/lcdclient.spec.ts
webmaster128: Every LCD client can broadcast. It is a trivial API call open to everyone. The hard part is the signing, which is where the differentiation is. So this is not exactly read-only vs. read-write.

packages/sdk38/src/lcdclient.ts
webmaster128: Implemented in packages/sdk38/src/lcdapi/auth.ts. The basic CosmosClient now uses

```ts
  protected readonly lcdClient: LcdClient & AuthExtension;
```

and


```ts
    this.lcdClient = LcdClient.withExtensions(
      { apiUrl: apiUrl, broadcastMode: broadcastMode },
      setupAuthExtension,
    );
```

packages/sdk38/src/lcdclient.ts
webmaster128: With the latest update, the modules are prefixed (`client.wasm.getContractInfo`, `client.auth.account` etc.)

packages/sdk38/src/lcdapi/lcdclient.ts
```diff
@@ -0,0 +1,331 @@
...
+  public async encodeTx(tx: CosmosSdkTx): Promise<EncodeTxResponse> {
+    const responseData = await this.post("/txs/encode", tx);
+    if (!responseData.tx) {
+      throw new Error("Unexpected response data format");
+    }
+    return responseData as EncodeTxResponse;
+  }
+
+  /**
+   * Broadcasts a signed transaction to into the transaction pool.
```
willclarktech, line 314: Typo "to into"

packages/sdk38/src/lcdapi/lcdclient.spec.ts
```diff
@@ -1,200 +1,199 @@
...
-      const { result } = await client.authAccounts(nonExistentAccount);
-      expect(result).toEqual({
-        type: "cosmos-sdk/Account",
-        value: jasmine.objectContaining({ address: "" }),
-      });
-    });
-  });
 
-  // The /blocks endpoints
+      interface TotalSupplyAllReponse {
```
willclarktech, line 120: Typo: "Response"

packages/sdk38/src/lcdapi/lcdclient.spec.ts
```diff
@@ -1,200 +1,199 @@
...
+            denom: "ustake",
+          },
+        ],
+      });
     });
-  });
 
-  describe("blocks", () => {
-    it("works for block by height", async () => {
+    it("can merge two extension into the same module", async () => {
```
willclarktech, line 164: Typo: "extensions"

packages/sdk38/src/lcdapi/lcdclient.ts
```diff
@@ -0,0 +1,331 @@
...
+
+  public static withExtensions<
+    A extends object,
+    B extends object,
+    C extends object,
+    D extends object,
+    E extends object,
+    F extends object,
+    G extends object,
+    H extends object
```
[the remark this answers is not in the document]
willclarktech: Do we need this many though?

packages/sdk38/src/lcdapi/base.ts
```diff
@@ -0,0 +1,145 @@
...
+  /** Can be empty */
+  readonly evidence_hash: string;
+  readonly proposer_address: string;
+}
+
+interface Block {
+  readonly header: BlockHeader;
+  readonly data: {
+    /** Array of base64 encoded transactions */
+    readonly txs: ReadonlyArray<string> | null;
```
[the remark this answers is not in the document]
webmaster128: Good question. I found the linter rule for that again: https://github.com/CosmWasm/cosmjs/pull/282

packages/sdk38/src/lcdapi/lcdclient.ts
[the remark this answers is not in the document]
webmaster128: Yes, it is a bit of a monster. But it is TypeScript overloading as documented.

Looking at https://github.com/CosmWasm/cosmjs/issues/276 there are 8 modules including CosmWasm. This can easily get more when custom chain modules come in. So I think the overloads are reasonable, even if we don't have examples that big.

packages/sdk38/src/lcdapi/lcdclient.ts
```diff
@@ -0,0 +1,331 @@
...
+    setupExtensionD?: LcdExtensionSetup<D>,
+    setupExtensionE?: LcdExtensionSetup<E>,
+    setupExtensionF?: LcdExtensionSetup<F>,
+    setupExtensionG?: LcdExtensionSetup<G>,
+    setupExtensionH?: LcdExtensionSetup<H>,
+  ): any {
+    const client = new LcdClient(options.apiUrl, options.broadcastMode);
+
+    const extensions = new Array<object>();
+    if (setupExtensionA) extensions.push(setupExtensionA(client));
```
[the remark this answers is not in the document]
webmaster128: We can use the rest operator, which is probably nicer to read:

```ts
  public static withExtensions(
    options: LcdClientBaseOptions,
    ...extensionSetups: LcdExtensionSetup<any>[]
  ): any {
    const client = new LcdClient(options.apiUrl, options.broadcastMode);
    const extensions = extensionSetups.map((setupExtension) => setupExtension(client));
    for (const extension of extensions) {
      assert(isNonNullObject(extension), `Extension must be a non-null object`);
      for (const [moduleKey, moduleValue] of Object.entries(extension)) {
        const current = (client as any)[moduleKey] || {};
        (client as any)[moduleKey] = Object.assign(current, moduleValue);
      }
    }
    return client;
  }
```

packages/sdk38/src/lcdapi/lcdclient.ts
```diff
@@ -0,0 +1,331 @@
...
+    if (setupExtensionD) extensions.push(setupExtensionD(client));
+    if (setupExtensionE) extensions.push(setupExtensionE(client));
+    if (setupExtensionF) extensions.push(setupExtensionF(client));
+    if (setupExtensionG) extensions.push(setupExtensionG(client));
+    if (setupExtensionH) extensions.push(setupExtensionH(client));
+    for (const extension of extensions) {
+      assert(isNonNullObject(extension), `Extension must be a non-null object`);
+      for (const [moduleKey, moduleValue] of Object.entries(extension)) {
+        const current = (client as any)[moduleKey] || {};
+        (client as any)[moduleKey] = Object.assign(current, moduleValue);
```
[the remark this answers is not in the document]
webmaster128: It should be, yes. But I'll add an assertion

packages/sdk38/src/lcdapi/lcdclient.ts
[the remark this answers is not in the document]
willclarktech: Yep, great, then we don't even let people put in `undefined` followed by a valid setup.

packages/sdk38/src/lcdapi/lcdclient.ts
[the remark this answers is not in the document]
willclarktech: Then maybe this is nicer than the potential reassignment of `current`:
```ts
(client as any)[moduleKey] = {
  ...current,
  ...moduleValue,
};
```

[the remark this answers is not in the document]
webmaster128: I messed up reading the docs for Object.assign. Your version is what we want.

packages/sdk38/src/lcdapi/lcdclient.ts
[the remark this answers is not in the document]
willclarktech: Although now we've got a nice `reduce` situation:
```ts
return extensionSetups.reduce((c, setupExtension) => {
  const extension = setupExtension(c);
  assert(isNonNullObject(extension), `Extension must be a non-null object`);
  return Object.entries(extension).reduce((c_, [k, v]) => ({
    ...c_,
    [k]: {
      ...((c_ as any)[k] || {}),
      ...v,
    },
  }), c);
}, client);
```

webmaster128:
I think I addressed all comments in the latest 5 commits

packages/sdk38/src/lcdapi/base.ts
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * The mode used to send transaction
+ *
+ * @see https://cosmos.network/rpc/#/Transactions/post_txs
+ */
+export enum BroadcastMode {
+  /** Return after tx commit */
+  Block = "block",
+  /** Return afer CheckTx */
```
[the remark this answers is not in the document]
willclarktech: This one's still there.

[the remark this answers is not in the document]
webmaster128: Ah, yeah, sorry, thanks. Fixed in updated commit 0c0f015a32d76cfb56b41e4ff00c84cec8191c9e

PR #281, opened 2020-07-07
Fix genesis response type

### LINKED ISSUE #269: Fix genesis response in Tendermint v0.33
A `validators` key is used to specify a set of validators for testnets or PoA style. In PoS, they use the `genutil.gentxs` field to stake and bond a number of validators in the first block. After that is processed, they have a validators set for block 1.

We need to:
- Make `validators` optional.
- Parse `genutil.gentxs` into a separate field

willclarktech:
Closes #269

Do we need to do something specific with `genutil.gentxs`? Right now it's shoved in `appState` along
with everything else.


ethanfrey:
I'd leave it as is in appState.
If anyone wants to inspect it, (like any of the other module stuff), they should know what they are doing. There is no clear format for it.

PR #282, opened 2020-07-07
Add array-type linter configuration

webmaster128:


packages/faucet/src/faucet.ts
```diff
@@ -38,7 +38,7 @@ export class Faucet {
     apiUrl: string,
     addressPrefix: string,
     config: TokenConfiguration,
-    pens: readonly [string, Pen][],
+    pens: ReadonlyArray<[string, Pen]>,
```
willclarktech, line 41: Now can we have `ReadonlyArray<readonly [string, Pen]>`?

.eslintrc.js
```diff
@@ -27,10 +27,11 @@ module.exports = {
     "no-shadow": "warn",
     "no-unused-vars": "off", // disabled in favour of @typescript-eslint/no-unused-vars, see https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars.md
     "prefer-const": "warn",
-    "radix": ["warn", "always"],
+    radix: ["warn", "always"],
     "spaced-comment": ["warn", "always", { line: { markers: ["/ <reference"] } }],
     "import/no-cycle": "warn",
     "simple-import-sort/sort": "warn",
+    "@typescript-eslint/array-type": ["warn", { default: "array-simple" }],
```
willclarktech, line 34: This seems like the most confusing configuration, but I'm OK with it.

packages/faucet/src/faucet.ts
webmaster128: Indeed, we can

.eslintrc.js
webmaster128: I don't have a strong opinion here. This was used in the React projects if I recall correctly and worked well.

PR #284, opened 2020-07-08
Define Wallet interface and implement for Secp256k1Wallet

### LINKED ISSUE #264: Define common Wallet interface
Figure out a common API to represent local keys, chrome extensions that can sign, or ledger nano. Anything that can sign our transactions for us.

It would be good to talk with @thunnini about his design from Keplr / CosmosJS, he was interested in using this with CosmJS

### LINKED ISSUE #265: Implement Wallet interface with Secp256k1Pen
Depends on #264 

We have a simple singer for demo code, adapt this to fit in the generic Wallet interface, so we can use this interface in all test code and demos.

willclarktech:
Closes #264
Closes #265

Opening as a draft to check I'm going in the right direction here. Questions from my end:
- What to do about the transaction builder context (fees, gas, etc)?
- How to deal with the awkwardness of switching from a `StdSignature` as used by most of our
  codebase to a `Uint8Array`?
- I kept `address` and `pubkey` as public properties on the `Secp256k1Wallet` class to avoid lots of
  tedious refactors, but should they be private in this case?
- Should we include prehashing in the `sign` method or require the app to pass a prehashed message?
- Should `algo` in `AccountData` really be a `string` or should it be an `enum`/union of valid strings?
- I made `getAccounts` and `sign` depend on the wallet having been enabled to match the logic of eg a browser extension, but I guess this should just be removed for this class?


webmaster128:
> How to deal with the awkwardness of switching from a `StdSignature` as used by most of our
> codebase to a `Uint8Array`?

A Uint8Array signature contains no informations about who signed it and which algorithm was used. I think `sign` should return `StdSignature`.

> I kept `address` and `pubkey` as public properties on the `Secp256k1Wallet` class to avoid lots of
> tedious refactors, but should they be private in this case?

There cannot be a direct mapping between a wallet and an address, since a wallet can store multiple addresses, right?

> Should we include prehashing in the `sign` method or require the app to pass a prehashed message?

The wallet needs to be able to show the transactin to users before signing. So is probably best encoded as a (JSON) string and andcoded/prehased during signing.

> I made `getAccounts` and `sign` depend on the wallet having been enabled to match the logic of eg a browser extension, but I guess this should just be removed for this class?

The implementation is good. It will enfore the app developer to access the Wallet interface correctly during development.

willclarktech:
> A Uint8Array signature contains no informations about who signed it and which algorithm was used. I think `sign` should return `StdSignature`.

I'd be in favour of that, although if an app asks the wallet to sign then it needs to know the address, which presumably comes from asking the wallet, in which case the app also knows the public key, so could construct the `StdSignature` (or equivalent) outside the wallet.

> There cannot be a direct mapping between a wallet and an address, since a wallet can store multiple addresses, right?

In general yes, I guess I'm unclear on what this particular wallet is supposed to be for.

> The wallet needs to be able to show the transactin to users before signing. So is probably best encoded as a (JSON) string and andcoded/prehased during signing.

:+1:

> The implementation is good. It will enfore the app developer to access the Wallet interface correctly during development.

:+1:

ethanfrey:
I agree with the points Simo n made above, in addition to returning a `StdSignature`. We are Cosmos-only and there is one way to produce a signature on all cosmos-sdk apps, so that is helpful.

> What to do about the transaction builder context (fees, gas, etc)?

That would only be interesting for an `OnlineSigner` / `OnlineWallet`. It is a very important discussion and allowing the Wallet to complete the transaction (as well as submit it), but I think we can push that to another PR. 

I would rename your current interface to `OfflineWallet` as it has no access to the network and just signs bytes. It makes it clear that `txBuilder` has no place there - it needs a complete transaction.  Local key storage as well as Ledger are all `OfflineWallets`. a chrome extension would be an `OnlineWallet` (or possibly both), and we can make a local wrapper that takes an `OfflineWallet` and a `LcdClient` to fulfill the `OnlineWallet` interface - both as a demo and as a default provider for `CosmWasmSigner` (or whatever that high-level class gets renamed to).

ethanfrey:
> The wallet needs to be able to show the transactin to users before signing. So is probably best encoded as a (JSON) string and andcoded/prehased during signing.

Yes, it should be visible. In Cosmos SDK, the "sign bytes" are designed to be "human readable"(??) or at least contain enough type info that a UI should be able to produce a reasonable representation of them to a human. Pre-stargate (sdk38), this is a JSON string. Post-stargate this may also be a protobuf object using Any - which means, a program could download the type definitions and figure out how to display it (don't worry about this for now... that is another epic)

ethanfrey:
>>  A Uint8Array signature contains no informations about who signed it and which algorithm was used. I think sign should return StdSignature.

> I'd be in favour of that, although if an app asks the wallet to sign then it needs to know the address, which presumably comes from asking the wallet, in which case the app also knows the public key, so could construct the StdSignature (or equivalent) outside the wallet.

One note - the StdSignature should include the public key if the account_sequence is 0. I'm not sure if it needs the address as well, or only when the pubkey is not present. We can try without and see if it fails. But let's make sure never to include pubkey in StdSignature if we don't need to... then the Wallet takes care of one more piece of logic that the caller doesn't have to think about.

packages/sdk38/src/wallet.ts
```diff
@@ -0,0 +1,131 @@
...
+  return [
+    Slip10RawIndex.hardened(44),
+    Slip10RawIndex.hardened(118),
+    Slip10RawIndex.hardened(0),
+    Slip10RawIndex.normal(0),
+    Slip10RawIndex.normal(a),
+  ];
+}
+
+export class Secp256k1Wallet implements Wallet {
```
webmaster128, line 70: Having this as a 1-address test implementation is good as long as the interfaces works for multiple addresses

packages/sdk38/src/wallet.ts
```diff
@@ -0,0 +1,131 @@
...
+  readonly address: string;
+  readonly algo: Algo;
+  readonly pubkey: Uint8Array;
+}
+
+export interface Wallet {
+  /**
+   * Request access to the user's accounts. Wallet should ask the user to approve or deny access. Returns true if granted access or false if denied.
+   */
+  readonly enable: () => Promise<boolean>;
```
webmaster128, line 30: I wonder if a wallet can be disabled over time. In this case, the caller needs to call `.enable()` before every wallet interaction. In this case, the method gets pointless as the mechanism could and should be integrated in `getAccounts` and `sign`.

willclarktech: 🤔

packages/sdk38/src/cosmosclient.spec.ts
```diff
@@ -193,7 +193,9 @@ describe("CosmosClient", () => {
   describe("postTx", () => {
     it("works", async () => {
       pendingWithoutWasmd();
-      const pen = await Secp256k1Pen.fromMnemonic(faucet.mnemonic);
+      const wallet = await Secp256k1OfflineWallet.fromMnemonic(faucet.mnemonic);
+      const accounts = await wallet.getAccounts();
+      const { address: walletAddress } = accounts[0];
```
webmaster128, line 198: Ever tried nested array and object destruction?

```ts
const [{ address: walletAddress }] = await wallet.getAccounts();
```

No idea if that works and is still readable.

packages/sdk38/src/wallet.ts
```diff
@@ -0,0 +1,131 @@
...
+      },
+    ];
+  }
+
+  public async sign(
+    address: string,
+    message: Uint8Array,
+    prehashType: PrehashType = "sha256",
+  ): Promise<StdSignature> {
+    if (!this.enabled && !(await this.enable())) {
```
webmaster128, line 120: The wallet enabling itself is not so useful, right? If we don't have it in the interface anymore, I think we can remove `this.enabled` completely.

packages/sdk38/src/cosmosclient.spec.ts
willclarktech: I like this and it works. :)

packages/sdk38/src/wallet.ts
willclarktech: Yeah, I guess I was trying to showcase how the flow of a more user-interactive wallet might look, but it's not necessary for this simple one.

willclarktech: Or I guess somebody could extend this class and override the `enable` method.

packages/cosmwasm/src/signingcosmwasmclient.ts
```diff
@@ -175,22 +176,22 @@ export class SigningCosmWasmClient extends CosmWasmClient {
    *
    * @param apiUrl The URL of a Cosmos SDK light client daemon API (sometimes called REST server or REST API)
    * @param senderAddress The address that will sign and send transactions using this instance
-   * @param signCallback An asynchonous callback to create a signature for a given transaction. This can be implemented using secure key stores that require user interaction.
+   * @param signer A wallet provider which can provide signatures for transactions, potentially requiring user input.
```
webmaster128, line 179: "A wallet provider" -> "An implementation of OfflineSigner" (or something like that)

packages/sdk38/src/signingcosmosclient.ts
```diff
@@ -49,22 +46,22 @@ export class SigningCosmosClient extends CosmosClient {
    *
    * @param apiUrl The URL of a Cosmos SDK light client daemon API (sometimes called REST server or REST API)
    * @param senderAddress The address that will sign and send transactions using this instance
-   * @param signCallback An asynchonous callback to create a signature for a given transaction. This can be implemented using secure key stores that require user interaction.
+   * @param signer A wallet provider which can provide signatures for transactions, potentially requiring user input.
```
webmaster128, line 49: As above

packages/sdk38/src/wallet.ts
```diff
@@ -0,0 +1,116 @@
...
+  return [
+    Slip10RawIndex.hardened(44),
+    Slip10RawIndex.hardened(118),
+    Slip10RawIndex.hardened(0),
+    Slip10RawIndex.normal(0),
+    Slip10RawIndex.normal(a),
+  ];
+}
+
+export class Secp256k1OfflineWallet implements OfflineSigner {
```
webmaster128, line 66: I'd simplify this to `class Secp256k1Wallet implements OfflineSigner`. What do you think?

packages/sdk38/src/wallet.ts
```diff
@@ -0,0 +1,116 @@
...
+    Slip10RawIndex.hardened(0),
+    Slip10RawIndex.normal(0),
+    Slip10RawIndex.normal(a),
+  ];
+}
+
+export class Secp256k1Wallet implements OfflineSigner {
+  public static async fromMnemonic(
+    mnemonic: string,
+    hdPath: readonly Slip10RawIndex[] = makeCosmoshubPath(0),
```
ethanfrey, line 69: Ah, this is clever.

I was wondering how `fromMnemonic` always had a key available, and how to set what it was. I like the optional arguments

PR #285, opened 2020-07-09
Add bank extension

webmaster128:
For #276

### DIFF
--- a/packages/sdk38/src/index.ts
+++ b/packages/sdk38/src/index.ts
@@ -23,6 +23,8 @@ export { makeSignBytes } from "./encoding";
 export {
   AuthAccountsResponse,
   AuthExtension,
+  BankBalancesResponse,
+  BankExtension,
   BlockResponse,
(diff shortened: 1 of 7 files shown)


webmaster128:
> but are we not adding the other functions?

Which other functions? Those as all queries for the bank module. We don't connect POST endpoints as those require keypairs on the REST server. Instead, all writing operations are done via tx submission.

willclarktech:
What about `total`?

https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/client/rest/rest.go

webmaster128:
> What about `total`?

Ah, interesting. I took the ones from the documentation at https://cosmos.network/rpc/v0.38.3. Turns out in 0.38 there is only one query: https://github.com/cosmos/cosmos-sdk/blob/v0.38.5/x/bank/client/rest/tx.go#L17-L20. Looks like the `/supply` queries moves to bank on master.

PR #288, opened 2020-07-09
Add StakingExtension

webmaster128:
For #276


packages/sdk38/src/lcdapi/staking.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+              },
+              update_time: "2020-06-03T06:01:17.4747987Z",
+            },
+            min_self_delegation: "1",
+          },
+        ],
+      });
+    });
+
+    it("can filter by status", async () => {
```
willclarktech, line 50: It would be nice to have a case where something matches the filter.

packages/sdk38/src/lcdapi/staking.ts
```diff
@@ -0,0 +1,122 @@
...
+    readonly max_entries: number;
+    readonly historical_entries: number;
+    readonly bond_denom: string;
+  };
+}
+
+export interface StakingExtension {
+  readonly staking: {
+    // Get all delegations from a delegator
+    // /staking/delegators/{delegatorAddr}/delegations
```
willclarktech, line 69: What's the plan with these endpoints?

webmaster128: We need to implement all of them. Lots of boring work ahead …

packages/sdk38/src/lcdapi/staking.spec.ts
```diff
@@ -0,0 +1,508 @@
+import { assert, sleep } from "@cosmjs/utils";
+
+import { coin, coins } from "../coins";
+import { isPostTxFailure } from "../cosmosclient";
+import { makeSignBytes } from "../encoding";
+import { MsgDelegate, MsgUndelegate } from "../msgs";
+import { SigningCosmosClient } from "../signingcosmosclient";
+/* eslint-disable @typescript-eslint/naming-convention */
```
webmaster128, line 8: Why is this here in the middle of the file?

packages/sdk38/src/lcdapi/staking.spec.ts
```diff
@@ -0,0 +1,508 @@
...
+    it("works", async () => {
+      pendingWithoutWasmd();
+      const client = makeStakingClient(wasmd.endpoint);
+      const response = await client.staking.delegatorDelegations(faucet.address);
+      expect(response).toEqual({
+        height: jasmine.stringMatching(nonNegativeIntegerMatcher),
+        result: [
+          {
+            delegator_address: "cosmos1pkptre7fdkl6gfrzlesjjvhxhlc3r4gmmk8rs6",
+            validator_address: "cosmosvaloper1gjvanqxc774u6ed9thj4gpn9gj5zus5u32enqn",
```
webmaster128, line 100: We can use this from `validatorAddress` in `packages/sdk38/src/testutils.spec.ts`

packages/sdk38/src/lcdapi/staking.spec.ts
```diff
@@ -0,0 +1,508 @@
...
+  describe("delegatorUnbondingDelegations", () => {
+    it("works", async () => {
+      pendingWithoutWasmd();
+      const client = makeStakingClient(wasmd.endpoint);
+      const response = await client.staking.delegatorUnbondingDelegations(faucet.address);
+      expect(response.height).toMatch(nonNegativeIntegerMatcher);
+      expect(response.result.length).toEqual(1);
+      expect(response.result[0].delegator_address).toEqual("cosmos1pkptre7fdkl6gfrzlesjjvhxhlc3r4gmmk8rs6");
+      expect(response.result[0].validator_address).toEqual(
+        "cosmosvaloper1gjvanqxc774u6ed9thj4gpn9gj5zus5u32enqn",
```
webmaster128, line 118: `validatorAddress`

packages/sdk38/src/lcdapi/staking.spec.ts
```diff
@@ -0,0 +1,508 @@
...
+      });
+    });
+  });
+
+  describe("delegatorTransactions", () => {
+    it("works", async () => {
+      pendingWithoutWasmd();
+      const client = makeStakingClient(wasmd.endpoint);
+      const response = await client.staking.delegatorTransactions(faucet.address);
+      expect(response.length).toEqual(3);
```
webmaster128, line 135: Is this stable when you run the tests multiple times?

packages/sdk38/src/lcdapi/staking.spec.ts
```diff
@@ -0,0 +1,508 @@
...
+      const response = await client.staking.delegatorValidators(faucet.address);
+      expect(response).toEqual({
+        height: jasmine.stringMatching(nonNegativeIntegerMatcher),
+        result: [
+          {
+            operator_address: "cosmosvaloper1gjvanqxc774u6ed9thj4gpn9gj5zus5u32enqn",
+            consensus_pubkey:
+              "cosmosvalconspub1zcjduepqau36ht2r742jh230pxlu4wjmwcmkwpeqava80acphsu87vt5xlpqx6g7qh",
+            jailed: false,
+            status: 2,
```
webmaster128, line 152: Can we create an enum with exlicit raw values for those numeric status representations?

packages/sdk38/src/lcdapi/staking.spec.ts
```diff
@@ -0,0 +1,508 @@
...
+        },
+      });
+    });
+  });
+
+  describe("unbondingDelegation", () => {
+    it("works", async () => {
+      pendingWithoutWasmd();
+      const client = makeStakingClient(wasmd.endpoint);
+      const response = await client.staking.unbondingDelegation(faucet.address, validatorAddress);
```
webmaster128, line 238: if you want you can write

```ts
const { height, result } = await client.staking.unbondingDelegation(faucet.address, validatorAddress);
assert(result);
// now test result, which is known to be non-null
```

packages/sdk38/src/lcdapi/staking.spec.ts
```diff
@@ -0,0 +1,508 @@
...
+        completion_time: jasmine.stringMatching(dateTimeStampMatcher),
+        initial_balance: "100",
+        balance: "100",
+      });
+    });
+  });
+
+  describe("redelegations", () => {
+    it("works", async () => {
+      // TODO: Set up a result for this test
```
webmaster128, line 256: 👍 No idea what a redelegation is …

packages/sdk38/src/lcdapi/staking.ts
```diff
@@ -0,0 +1,245 @@
...
+      delegation: async (delegatorAddress: string, validatorAddress: string) =>
+        base.get(`/staking/delegators/${delegatorAddress}/delegations/${validatorAddress}`),
+      unbondingDelegation: async (delegatorAddress: string, validatorAddress: string) =>
+        base.get(`/staking/delegators/${delegatorAddress}/unbonding_delegations/${validatorAddress}`),
+      redelegations: async () => base.get(`/staking/redelegations`),
+      validators: async (params?: StakingValidatorsParams) => base.get(`/staking/validators`, params),
+      validator: async (validatorAddress: string) => base.get(`/staking/validators/${validatorAddress}`),
+      validatorDelegations: async (validatorAddress: string) =>
+        base.get(`/staking/validators/${validatorAddress}/delegations
+      `),
```
webmaster128, line 237: here is a stray line break

packages/sdk38/src/lcdapi/staking.spec.ts
willclarktech: Ah sorry, this is from my VSCode setup: tab completion will automatically add an import statement, but has no respect for eslint.

packages/sdk38/src/lcdapi/staking.spec.ts
willclarktech: Yes but I don't yet understand why.

packages/sdk38/src/lcdapi/staking.spec.ts
willclarktech: https://docs.cosmos.network/master/modules/staking/02_state_transitions.html#begin-redelegation

Basically an unbond plus a delegation in one step.

packages/sdk38/src/lcdapi/staking.spec.ts
willclarktech: Since this appears in the results when we filter for "bonded", I'm assuming the status should be
```
Unbonded: 0
Unbonding: 1
Bonded: 2
```
Even though in the Go code it's 1–3.

webmaster128: 0, 1, 2 is what I found last time as well: https://github.com/cosmos/cosmos-sdk/blob/v0.38.5/types/staking.go#L43-L49

willclarktech: Ah ok then we may need to update this later. See current master: https://github.com/cosmos/cosmos-sdk/blob/53bf227/types/staking.go

webmaster128: Oh, and we already have

```ts
/**
 * Numeric bonding status
 *
 * @see https://github.com/cosmos/cosmos-sdk/blob/v0.38.5/types/staking.go#L43-L49
 */
export enum BondStatus {
  Unbonded = 0,
  Unbonding = 1,
  Bonded = 2,
}
```

webmaster128: Yeah this update is good since protobuf does not differentiate between zero and unset

PR #295, opened 2020-07-15
Add LCD support for distribution module

willclarktech:
Part of #276


packages/sdk38/src/lcdapi/distribution.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    gas: "1500000", // 1.5 million
+  };
+
+  beforeAll(async () => {
+    if (wasmdEnabled()) {
+      const wallet = await Secp256k1Wallet.fromMnemonic(faucet.mnemonic);
+      const client = new SigningCosmosClient(wasmd.endpoint, faucet.address, wallet, {});
+
+      const chainId = await client.getChainId();
+      const msg = {
```
webmaster128, line 38: For this one we even have a type: `const msg: MsgDelegate = {`

packages/sdk38/src/lcdapi/distribution.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+  describe("validator", () => {
+    it("works", async () => {
+      pendingWithoutWasmd();
+      const client = makeDistributionClient(wasmd.endpoint);
+      const response = await client.distribution.validator(validatorAddress);
+      expect(response).toEqual({
+        height: jasmine.stringMatching(nonNegativeIntegerMatcher),
+        result: {
+          // TODO: This smells like a bug in the backend to me
+          operator_address: Bech32.encode("cosmos", Bech32.decode(validatorAddress).data),
```
webmaster128, line 119: @ethanfrey? 👆

webmaster128:
@willclarktech can you update this (and merge when green)?

packages/sdk38/src/lcdapi/distribution.spec.ts
ethanfrey: Yes this does seem like they use the wrong bech32 prefix here.

PR #305, opened 2020-07-21
Use pbkdf2 Uint8Array support

### LINKED ISSUE #219: Add Uint8Array support in @types/pbkdf2
In https://github.com/crypto-browserify/pbkdf2/pull/81, Uint8Array support was added to the npm package pbkdf2. Unfortunately we cannot use it yet because the new type is not available in @types/pbkdf2.

So we should create a PR to DefinitelyTyped to generalize the input types for `password` and `salt`.

In theory the types should be the same as in https://nodejs.org/api/crypto.html#crypto_crypto_pbkdf2_password_salt_iterations_keylen_digest_callback and https://nodejs.org/api/crypto.html#crypto_crypto_pbkdf2sync_password_salt_iterations_keylen_digest, but the mapping to TypeScript types might be a bit tricky. As far as I can see, the only relevant input types are `string | Buffer | Uint8Array`.

If this is done, we can remove the Buffers from

```ts
  // convert pbkdf2's calllback interface to Promise interface
  private static async pbkdf2(
    secret: Uint8Array,
    salt: Uint8Array,
    iterations: number,
    keylen: number,
    digest: string,
  ): Promise<Uint8Array> {
    return new Promise<any>((resolve, reject) => {
      // TODO: Patch @types/pbkdf2 to allow Uint8Array as secret and salt argument
      pbkdf2(Buffer.from(secret), Buffer.from(salt), iterations, keylen, digest, (err, derivedKey) => {
        if (err) {
          reject(err);
        } else {
          resolve(new Uint8Array(derivedKey));
        }
      });
    });
  }
```

willclarktech:
Closes #219

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1327,10 +1327,10 @@
   resolved "https://registry.yarnpkg.com/@types/pako/-/pako-1.0.1.tgz#33b237f3c9aff44d0f82fe63acffa4a365ef4a61"
   integrity sha512-GdZbRSJ3Cv5fiwT6I0SQ3ckeN2PWNqxd26W9Z2fCK1tGrrasGy4puvNFtnddqH9UJFMQYXxEuuB7B8UK+LLwSg==
 
-"@types/pbkdf2@^3.0.0":
-  version "3.0.0"
-  resolved "https://registry.yarnpkg.com/@types/pbkdf2/-/pbkdf2-3.0.0.tgz#5d9ca5f12a78a08cc89ad72883ad4a30af359229"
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -6560,7 +6560,7 @@ path-type@^3.0.0:
   dependencies:
     pify "^3.0.0"
 
-pbkdf2@^3.0.16, pbkdf2@^3.0.3, pbkdf2@^3.0.9:
+pbkdf2@^3.0.3, pbkdf2@^3.0.9:
```
webmaster128, line 6563: We could get rid of the two pbkdf2 installations by removing this block and letting yarn repair it with `yarn install`

Issue #266, opened 2020-07-01
Implement secure local storage for software wallet


ethanfrey:
Depends on #264 

* Encrypt and store key material in local storage (both browser and node) with a secure algorithm.
* Allow handling multiple accounts.
* Implement the Wallet interface

Inspiration:

* Some tested code to [handle such encryption](https://github.com/iov-one/iov-core/blob/master/packages/iov-keycontrol/src/keyringencryptor.ts)
* And an [implementation of keyring](https://github.com/iov-one/iov-core/blob/master/packages/iov-keycontrol/src/keyring.ts) to hold multiple keys

webmaster128:
I don't think we need all the keyring overhead anytime soon. If we stick with one mnemonic per Wallet, we can still have as many accounts as we want in arbitrary dimensions with HD paths. If an application really needs multiple wallets, it can encrypt them individually.

ethanfrey:
I agree.

* One mnemonic. 
* Multiple derivation paths.
* Mnemonic encrypted securely.
* Plugable storage backend (node and browser implementations provided)

How do we handle multiple chains? I assume each has a different bech32 prefix and possibly a different derivation path. 

Shall we add `createAccount(bech32prefix, derivationPath)` to the object, not the interface, so I can record all available keys?

webmaster128:
> Plugable storage backend (node and browser implementations provided)

Why would we care about storage at all as long as we're not building a wallet? It results in annoying platform-dependent dependencies. We could allow encrypted serialization and leave storage to the entity that controlls storage.

> How do we handle multiple chains? I assume each has a different bech32 prefix and possibly a different derivation path.
> 
> Shall we add `createAccount(bech32prefix, derivationPath)` to the object, not the interface, so I can record all available keys?

Sounds good.

ethanfrey:
> Why would we care about storage at all as long as we're not building a wallet? It results in annoying platform-dependent dependencies. We could allow encrypted serialization and leave storage to the entity that controlls storage.

I think we have two choices:

* generate a mnemonic and return it password encrypted to the caller + accept an encrypted mnemonic and password in the constructor. This makes us completely agnostic to how it is stored, but take all responsibility for the data that will be stored.
* accept a storage interface and provide save/load methods that do the load and decrypt or such

I assume you prefer the first, which has the bonus of us not trying to define some generic storage interface. Can you please define such interfaces?

ethanfrey:
I would propose a class something like this:

```ts
[...]

PR #310, opened 2020-07-22
Add wallet serialization/deserialization

webmaster128:
Closes https://github.com/CosmWasm/cosmjs/issues/266

- [x] Move symmetric encryption from `packages/sdk38/src/secp256k1wallet.ts` to `packages/sdk38/src/wallet.ts`


packages/sdk38/src/testutils.spec.ts
```diff
@@ -19,6 +19,9 @@ export const tendermintAddressMatcher = /^[0-9A-F]{40}$/;
 export const tendermintShortHashMatcher = /^[0-9a-f]{40}$/;
 export const dateTimeStampMatcher = /^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}(?:\.[0-9]+)?Z$/;
 export const semverMatcher = /^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/;
+/** @see https://rgxdb.com/r/1NUN74O6 */
+export const base64Matcher = /^(?:[a-zA-Z0-9+/]{4})*(?:|(?:[a-zA-Z0-9+/]{3}=)|(?:[a-zA-Z0-9+/]{2}==)|(?:[a-zA-Z0-9+/]{1}===))$/;
```
willclarktech, line 23: Nice, but isn't padding technically optional in base64?

packages/sdk38/src/secp256k1wallet.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+    it("can use different mnemonic lengths", async () => {
+      expect((await Secp256k1Wallet.generate(12)).mnemonic.split(" ").length).toEqual(12);
+      expect((await Secp256k1Wallet.generate(15)).mnemonic.split(" ").length).toEqual(15);
+      expect((await Secp256k1Wallet.generate(18)).mnemonic.split(" ").length).toEqual(18);
+      expect((await Secp256k1Wallet.generate(21)).mnemonic.split(" ").length).toEqual(21);
+      expect((await Secp256k1Wallet.generate(24)).mnemonic.split(" ").length).toEqual(24);
+    });
+  });
+
+  describe("deserialize", () => {
```
willclarktech, line 38: How about `serialize and deserialize`? Or have a test from a known serialization for `deserialize` and explicitly name this an integration test.

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+
+  switch ((root as any).type) {
+    case serializationTypeV1:
+      return extractKdfConfigurationV1(root);
+    default:
+      throw new Error("Unsupported serialization type");
+  }
+}
+
+export class Secp256k1Wallet implements OfflineSigner {
```
ethanfrey, line 107: Okay, so one class for the basic usage (provide mnemonic) and for the complex usage (encrypted storage). Let me see if this does make sense to combine.

packages/sdk38/src/secp256k1wallet.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+        {
+          algo: "secp256k1",
+          address: defaultAddress,
+          pubkey: defaultPubkey,
+        },
+      ]);
+    });
+  });
+
+  describe("deserializeWithEncryptionKey", () => {
```
willclarktech, line 55: `serializeWithEncryptionKey and deserializeWithEncryptionKey`?

packages/sdk38/src/secp256k1wallet.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+      const accounts = await wallet.getAccounts();
+      expect(accounts.length).toEqual(1);
+      expect(accounts[0]).toEqual({
+        address: defaultAddress,
+        algo: "secp256k1",
+        pubkey: defaultPubkey,
+      });
+    });
+
+    it("creates the same address as Go implementation", async () => {
```
willclarktech, line 101: It would be nice to have a link to the source here. (I realise this was here before.)

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+    hdPath: readonly Slip10RawIndex[] = makeCosmoshubPath(0),
+    prefix = "cosmos",
+  ): Promise<Secp256k1Wallet> {
+    const entropyLength = 4 * Math.floor((11 * length) / 33);
+    const entropy = Random.getBytes(entropyLength);
+    const mnemonic = Bip39.encode(entropy);
+    return Secp256k1Wallet.fromMnemonic(mnemonic.toString(), hdPath, prefix);
+  }
+
+  public static async deserialize(serialization: string, password: string): Promise<Secp256k1Wallet> {
```
ethanfrey, line 151: The interplay is a bit complex, but in short `deserialize` will do the kdf inside of the call (easier) and `deserailizeWithEncryptionKey` will assume the kdf is already performed (in another thread/web worker
as to not block the UI).

Nice to provide both options. It would be good to add comments explaining this choice and when to use each.

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+        algo: "secp256k1",
+        hdPath: hdPath,
+        prefix: prefix,
+      },
+    ];
+    this.privkey = privkey;
+    this.pubkey = pubkey;
+  }
+
+  public get mnemonic(): string {
```
ethanfrey, line 250: Useful for the simple case that doesn't care about secure storage or much else.
`generate`, `mnemonic`, store somewhere unecypted and good enough for unsecure demo app.

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+    return this.secret.toString();
+  }
+
+  private get address(): string {
+    return rawSecp256k1PubkeyToAddress(this.pubkey, this.accounts[0].prefix);
```
ethanfrey, line 255: Why special case for `this.accounts[0]`?

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  public async getAccounts(): Promise<readonly AccountData[]> {
+    return [
```
ethanfrey, line 259: Ah... we only allow one account currently, but code is written to be extensible. :+1:

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+   * @param password The user provided password used to generate an encryption key via a KDF.
+   *                 This is not normalized internally (see "Unicode normalization" to learn more).
+   */
+  public async serialize(password: string): Promise<string> {
+    const kdfConfiguration = basicPasswordHashingOptions;
+    const encryptionKey = await executeKdf(password, kdfConfiguration);
+    return this.serializeWithEncryptionKey(encryptionKey, kdfConfiguration);
+  }
+
+  /**
```
ethanfrey, line 294: Nice comments here for the two variants. I would put similar ones on the `static deserialize` methods

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+} from "./wallet";
+
+const serializationTypeV1 = "secp256k1wallet-v1";
+
+/**
+ * A fixed salt is chosen to archive a deterministic password to key derivation.
+ * This reduces the scope of a potential rainbow attack to all CosmJS users.
+ * Must be 16 bytes due to implementation limitations.
+ */
+const cosmjsSalt = toAscii("The CosmJS salt.");
```
willclarktech, line 39: Why not import this from `wallet.ts`?

packages/sdk38/types/wallet.d.ts
```diff
@@ -17,23 +17,18 @@ export interface OfflineSigner {
...
-  private get address();
-  getAccounts(): Promise<readonly AccountData[]>;
-  sign(address: string, message: Uint8Array, prehashType?: PrehashType): Promise<StdSignature>;
+export interface KdfConfiguration {
+  /**
+   * An algorithm identifier, such as "argon2id" or "scrypt".
+   */
+  readonly algorithm: string;
+  /** A map of algorithm-specific parameters */
+  readonly params: Record<string, unknown>;
```
ethanfrey, line 32: Okay, so this is left quite open rather than strictly typed for each algorithm we support.

I think we will not add many algorithms and strong-typing here (union type with different set of params based on the algorithm literal string) would be useful when using this (both as documentation as well as auto-correct in the IDE)

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+const algorithmIdXchacha20poly1305Ietf = "xchacha20poly1305-ietf";
+
+/**
+ * This interface describes a JSON object holding the encrypted wallet and the meta data.
+ * All fields in here must be JSON types.
+ */
+export interface Secp256k1WalletSerialization {
+  /** A format+version identifier for this serialization format */
+  readonly type: string;
+  /** Information about the key derivation function (i.e. password to encrytion key) */
```
willclarktech, line 63: Typo: "encryption"

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+export interface Secp256k1WalletData {
+  readonly mnemonic: string;
+  readonly accounts: ReadonlyArray<{
+    readonly algo: string;
+    readonly hdPath: string;
+    readonly prefix: string;
+  }>;
+}
+
+function extractKdfConfigurationV1(document: any): KdfConfiguration {
```
willclarktech, line 91: `document` is a global in some environments

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+   * @param length The number of words in the mnemonic (12, 15, 18, 21 or 24).
+   * @param hdPath The BIP-32/SLIP-10 derivation path. Defaults to the Cosmos Hub/ATOM path `m/44'/118'/0'/0/0`.
+   * @param prefix The bech32 address prefix (human readable part). Defaults to "cosmos".
+   */
+  public static async generate(
+    length: 12 | 15 | 18 | 21 | 24 = 12,
+    hdPath: readonly Slip10RawIndex[] = makeCosmoshubPath(0),
+    prefix = "cosmos",
+  ): Promise<Secp256k1Wallet> {
+    const entropyLength = 4 * Math.floor((11 * length) / 33);
```
willclarktech, line 145: ??

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+        assert(typeof prefix === "string");
+
+        return Secp256k1Wallet.fromMnemonic(mnemonic, stringToPath(hdPath), prefix);
+      }
+      default:
+        throw new Error("Unsupported serialization type");
+    }
+  }
+
+  private static async deserializeType1(serialization: string, password: string): Promise<Secp256k1Wallet> {
```
willclarktech, line 196: `deserializeTypeV1`?

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+    const root = JSON.parse(serialization);
+    if (!isNonNullObject(root)) throw new Error("Root document is not an object.");
+    const untypedRoot: any = root;
+    switch (untypedRoot.type) {
```
willclarktech, line 200: I think this makes more sense than a switch in this method:
```ts
if (untypedRoot.type !== serializationTypeV1) {
    throw new Error("Unexpected serialization type");
}
```

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+      case serializationTypeV1: {
+        let encryptionKey: Uint8Array;
```
willclarktech, line 202: Hmm, can we assign something a default empty `Uint8Array` or allow this type to be undefined?

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+        const decryptedBytes = await Xchacha20poly1305Ietf.decrypt(
+          fromBase64(untypedRoot.data),
+          encryptionKey,
+          nonce,
+        );
+        const decryptedDocument = JSON.parse(fromUtf8(decryptedBytes));
+        const { mnemonic, accounts } = decryptedDocument;
+        assert(typeof mnemonic === "string");
+        if (!Array.isArray(accounts)) throw new Error("Property 'accounts' is not an array");
+        if (accounts.length !== 1) throw new Error("Property 'accounts' only supports one entry");
```
willclarktech, line 181: Is this check necessary?

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+        const account = accounts[0];
+        if (!isNonNullObject(account)) throw new Error("Account is not an object.");
+        const { algo, hdPath, prefix } = account as any;
```
willclarktech, line 184: How about extracting an `AccountData` type for use here and in `Secp256k1WalletData` etc?

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+        return Secp256k1Wallet.deserializeWithEncryptionKey(serialization, encryptionKey);
+      }
+      default:
+        throw new Error("Unsupported serialization type");
+    }
+  }
+
+  /** Base secret */
+  private readonly secret: EnglishMnemonic;
+  /** Derivation instrations */
```
willclarktech, line 221: "instrations"?

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+  public async serialize(password: string): Promise<string> {
+    const kdfConfiguration = basicPasswordHashingOptions;
+    const encryptionKey = await executeKdf(password, kdfConfiguration);
+    return this.serializeWithEncryptionKey(encryptionKey, kdfConfiguration);
+  }
+
+  /**
+   * Generates an encrypted serialization of this wallet.
+   *
+   * This is an advanced alternative of calling `serialize(password)` directly, which allows you to
```
willclarktech, line 297: Typo: "alternative to"

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+   * offload the KDF execution to an non-UI thread (e.g. in a WebWorker).
```
willclarktech, line 298: Typo: "a non-UI"

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+   *
+   * The caller is responsible for ensuring the key was derived with the given kdf options. If this
+   * is not the case, the wallet cannot be restored with the original password.
+   */
+  public async serializeWithEncryptionKey(
+    encryptionKey: Uint8Array,
+    kdfConfiguration: KdfConfiguration,
+  ): Promise<string> {
+    const encrytedData: Secp256k1WalletData = {
```
willclarktech, line 307: Typo: "encrypted"

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+    return this.serializeWithEncryptionKey(encryptionKey, kdfConfiguration);
+  }
+
+  /**
+   * Generates an encrypted serialization of this wallet.
+   *
+   * This is an advanced alternative of calling `serialize(password)` directly, which allows you to
+   * offload the KDF execution to an non-UI thread (e.g. in a WebWorker).
+   *
+   * The caller is responsible for ensuring the key was derived with the given kdf options. If this
```
willclarktech, line 300: "KDF"

packages/sdk38/src/secp256k1wallet.ts
willclarktech, line 307: `dataToEncrypt`?

packages/sdk38/src/secp256k1wallet.ts
```diff
@@ -0,0 +1,332 @@
...
+      mnemonic: this.mnemonic,
+      accounts: this.accounts.map((account) => ({
+        algo: account.algo,
+        hdPath: pathToString(account.hdPath),
+        prefix: account.prefix,
+      })),
+    };
+    const message = toUtf8(JSON.stringify(encrytedData));
+    const nonce = Random.getBytes(xchacha20NonceLength);
+    const encrypted = await Xchacha20poly1305Ietf.encrypt(message, encryptionKey, nonce);
```
willclarktech, line 317: `encryptedData`?

packages/sdk38/src/secp256k1wallet.ts
webmaster128: This is undefined by default. Since TypeScript checks that the variable is assigned in all possible branches before being used, the type can be `Uint8Array`

willclarktech: Ah right, I was thinking there was the possibility of serializing without encrypting, but great.

packages/sdk38/src/secp256k1wallet.ts
webmaster128: It turned out that the existing use cases (in-memory wallet and signing) don't become more complicated. We can even add a very simple `generate`. The interfaces only become non-trivial when dealing with serialization/deserialization.

packages/sdk38/src/testutils.spec.ts
webmaster128: Depends who you ask. The strict and original version requires padding. But there are other flavours and specs around.

packages/sdk38/src/secp256k1wallet.spec.ts
webmaster128: I agree but don't have any reference available.

packages/sdk38/src/secp256k1wallet.ts
webmaster128: Sure, thanks. This is a missing deduplication from the `wallet.ts`/`secp256k1wallet.ts` split

packages/sdk38/src/secp256k1wallet.ts
webmaster128: This is the equation from https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#generating-the-mnemonic

```
CS = ENT / 32
MS = (ENT + CS) / 11
```

rewritten to get `ENT` from `MS`. If we want better readability, I suggest switching over the possible word counts.

packages/sdk38/src/secp256k1wallet.ts
webmaster128: Thanks. This is leftover from when I split the functions. The switch only belongs in the version switcher `Secp256k1Wallet.deserialize`

packages/sdk38/src/secp256k1wallet.ts
webmaster128: It protects the user from losing data when loading and storing a multi-account seralization with this software

packages/sdk38/src/secp256k1wallet.ts
willclarktech: Ah, there's a bytes/bits conversion. 👌

packages/sdk38/src/secp256k1wallet.spec.ts
willclarktech: https://github.com/CosmWasm/cosmjs/blob/master/scripts/wasmd/README.md#preset-accounts

Unused testing account.

webmaster128: Yeah I know. But it is unclear how this came into existence. Maybe we generated it with the implementation we are testing here? I don't think so but have no reference.

packages/sdk38/src/secp256k1wallet.ts
webmaster128: Done in 2b0c4ec381f827d5bd352d7745d63479cf507b7e

packages/sdk38/src/secp256k1wallet.spec.ts
webmaster128: Hmm, what I want to test here is "deserialize", so this should be highlighted. Having to execute serialize is a necessary evil. Maintaining a serialization in code can be done, but is a lot of work because of the encrypted data and comes from serialize() anyways.

packages/sdk38/src/secp256k1wallet.ts
webmaster128: Done in b85af42919bff3fde2daa6b30f4af0d98e63f74d

packages/sdk38/types/wallet.d.ts
webmaster128: Since this is part of the JSON serialization, I left the types quite open. This enforces us to have runtime checks like

```ts
  switch (configuration.algorithm) {
    case "argon2id": {
      const options = configuration.params;
      if (!isArgon2idOptions(options)) throw new Error("Invalid format of argon2id params");
      return Argon2id.execute(password, cosmjsSalt, options);
    }
    default:
      throw new Error("Unsupported KDF algorithm");
  }
```

webmaster128:
@ethanfrey @willclarktech Added a bunch of commits, starting with _Reuse cosmjsSalt from wallet.ts_. Could you have another look?

CHANGELOG.md
```diff
@@ -11,3 +11,8 @@
 - @cosmjs/sdk38: Remove `Pen` type in favour of `OfflineSigner` and remove
   `Secp256k1Pen` class in favour of `Secp256k1Wallet` which takes an
   `OfflineSigner` instead of a `SigningCallback`.
+- @cosmjs/sdk38: `Secp256k1Wallet`s can not be generated randomly with
```
willclarktech, line 14: Typo: "can now"?

Issue #317, opened 2020-07-27
Rename "PostTx" to "BroadcastTx"


webmaster128:
In order to make the codebase more familiar for Cosmos developers, I think we should use the term "broadcast" instead of "post" or "submit". If I understand correctly, this is the primary term for dending a signed transaction to a node.

Thoughts @ethanfrey, @Thunnini, @fadeev?

ethanfrey:
Sounds good. "sign" and "broadcast" are the common terms i see in the cosmos-sdk and tendermint

fadeev:
Yes! "Broadcast" for sending signed transactions to `/txs` endpoint. "Submit" is a useful word for a higher level type of interaction: you submit a form that broadcasts a transaction.

PR #318, opened 2020-07-27
Add Signing{Cosmos,CosmWasm}Client.signAndPost

### LINKED ISSUE #300: sendTransaction based on sendToken
There could be a more generalized version of sendToken that sends a custom transaction. So, it would get account number, sequence and chain id, use an existing client, post request (base_req and body) to an API to get a formed tx, sign it locally and broadcast it to the chain.

Something like this:

```js
async sendTransaction({ state }, { type, body }) { // don't mind the state, it's a Vuex thing
  const { accountNumber, sequence } = await state.client.getNonce();
  const address = state.client.senderAddress;
  const chain_id = await state.client.getChainId();
  const req = {
    base_req: { chain_id, from: address },
    creator: address,
    ...body,
  };
  const { data } = await axios.post(`${API}/${chain_id}/${type}`, req);
  const { msg, fee, memo } = data.value;
  const signBytes = makeSignBytes(
    msg,
    fee,
    chain_id,
    memo,
    accountNumber,
    sequence
  );
  const signatures = [await state.wallet.sign(signBytes)];
  const signedTx = { msg: msg, fee, memo, signatures };
  return await state.client.postTx(signedTx);
}
```

Pretty useful method to have.

webmaster128:
Closes #300

ping @fadeev


fadeev:
Great, thanks for including this! It definitely saves a couple of steps and makes the lib friendlier.

@webmaster128 will you be renaming `signAndPost` to `signAndBroadcast` (as per https://github.com/CosmWasm/cosmjs/issues/317)?

webmaster128:
> will you be renaming `signAndPost` to `signAndBroadcast` (as per #317)?

Yeah, I want to do it as part of #317. Otherwise the method name and return types becomes super inconsistent.

PR #321, opened 2020-07-28
Organize WasmExtension tests by method

webmaster128:


packages/cosmwasm/src/lcdapi/wasm.spec.ts
```diff
@@ -99,20 +94,13 @@ async function instantiateContract(
...
+    amount: coins(5000000, "ucosm"),
     gas: "89000000",
   };
 
-  const { account_number, sequence } = (await client.auth.account(alice.address0)).result.value;
-  const signBytes = makeSignBytes([theMsg], fee, wasmd.chainId, memo, account_number, sequence);
-  const signature = await signer.sign(alice.address0, signBytes);
-  const signedTx = makeSignedTx(theMsg, fee, memo, signature);
-  return client.postTx(signedTx);
+  const firstAddress = (await signer.getAccounts())[0].address;
```
willclarktech, line 101: How about this?
```ts
const [{address}] = await signer.getAccounts();
```

webmaster128: I'd rather keep the explicit local variable name `firstAddress` because here this is an `OfflineSigner`, which can have multiple accounts.

PR #323, opened 2020-07-29
Update simapp to nightly-2020-07-29

webmaster128:
This uses the latest previewof the simapp docker image. See https://github.com/cosmos/cosmos-sdk/pull/6517


marbar3778:
pushing to `interchainio` right now, so you should be able to point there

webmaster128:
We now get CI fails in the coverage step because the chain creates files owned by root in `./scripts/simapp`. We don't have this problem with wasmd because all data is mounted into a volume:

```
  --mount type=bind,source="$SCRIPT_DIR/template",target=/template \
  --mount type=volume,source=wasmd_data,target=/root \
```

instead of

```
  --mount type=bind,source="$SCRIPT_DIR/template",target=/root \
```

Let's see if we can do the same here.

marbar3778:
would you want an image that is tagged as `latest`, that way you don't have to keep updating? although then you could get daily breakage.

webmaster128:
> would you want an image that is tagged as `latest`, that way you don't have to keep updating?

No thank you. We can update the backend once in a while or even daily, no problem. For me reproducibility is more important than simplicity. You also want the same behaviour locally and in the CI. Having to `docker pull` is annoyingly slow and `docker run ...:latest` will not update the image if you have any version of the latest tag.

PR #325, opened 2020-07-30
Upgrade wasmd to v0.10.0-beta2

webmaster128:


packages/launchpad/src/lcdapi/slashing.spec.ts
```diff
@@ -17,7 +17,7 @@ describe("SlashingExtension", () => {
         height: jasmine.stringMatching(nonNegativeIntegerMatcher),
         result: [
           {
-            address: "cosmosvalcons1cw2p365mwycxuy89ml3crg96tqen7vvm0xgujv",
+            address: "cosmosvalcons14dsaw6zrrk3zugfgwwhmwap8p5cxsemfesgcaw",
```
willclarktech, line 20: Maybe this should be in test utils too?

webmaster128: I was thinking about this as well but then I did not know where the address comes from. Since I don't understand the sematics, I have no name for it. Thus I left it for now. Ideas?

willclarktech: 🤐

PR #332, opened 2020-08-03
Test CLI example scripts

webmaster128:
This ensures the examples still compile and run

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -140,6 +140,18 @@ jobs:
             SKIP_BUILD: 1
           command: yarn selftest
       - run:
+          name: Run CLI examples
+          working_directory: packages/cli
+          environment:
+            SKIP_BUILD: 1
+          command: |
+            ./bin/cosmwasm-cli --init examples/delegate.ts --code "process.exit(0)"
+            ./bin/cosmwasm-cli --init examples/faucet_addresses.ts --code "process.exit(0)"
+            ./bin/cosmwasm-cli --init examples/generate_address.ts --code "process.exit(0)"
+            ./bin/cosmwasm-cli --init examples/helpers.ts --code "process.exit(0)"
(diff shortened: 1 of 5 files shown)


webmaster128:
In this PR I added `--init` with URLs. I tested `./bin/cosmwasm-cli --debug --init https://raw.githubusercontent.com/CosmWasm/cosmjs/v0.22.0-alpha.1/packages/cli/examples/generate_address.ts`, which works.

```sh
npx @cosmjs/cli --init https://raw.githubusercontent.com/CosmWasm/cosmjs/v0.22.0-alpha.1/packages/cli/examples/generate_address.ts
```

should work as well.

.circleci/config.yml
```diff
@@ -139,6 +139,18 @@ jobs:
           environment:
             SKIP_BUILD: 1
           command: yarn selftest
+      - run:
+          name: Run CLI examples
+          working_directory: packages/cli
+          environment:
+            SKIP_BUILD: 1
+          command: |
+            ./bin/cosmwasm-cli --init examples/delegate.ts --code "process.exit(0)"
```
ethanfrey, line 148: good to test this.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## HEAD
 
+- @cosmjs/cli: Now supports HTTPs URLs for `--init` code sources.
```
ethanfrey, line 5: This is a great feature add.

Do you want to test that one time in the CI? Or would it be too hard to find some stable code to test against? 

I mean, I guess generate_address should be relatively stable and we could point to master until some PR needs to point to it's own branch?

webmaster128: Yeah, testing against an external resource is a problem. I'd leave it as is and use a local HTTP server is we really have problem with this.

PR #340, opened 2020-08-04
Allow Bech32.decode to decode long addresses

### LINKED ISSUE #335: Ensure Bech32.encode throws for output length > 90
The [bech32 spec says an address is no longer than 90 characters](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#Bech32). To avoid any kind of decoding trouble later on, we should ensure this library throws when the result address exceeds 90 characters.

For decoding on the other hand there is no reason to enfore this restriction. Here we could even test if the current implementation supports very long addresses.

Some sample data here: https://github.com/cosmos/cosmos-sdk/pull/6237

willclarktech:
Closes #335

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 - @cosmjs/cosmwasm: Rename `SigningCosmWasmClient.signAndPost` method to
   `.signAndBroadcast`.
 - @cosmjs/demo-staking: Remove package and supporting scripts.
+- @cosmjs/encoding: Add `limit` parameter to `Bech32.encode` and `.decode`. The
+  new default limit for decoding is infinity (was 90 before). Set it to 90 to
+  create a strict decoder.
 - @cosmjs/launchpad: Rename `CosmosClient.postTx` method to `.broadcastTx`.
 - @cosmjs/launchpad: Rename `SigningCosmosClient.signAndPost` method to
   `.signAndBroadcast`.
--- a/packages/encoding/src/bech32.spec.ts
+++ b/packages/encoding/src/bech32.spec.ts
@@ -6,14 +6,69 @@ describe("Bech32", () => {
   // bech32 -e -h eth 9d4e856e572e442f0a4b2763e72d08a0e99d8ded
   const ethAddressRaw = fromHex("9d4e856e572e442f0a4b2763e72d08a0e99d8ded");
 
-  it("encodes", () => {
-    expect(Bech32.encode("eth", ethAddressRaw)).toEqual("eth1n48g2mjh9ezz7zjtya37wtgg5r5emr0drkwlgw");
+  describe("encode", () => {
+    it("works", () => {
+      expect(Bech32.encode("eth", ethAddressRaw)).toEqual("eth1n48g2mjh9ezz7zjtya37wtgg5r5emr0drkwlgw");
+    });
+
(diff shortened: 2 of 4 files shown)


packages/encoding/src/bech32.spec.ts
```diff
@@ -6,14 +6,47 @@ describe("Bech32", () => {
...
+
+    // See https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#Bech32
+    it("works if result is 90 characters", () => {
+      const result = Bech32.encode("eth", fromHex("a".repeat(100)));
+      expect(result.length).toEqual(90);
+    });
+
+    it("throws if result exceeds 90 characters", () => {
+      expect(() => Bech32.encode("eth", fromHex("a".repeat(102)))).toThrowError(/exceeds length limit/i);
+    });
```
webmaster128, line 22: Can you test short data and very long prefix as well?

packages/encoding/src/bech32.spec.ts
```diff
@@ -6,14 +6,47 @@ describe("Bech32", () => {
...
+
+    // See https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#Bech32
+    it("works if result is 90 characters", () => {
+      const result = Bech32.encode("eth", fromHex("a".repeat(100)));
+      expect(result.length).toEqual(90);
+    });
+
+    it("throws if result exceeds 90 characters", () => {
+      expect(() => Bech32.encode("eth", fromHex("a".repeat(102)))).toThrowError(/exceeds length limit/i);
```
webmaster128, line 21: `new Uint8Array(51)` is probably an easier way to get 51 bytes of data

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 - @cosmjs/cosmwasm: Rename `CosmWasmClient.postTx` method to `.broadcastTx`.
 - @cosmjs/cosmwasm: Rename `SigningCosmWasmClient.signAndPost` method to
   `.signAndBroadcast`.
+- @cosmjs/encoding: Add `limit` parameter to `Bech32.decode`. If unspecified,
+  this will now allow decoding of addresses which exceed the specification limit
+  of 90 characters.
```
webmaster128, line 10: The describtion is hard to understand I thing. Suggestion: "The new default limit is infinity (was 90 before). Set it to 90 to create a strict decoder."

packages/encoding/src/bech32.spec.ts
```diff
@@ -6,14 +6,69 @@ describe("Bech32", () => {
...
+    });
+
+    // See https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#Bech32
+    it("works if result is 90 characters", () => {
+      const result = Bech32.encode("eth", new Uint8Array(50));
+      expect(result.length).toEqual(90);
+    });
+
+    it("throws if result exceeds 90 characters", () => {
+      expect(() => Bech32.encode("eth", new Uint8Array(51))).toThrowError(/exceeds length limit/i);
```
webmaster128, line 29: Okay, a lot more tests, great 👍 But the one I was thinking about is still missing. It can be added in this block:

```ts
expect(() => Bech32.encode("p".repeat(70), new Uint8Array(20))).toThrowError(/exceeds length limit/i);
```

PR #343, opened 2020-08-05
Break && lists into separate commands

webmaster128:
A shell script with set -o errexit is not stopped when there is an error in the first command of commandA && commandB because

> The shell does not exit if the command that fails is ... part of any command executed in a && or || list

https://unix.stackexchange.com/a/490177


webmaster128:
`automerge` only works with `master` as the base branch: https://github.com/CosmWasm/cosmjs/blob/master/.mergify.yml#L7

PR #345, opened 2020-08-05
Start simapp's in-process API

webmaster128:
With this change, we can use the HTTP API for debugging some stuff, e.g. `GET http://localhost:1318/auth/accounts/cosmos1pkptre7fdkl6gfrzlesjjvhxhlc3r4gmmk8rs6`:

```json
{
  "height": "0",
  "result": {
    "type": "cosmos-sdk/BaseAccount",
    "value": {
      "address": "cosmos1pkptre7fdkl6gfrzlesjjvhxhlc3r4gmmk8rs6",
      "account_number": "1"
    }
  }
}
```

The missing sequence indicates it is 0.

### DIFF
--- a/scripts/simapp/start.sh
+++ b/scripts/simapp/start.sh
@@ -4,6 +4,8 @@ command -v shellcheck > /dev/null && shellcheck "$0"
 
 TENDERMINT_PORT_GUEST="26657"
 TENDERMINT_PORT_HOST="26657"
+API_PORT_GUEST="1317"
+API_PORT_HOST="1318"
 
 SCRIPT_DIR="$(realpath "$(dirname "$0")")"
 # shellcheck source=./env
@@ -21,6 +23,7 @@ docker volume rm -f simapp_data
 docker run --rm \
   --name "$CONTAINER_NAME" \
   -p "$TENDERMINT_PORT_HOST":"$TENDERMINT_PORT_GUEST" \
+  -p "$API_PORT_HOST":"$API_PORT_GUEST" \
   --mount type=bind,source="$SCRIPT_DIR/template",target=/template \
   --mount type=volume,source=simapp_data,target=/root \
   "$REPOSITORY:$VERSION" \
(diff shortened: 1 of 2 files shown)


scripts/simapp/template/.simapp/config/app.toml
```diff
@@ -99,4 +99,4 @@ rpc-write-timeout = 0
 rpc-max-body-bytes = 1000000
 
 # EnableUnsafeCORS defines if CORS should be enabled (unsafe - use it at your own risk)
-enabled-unsafe-cors = false
+enabled-unsafe-cors = true
```
ethanfrey, line 102: 😄

scripts/simapp/start.sh
```diff
@@ -4,6 +4,8 @@ command -v shellcheck > /dev/null && shellcheck "$0"
 
 TENDERMINT_PORT_GUEST="26657"
 TENDERMINT_PORT_HOST="26657"
+API_PORT_GUEST="1317"
+API_PORT_HOST="1318"
```
ethanfrey, line 8: Why not 1317 on the host? It would be what everyone expects

ethanfrey:
Nice one. Great for an smooth transition.

I got excited about the raw queries today, and want to see if I can do the whole proof stack. But it will take a while to get them all

scripts/simapp/start.sh
webmaster128: wasmd runs on 1317. Since the simd API is only used for debugging, I thought it is better to move this one port up.

ethanfrey: Fair enough, but then we should probably bump up the tendermint rpc port as well?

webmaster128: Possible, but not needed as long as we don't connect to wasmd Tendermint, which we don't:

```sh
LCD_API_PORT_GUEST="1317"
LCD_API_PORT_HOST="1317"
# Tendermint port (26657) and p2p port (26656) are not exposed since we don't need them for testing
```

I'll add a port overview for this repo …

PR #351, opened 2020-08-06
SIGN_MODE_DIRECT

### LINKED ISSUE #212: Add support for SIGN_MODE_DIRECT
After #211 we need to create a `SignDoc` and `TxRaw` in order to submit a binary tx to Tendermint.

This can either be tested by using curl to send the tranyaction to Tendermint (https://docs.tendermint.com/master/rpc/) or by doing #204 first

willclarktech:
Opening here to avoid conflicts on 212-sign-mode-direct.

Closes #212 

Signatures are different despite the transaction objects otherwise being identical. I'll investigate this afternoon.

### DIFF
--- a/packages/proto-signing/package.json
+++ b/packages/proto-signing/package.json
@@ -48,6 +48,7 @@
   },
   "devDependencies": {
     "@cosmjs/encoding": "^0.22.1",
+    "@cosmjs/launchpad": "^0.22.1",
     "@cosmjs/utils": "^0.22.1"
   }
 }
--- a/packages/proto-signing/src/signing.spec.ts
+++ b/packages/proto-signing/src/signing.spec.ts
@@ -2,6 +2,7 @@
 import { Bech32, fromBase64, fromHex, toHex } from "@cosmjs/encoding";
 import { Secp256k1Wallet } from "@cosmjs/launchpad";
 
+import { omitDefault } from "./adr27";
 import { cosmos } from "./generated/codecimpl";
 import { defaultRegistry } from "./msgs";
 import { Registry, TxBodyValue } from "./registry";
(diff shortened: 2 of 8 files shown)


packages/proto-signing/package.json
```diff
@@ -44,6 +44,8 @@
     "postdefine-proto": "prettier --write \"src/generated/codecimpl.*\""
   },
   "dependencies": {
+    "@cosmjs/launchpad": "^0.22.0",
+    "@cosmjs/tendermint-rpc": "^0.22.0",
```
webmaster128, line 48: - Do we really need launchpad?
- Shouldn't those be dev dependencies?

packages/proto-signing/src/signing.spec.ts
```diff
@@ -0,0 +1,359 @@
...
+              },
+            ],
+          },
+        },
+      ],
+    };
+    const txBodyBytes = myRegistry.encode({
+      typeUrl: "/cosmos.tx.TxBody",
+      value: txBodyFields,
+    });
```
webmaster128, line 132: Does this do the two step serialization of messages? A message is added to TxBody as raw bytes

packages/proto-signing/package.json
willclarktech: We use launchpad for the wallet. It seems pretty convenient.

packages/proto-signing/src/signing.spec.ts
willclarktech: Yes: https://github.com/CosmWasm/cosmjs/blob/master/packages/proto-signing/src/registry.ts#L65-L90

packages/proto-signing/src/msgs.ts
```diff
@@ -26,3 +26,19 @@ export class MsgSend extends Message {
   @cosmosField.repeatedMessage(3, Coin)
   public readonly amount?: readonly Coin[];
 }
+
+// TODO: move somewhere else as needed
+@registered(defaultRegistry, "/cosmos.auth.BaseAccount")
+export class BaseAccount extends Message {
```
webmaster128, line 32: This type is now in the autogenerated code in @cosmjs/starport (where all the queries are) and not needed here anymore

Issue #350, opened 2020-08-06
Add tendermint rpc queries


ethanfrey:
Query the following modules using tendermint rpc. All queries that can be done directly via `key` should be done so (as they can get proofs). We can also add higher-level app-queries when it needs processing by the node. I will use the term `proveXyz` when it is a key query (that can be proven to a header) and `getXyz` when it is an app query (that can never be proven):

- [x] auth
  - [x] `proveAccount` (#361)
- [x] bank
  - [x] `proveBalance` (one denom)
  - [x] `getAllBalances` (all denoms)
- [ ] ibc
  - [x] unverified
  - [ ] verified

webmaster128:
@ethanfrey are those the IBC queries we need? https://github.com/cosmos/cosmos-sdk/blob/master/proto/ibc/channel/query.proto#L10-L39

ethanfrey:
Yes, those are the `unverified` queries we need. Good to start with them (super easy). Can you list them above?

I will hand-code the proven version, but it would be great to have those as reference. So the proven version should give the same data as the grpc version where the perform the same query.

webmaster128:
okay. Do you know why they are not added in an ibc folder as the other modules in https://github.com/cosmos/cosmos-sdk/tree/master/proto/cosmos?

ethanfrey:
> okay. Do you know why they are not added in an ibc folder as the other modules in https://github.com/cosmos/cosmos-sdk/tree/master/proto/cosmos?

They are subfolders as IBC is huge.

Actually we will probably want something from client, connection, and commitment at some point, maybe more.

webmaster128:
Last PR was here: https://github.com/CosmWasm/cosmjs/pull/410. Done

PR #352, opened 2020-08-06
Tendermint rpc queries

ethanfrey:
Start work on #350 

This begins adding some query functions to `@cosmjs/sdk40`.

If this works, we can merge this and I will make a new PR adding app-level queries (get), and then I can start working on the prove stuff.

### DIFF
--- a/packages/sdk40/src/query.ts
+++ b/packages/sdk40/src/query.ts
@@ -4,7 +4,10 @@ import { Client } from "@cosmjs/tendermint-rpc";
 
 import { Coin } from "./structs";
 
-async function proveByKey(client: Client, store: string, key: Uint8Array): Promise<Uint8Array> {
+// getByKey will use a raw (storename, key) pair to query state
+// and verify the merkle proof against the header
+// (when the client does light-client header verification, this will be fully proven)
+async function getByKey(client: Client, store: string, key: Uint8Array): Promise<Uint8Array> {
   const resp = await client.abciQuery({
     // we need the StoreKey for the module, not the module name
     // https://github.com/cosmos/cosmos-sdk/blob/8cab43c8120fec5200c3459cbf4a92017bb6f287/x/auth/types/keys.go#L12
@@ -26,19 +29,22 @@ async function proveByKey(client: Client, store: string, key: Uint8Array): Promi
   return resp.value;
 }
 
-export async function proveAccount(client: Client, address: string): Promise<google.protobuf.Any> {
+export async function getAccount(client: Client, address: string): Promise<google.protobuf.Any> {
   const binAddress = Bech32.decode(address).data;
 
   // https://github.com/cosmos/cosmos-sdk/blob/8cab43c8120fec5200c3459cbf4a92017bb6f287/x/auth/types/keys.go#L29-L32
   const accountKey = Uint8Array.from([1, ...binAddress]);
 
-  const response = await proveByKey(client, "acc", accountKey);
+  // getByKey will use a raw (storename, key) pair to query state
+  // and verify the merkle proof against the header
+  // (when the client does light-client header verification, this will be fully proven)
+  const response = await getByKey(client, "acc", accountKey);
 
   // TODO: is there a better way to do the second-level unwrapping (decode the Any account type)?
   return google.protobuf.Any.decode(response);
 }
 
-export async function proveBalance(client: Client, address: string, denom: string): Promise<Coin> {
+export async function getBalance(client: Client, address: string, denom: string): Promise<Coin> {
   const binAddress = Bech32.decode(address).data;
 
   // balance key is a bit tricker, using some prefix stores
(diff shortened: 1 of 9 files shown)


webmaster128:
This clearly shows the idea of what needs to be shared code and what needs to be query specific. Very helpful, thanks. Will integrate this into #353 now.

packages/sdk40/src/structs.ts
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { cosmosField, cosmosMessage, Registry } from "@cosmjs/proto-signing";
+import { Message } from "protobufjs";
+
+export const defaultRegistry = new Registry();
```
webmaster128, line 5: I think we don't need the registry at all for those use cases because we know what type is expected. In #352 I use `BaseAccount` without a registry.

packages/sdk40/types/query.d.ts
```diff
@@ -0,0 +1,5 @@
+import { google } from "@cosmjs/proto-signing";
+import { Client } from "@cosmjs/tendermint-rpc";
+import { Coin } from "./structs";
+export declare function proveAccount(client: Client, address: string): Promise<google.protobuf.Any>;
+export declare function proveBalance(client: Client, address: string, denom: string): Promise<Coin>;
```
webmaster128, line 5: "To prove" is the wrong term here, right? If I understand correctly, the chain proves that its answer is correct by providing a proof and the client "verifies" this proof.

packages/sdk40/src/query.ts
```diff
@@ -0,0 +1,54 @@
+import { Bech32, toAscii, toHex } from "@cosmjs/encoding";
+import { google } from "@cosmjs/proto-signing";
+import { Client } from "@cosmjs/tendermint-rpc";
+
+import { Coin } from "./structs";
+
+async function proveByKey(client: Client, store: string, key: Uint8Array): Promise<Uint8Array> {
```
webmaster128, line 7: This signature is very helpful

packages/sdk40/src/structs.ts
ethanfrey: This may be ModuleAccount, VestingAccount or ContinuousVestingAccount.

Forcing it to BaseAccount may fail on some chains

packages/sdk40/types/query.d.ts
ethanfrey: Good point.

What terms should we use. I want two verbs.

`aaaaAccount`  would signify some way to get the account data that has been validated via merkle and light client proofs (later we just swap in a header-verifying tendermint rpc implementations and we have the full validation here)

`bbbbAccount` would get the account data, just trusting the node it was talking to.

I thought of `proveAccount` and `getAccount`, but you have a valid point that isn't correct terms.

`getAccount` / `getProvenAccount`?

`getAccount` / `verifyAccount` ?

Other ideas?

packages/sdk40/src/structs.ts
webmaster128: I see. Happy to generalize this. But with the current registry aproach we can only encode in a typesafe way, not decode. We can extend this switch to more account types:

```ts
const { typeUrl, value } = decodeAny(responseData);
    switch (typeUrl) {
      case "/cosmos.auth.BaseAccount": {
        const { account_number, sequence } = BaseAccount.decode(value);
        assert(account_number !== undefined);
        assert(sequence !== undefined);
        return {
          accountNumber: uint64FromProto(account_number).toNumber(),
          sequence: uint64FromProto(sequence).toNumber(),
        };
      }
      default:
        throw new Error(`Unsupported type: ${typeUrl}`);
    }
```

ethanfrey: I should add some other types of accounts to genesis...

ethanfrey:
```
Usage:
  simd add-genesis-account [address_or_key_name] [coin][,[coin]] [flags]

Flags:
      --vesting-amount string    amount of coins for vesting accounts
      --vesting-end-time int     schedule end time (unix epoch) for vesting accounts
      --vesting-start-time int   schedule start time (unix epoch) for vesting accounts
```

packages/sdk40/types/query.d.ts
webmaster128: Let's aim high and take verification for granted. Then unverified queries get a warning in the name. E.g.

```
getSequence
getBalance
getUnverifiedBalances
```

ethanfrey: I will add a non verified command here to demo it (and see if I can translate from the new grpcQuery types).

ethanfrey: Okay, names changed and grpc query added.

To do this in scale, we are going to need to auto-gen the decorator structs from the cosmos proto files.

ethanfrey:
Test failures due to startup issues with `scripts/simapp/start.sh`

willclarktech:
@ethanfrey OK to delete the branch?

ethanfrey:
Sure, this was deprecated by Simon's work. Let's delete

PR #355, opened 2020-08-06
Add getUnverifiedAllBalances

webmaster128:
This includes the remaining work from #352 (Thank you!) into the new StargateClient created in #353.

Closes #352

@willclarktech do you have a better but similar alternative to "getUnverifiedAllBalances" in mind? The idea is to highlight those queries that cannot be verified as unsafe.


webmaster128:
We should think about removing the hand-written request and response types and use auto-generated code instead. The decorator version was meant to be used for message types in order to allow easy creation of custom types.

ethanfrey:
> We should think about removing the hand-written request and response types and use auto-generated code instead. The decorator version was meant to be used for message types in order to allow easy creation of custom types.

Sounds good. Once this works, we can look at ways to better import... less manual fiddling.

I was going to start documenting where all this nonsense comes from but the branches were in flux.

packages/stargate/src/stargateclient.ts
```diff
@@ -80,6 +78,29 @@ export class StargateClient {
...
+   * Uses the grpc queries (which iterates over the store internally), and we cannot get
+   * proofs from such a method.
+   */
+  public async getUnverifiedAllBalances(address: string): Promise<readonly Coin[]> {
+    const path = "/cosmos.bank.Query/AllBalances";
+    const request = QueryAllBalancesRequest.encode({ address: Bech32.decode(address).data }).finish();
+    const responseData = await this.queryUnverified(path, request);
+    const response = QueryAllBalancesResponse.decode(responseData);
+    return (response.balances || []).map(
+      (balance): Coin => {
```
ethanfrey, line 93: This map is quite long and obscured the actual functionality.
Maybe an `assertCoin(any): Coin` function could be added and we just map with that.

packages/stargate/src/stargateclient.ts
```diff
@@ -80,6 +78,29 @@ export class StargateClient {
     }
   }
 
+  /**
+   * Queries all balances for all denoms that belong to this address.
+   *
+   * Uses the grpc queries (which iterates over the store internally), and we cannot get
+   * proofs from such a method.
+   */
+  public async getUnverifiedAllBalances(address: string): Promise<readonly Coin[]> {
```
ethanfrey, line 87: Otherwise, yes this captures the essence of what I was doing

packages/stargate/src/stargateclient.ts
webmaster128: 👍 sure, will pull out (and document and test) the proto Coin to launchpad coin mapping once we need it in a few places.

packages/stargate/package.json
```diff
@@ -39,6 +39,7 @@
   },
   "dependencies": {
     "@cosmjs/encoding": "^0.22.0",
+    "@cosmjs/launchpad": "^0.22.0",
```
willclarktech, line 42: ???

packages/stargate/src/stargateclient.ts
willclarktech, line 87: Suggest `getAllBalancesUnverified`, otherwise it sounds really weird. (And using `Unverified` generally as a suffix.)

packages/stargate/src/stargateclient.ts
willclarktech: Also suggest a `.forEach` since if an array is returned all the elements are the same as the original array.

webmaster128: The idea is to return new element of a different `Coin` type. Input is the protobuf type, outut is the `Coin` interface from @cosmjs/launchpad.

packages/stargate/package.json
webmaster128: We use `import { Coin } from "@cosmjs/launchpad";`

willclarktech: Why don't we have `Coin` in Stargate?

webmaster128: We can also have a copy of the type, no problem. In the meantime we use a few more amino helpers from `@cosmjs/launchpad` in `@cosmjs/stargate` which are temporary (hopefully). But I agree, it would be nice to remove this dependency at some point, either by copying what we need or by extrating commin symbols in a shared package.

PR #358, opened 2020-08-09
Add omitDefault implementation according to ADR 027

webmaster128:


webmaster128:
Thank you for the cleanups, @willclarktech

PR #361, opened 2020-08-10
Add StargateClient.getAccount

webmaster128:
Based on #360

For #350

### DIFF
--- a/packages/stargate/src/stargateclient.ts
+++ b/packages/stargate/src/stargateclient.ts
@@ -26,6 +26,18 @@ function uint64FromProto(input: number | Long): Uint64 {
   return Uint64.fromString(input.toString());
 }
 
+function decodeBaseAccount(data: Uint8Array, prefix: string): Account {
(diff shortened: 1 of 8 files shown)


packages/stargate/src/stargateclient.ts
```diff
@@ -26,6 +26,18 @@ function uint64FromProto(input: number | Long): Uint64 {
   return Uint64.fromString(input.toString());
 }
 
+function decodeBaseAccount(data: Uint8Array, prefix: string): Account {
+  const { address, pubKey, accountNumber, sequence } = cosmos.auth.BaseAccount.decode(data);
+  // Pubkey is still Amino-encoded in BaseAccount (https://github.com/cosmos/cosmos-sdk/issues/6886)
```
ethanfrey, line 31: Can you make a tracking issue for this on CosmJS - so we can fix this when the sdk updates (or at least not be surprised why it broke)

webmaster128: Jupp: https://github.com/CosmWasm/cosmjs/issues/363

Issue #272, opened 2020-07-02
Create CosmosClient/SigningCosmosClient in @cosmjs/stargate


webmaster128:
We need higher level clients `CosmosClient`/`SigningCosmosClient` in @cosmjs/stargate that have the same interface as in @cosmjs/launchpad to make it easy to upgrade. They should use tendermint-rpc and https://github.com/CosmWasm/cosmjs/issues/212 under the hood.

**read-only**:

- [x] Add `StargateClient`
- [x] Add `getChainId` (directly via Tendermint RPC): #362
- [x] Add `getHeight` (directly via Tendermint RPC): #362
- [x] Add `getSequence`
- [x] Add `getAccount` (address, pubKey, accountNumber, sequence but no balances): #361
- [x] Add `getBalance`/`getAllBalancesUnverified`
- [x] Add `getBlock` #375
- [x] Add `searchTx` #380 
- [x] Add `broadcastTx` (directly via Tendermint RPC) #373 
- [x] Add `txsQuery` (directly via Tendermint RPC) #380 

**read/write**:

Depends on: #212

- [ ] Add `SigningStargateClient`
- [ ] Add `sendTokens`
- [ ] Add `signAndBroadcast`

webmaster128:
I started using `StargateClient` (and `SigningStargateClient`) as alternatives to CosmosClient/SigningCosmosClient in order to avoid unnecessary naming collisions in the project. So the description above is a bit outdated.

PR #362, opened 2020-08-10
Add StargateClient.getChainId and .getHeight

webmaster128:
~Based on #359~

For #272

### DIFF
--- a/packages/stargate/src/stargateclient.ts
+++ b/packages/stargate/src/stargateclient.ts
@@ -78,6 +78,11 @@ export class StargateClient {
     return this.chainId;
   }
 
+  public async getHeight(): Promise<number> {
+    const status = await this.tmClient.status();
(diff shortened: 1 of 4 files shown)


packages/stargate/src/stargateclient.ts
```diff
@@ -49,8 +49,14 @@ function coinFromProto(input: cosmos.ICoin): Coin {
   };
 }
 
+/** Use for testing only */
+export interface PrivateStargateClient {
```
willclarktech, line 53: If this is only for testing, why not put it in a test file?

webmaster128: Last time we did this (`export interface PrivateCosmWasmClient {`), it was needed in multiple test files. Also I think it is nice to have it close to the definitions of the fields. In this case we could move it, but if you don't have a strong opinion here, I'd leave it.

PR #368, opened 2020-08-11
Make faucet configurable for networks with different prefix/tokens

### LINKED ISSUE #364: Allow to set bech32 prefix, tokens for faucet on command line
Either via cli arguments or via environment variables with defaults.

See https://github.com/CosmWasm/cosmjs/blob/master/packages/faucet/src/constants.ts#L8-L24

For the testnet, I want to set the token to `SHELL`/`ushell` and the bech32 prefix to `coral` but this is not possible with the current binary.

webmaster128:
Closes #364

### DIFF
--- a/packages/faucet/src/actions/generate.ts
+++ b/packages/faucet/src/actions/generate.ts
@@ -4,17 +4,15 @@ import * as constants from "../constants";
 import { createWallets } from "../profile";
 
 export async function generate(args: readonly string[]): Promise<void> {
-  if (args.length < 1) {
-    throw Error(
-      `Not enough arguments for action 'generate'. See '${constants.binaryName} help' or README for arguments.`,
+  if (args.length > 0) {
+    console.warn(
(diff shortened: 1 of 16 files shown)


packages/faucet/src/tokens.spec.ts
```diff
@@ -0,0 +1,78 @@
+import { parseBankToken, parseBankTokens } from "./tokens";
+
+describe("tokens", () => {
```
ethanfrey, line 3: Good tests

packages/faucet/src/actions/generate.ts
```diff
@@ -4,17 +4,15 @@ import * as constants from "../constants";
...
   }
 
-  const chainId = args[0];
-
   const mnemonic = Bip39.encode(Random.getBytes(16)).toString();
   console.info(`FAUCET_MNEMONIC="${mnemonic}"`);
 
   // Log the addresses
-  await createWallets(mnemonic, chainId, constants.concurrency, true);
+  await createWallets(mnemonic, constants.addressPrefix, constants.concurrency, true);
```
ethanfrey, line 17: Okay, nice, so we just use the same env variable here :+1:

PR #372, opened 2020-08-12
Sync proto code generation with simd blockchain

webmaster128:
This ensures code matches the test blockchain. This process will hopefully get easier with https://github.com/cosmos/cosmos-sdk/issues/7013


webmaster128:
> but would it make sense for stargate to have a dependency on proto-signing, and export the codec from there?

The code generation we need for queries is going to be massive. I'd rather try to trim down the signing library as much as possible (maybe even converting all autogeneration to the decorator system once the .protos are stable).

PR #373, opened 2020-08-12
Add broadcastTx method to Stargate client

willclarktech:
Part of #272

### DIFF
--- a/packages/stargate/src/stargateclient.spec.ts
+++ b/packages/stargate/src/stargateclient.spec.ts
@@ -1,7 +1,23 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { Bech32, fromBase64 } from "@cosmjs/encoding";
+import { Secp256k1Wallet } from "@cosmjs/launchpad";
+import { makeSignBytes, omitDefaults, Registry } from "@cosmjs/proto-signing";
 import { assert, sleep } from "@cosmjs/utils";
 
-import { PrivateStargateClient, StargateClient } from "./stargateclient";
-import { nonExistentAddress, pendingWithoutSimapp, simapp, unused, validator } from "./testutils.spec";
+import { cosmos } from "./generated/codecimpl";
+import { assertIsBroadcastTxSuccess, PrivateStargateClient, StargateClient } from "./stargateclient";
+import {
+  faucet,
+  makeRandomAddressBytes,
+  nonExistentAddress,
+  pendingWithoutSimapp,
+  simapp,
+  unused,
+  validator,
+} from "./testutils.spec";
+
+const { AuthInfo, SignDoc, Tx, TxBody } = cosmos.tx;
+const { PublicKey } = cosmos.crypto;
 
 describe("StargateClient", () => {
   describe("connect", () => {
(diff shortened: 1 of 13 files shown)


packages/stargate/src/stargateclient.ts
```diff
@@ -22,6 +23,45 @@ export interface SequenceResponse {
...
+  readonly code: number;
+  readonly transactionHash: string;
+  readonly rawLog?: string;
+  readonly data?: Uint8Array;
+}
+
+export interface BroadcastTxSuccess {
+  readonly height?: number;
+  readonly transactionHash: string;
+  readonly logs: JsonCompatibleValue;
```
webmaster128, line 37: I tried this type for logs before, which ended up with big trouble. Since we create this field, I see two approaches:
1. Remove and just provide `rawLog` (for now). We can parse in test code.
2. Create a log parser that returns a checked and typed result

I'd start with 1. and create a ticket for 2.

packages/stargate/src/stargateclient.ts
```diff
@@ -22,6 +23,45 @@ export interface SequenceResponse {
...
+export interface BroadcastTxFailure {
+  readonly height?: number;
+  readonly code: number;
+  readonly transactionHash: string;
+  readonly rawLog?: string;
+  readonly data?: Uint8Array;
+}
+
+export interface BroadcastTxSuccess {
+  readonly height?: number;
```
webmaster128, line 35: Can height be unset in the success case?

packages/stargate/src/testutils.spec.ts
```diff
@@ -1,9 +1,16 @@
...
+
 export function pendingWithoutSimapp(): void {
   if (!process.env.SIMAPP_ENABLED) {
     return pending("Set SIMAPP_ENABLED to enable Simapp based tests");
   }
 }
 
+export function makeRandomAddress(): string {
+  return Bech32.encode("cosmos", Random.getBytes(20));
+}
```
webmaster128, line 12: This is unused now. I think it would be nice to have something similar creating a random 20 byte array. The function would not do much but having the name is better than `Random.getBytes(20)`

packages/stargate/src/testutils.spec.ts
```diff
@@ -12,6 +19,16 @@ export const simapp = {
...
 };
 
+export const faucet = {
+  mnemonic:
+    "economy stock theory fatal elder harbor betray wasp final emotion task crumble siren bottom lizard educate guess current outdoor pair theory focus wife stone",
+  pubkey: {
+    type: "tendermint/PubKeySecp256k1",
+    value: "A08EGB7ro1ORuFhjOnZcSgwYlpe0DSFjVNUIkNNQxwKQ",
+  },
+  address: "cosmos1pkptre7fdkl6gfrzlesjjvhxhlc3r4gmmk8rs6",
```
webmaster128, line 29: Can we rename the fields to `pubkey0` and `address0` to prepare for multi account usage? We create all of those on simd as well:

```
1. **Faucet**<br>
   economy stock theory fatal elder harbor betray wasp final emotion task crumble siren bottom lizard educate guess current outdoor pair theory focus wife stone<br>
   Address 0: cosmos1pkptre7fdkl6gfrzlesjjvhxhlc3r4gmmk8rs6<br>
   Address 1: cosmos10dyr9899g6t0pelew4nvf4j5c3jcgv0r73qga5<br>
   Address 2: cosmos1xy4yqngt0nlkdcenxymg8tenrghmek4nmqm28k<br>
   Address 3: cosmos142u9fgcjdlycfcez3lw8x6x5h7rfjlnfhpw2lx<br>
   Address 4: cosmos1hsm76p4ahyhl5yh3ve9ur49r5kemhp2r0dcjvx<br>
   Pubkey 0: A08EGB7ro1ORuFhjOnZcSgwYlpe0DSFjVNUIkNNQxwKQ<br>
   Pubkey 1: AiDosfIbBi54XJ1QjCeApumcy/FjdtF+YhywPf3DKTx7<br>
   Pubkey 2: AzQg33JZqH7vSsm09esZY5bZvmzYwE/SY78cA0iLxpD7<br>
   Pubkey 3: A3gOAlB6aiRTCPvWMQg2+ZbGYNsLd8qlvV28m8p2UhY2<br>
   Pubkey 4: Aum2063ub/ErUnIUB36sK55LktGUStgcbSiaAnL1wadu
```

packages/stargate/src/stargateclient.ts
```diff
@@ -155,6 +195,28 @@ export class StargateClient {
     this.tmClient.disconnect();
   }
 
+  public async broadcastTx(tx: cosmos.tx.Tx): Promise<BroadcastTxResponse> {
+    const params = {
+      tx: cosmos.tx.Tx.encode(tx).finish(),
```
webmaster128, line 200: I'd just take a serialized `Uint8Array` here and keep all the encoding at some encoding place.

packages/stargate/src/stargateclient.ts
willclarktech: I was forced to add this by the compiler, so if it can't we need to update the Tendermint type.

willclarktech: Oh, we only have one TM type for success and failure.

packages/stargate/src/stargateclient.ts
willclarktech: https://github.com/CosmWasm/cosmjs/issues/376

packages/stargate/src/stargateclient.ts
webmaster128: We split the success/error case depending on `code` in `StargateClient`. I think this is a good place to `assert(typeof response.height === "number")`. If this throws, something unexpected happems that needs debugging.

willclarktech: It seems like the height is present for both cases, and set to 0 for failure.

webmaster128: Yeah, well, you never know if 0 is omitted in the result from a Go interface

packages/stargate/src/stargateclient.spec.ts
```diff
@@ -182,4 +198,79 @@ describe("StargateClient", () => {
...
+
+  describe("broadcastTx", () => {
+    it("broadcasts a transaction", async () => {
+      pendingWithoutSimapp();
+      const client = await StargateClient.connect(simapp.tendermintUrl);
+      const wallet = await Secp256k1Wallet.fromMnemonic(faucet.mnemonic);
+      const [{ address, pubkey: pubkeyBytes }] = await wallet.getAccounts();
+      const publicKey = PublicKey.create({ secp256k1: pubkeyBytes });
+      const registry = new Registry();
+      const txBodyFields = {
```
ethanfrey, line 210: Great to have this working with all the details.

I bet these build steps could be replaced by a few helper functions and the whole process much easier to use.

willclarktech: I imagine a lot of this will be abstracted away in `SigningStargateClient`.

PR #375, opened 2020-08-12
Add getBlock method to StargateClient

willclarktech:
Part of #272

### DIFF
--- a/packages/stargate/src/stargateclient.spec.ts
+++ b/packages/stargate/src/stargateclient.spec.ts
@@ -3,6 +3,7 @@ import { Bech32, fromBase64 } from "@cosmjs/encoding";
 import { Secp256k1Wallet } from "@cosmjs/launchpad";
 import { makeSignBytes, omitDefaults, Registry } from "@cosmjs/proto-signing";
 import { assert, sleep } from "@cosmjs/utils";
+import { ReadonlyDate } from "readonly-date";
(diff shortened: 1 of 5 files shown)


packages/stargate/src/stargateclient.spec.ts
```diff
@@ -117,6 +125,43 @@ describe("StargateClient", () => {
...
+      expect(response.id).toMatch(tendermintIdMatcher);
+
+      expect(response.header.height).toEqual(height - 1);
+      expect(response.header.chainId).toEqual(await client.getChainId());
+      expect(new ReadonlyDate(response.header.time).getTime()).toBeLessThan(ReadonlyDate.now());
+      expect(new ReadonlyDate(response.header.time).getTime()).toBeGreaterThanOrEqual(
+        ReadonlyDate.now() - 5_000,
+      );
+
+      expect(Array.isArray(response.txs)).toEqual(true);
```
webmaster128, line 161: Jasmine can do this as well `expect(response.txs).toEqual(jasmine.any(Array)`, which allows you to cobine a lot of stuff like:

```ts
expect(response).toEqual(jasmine.objectContaining({
  header: {
    height: height - 1,
    chainId: await client.getChainId(),
  },
  txs: jasmine.any(Array)
});
// time needs to be tested separately
```

PR #378, opened 2020-08-12
Create query client

### LINKED ISSUE #371: Add extensible query layer between tendermint-rpc and StargateClient
Like LcdClient but for stargate

webmaster128:
Closes #371

This is an extensible query client that uses the autogenerated query code.

- [x] Add and test all queries from bank
- [x] Add and use AuthExtension

@ethanfrey looking at this it seems like verifyable queries and autogenerated RPC queries are completely different things. Since the autogenerated code decodes the responses for us, we cannot prove anything. So I wonder if QueryClient should do proof verification at all.


packages/stargate/src/queries/bank.ts
```diff
@@ -0,0 +1,57 @@
...
+        const binAddress = Bech32.decode(address).data;
+        const bankKey = Uint8Array.from([...toAscii("balances"), ...binAddress, ...toAscii(denom)]);
+        const responseData = await base.queryVerified("bank", bankKey);
+        return responseData.length ? toObject(cosmos.Coin.decode(responseData)) : null;
+      },
+      unverified: {
+        allBalances: async (address: string) => {
+          const { balances } = await queryService.allBalances({ address: Bech32.decode(address).data });
+          return balances.map(toObject);
+        },
```
webmaster128, line 53: Once everything is set up, those 4 lines is you need for each query

packages/stargate/src/queries/bank.ts
```diff
@@ -0,0 +1,57 @@
...
+    readonly unverified: {
+      readonly allBalances: (address: string) => Promise<cosmos.ICoin[]>;
+    };
+  };
+}
+
+export function setupBankExtension(base: QueryClient): BankExtension {
+  // Use this service to get easy typed access to query methods
+  // This cannot be used to for proof verification
+  const queryService = cosmos.bank.Query.create((method: any, requestData, callback) => {
```
webmaster128, line 26: The autogenerated class `cosmos.bank.Query` contains all the query methods including types, request encoding and response decoding.

ethanfrey:
> looking at this it seems like verifyable queries and autogenerated RPC queries are completely different things. Since the autogenerated code decodes the responses for us, we cannot prove anything. So I wonder if QueryClient should do proof verification at all.

That is very correct. All the auto-gen / grpc queries call into app logic to process data that is returned. This is like cosmwasm "smart queries", which cannot be proven. However, these are very valuable to:

1. Quickly provide some queries on a module AND
2. Because many queries cannot be done raw (require server-side processing / iteration)

The provable queries use raw key/value like cosmwasm "raw queries". I will work on those separately. Getting the auto-gen "smart queries" working without proofs is already a huge step forward. (And also these often overlap, so this provides a nice test vector for the other kind)

packages/stargate/src/queries/bank.ts
```diff
@@ -0,0 +1,57 @@
+import { Bech32, toAscii } from "@cosmjs/encoding";
+
+import { cosmos } from "../generated/codecimpl";
+import { QueryClient } from "../queryclient";
+
+/**
+ * Use this to convert a protobuf.js class to the interface (e.g. Coin to ICoin)
+ * in a ways that makes Jasmine's toEqual happy.
+ */
+function toObject<I extends object>(thing: I): Omit<I, never> {
```
ethanfrey, line 10: This is only needed for our test code? Or does it serve any purpose in production code as well?

packages/stargate/src/queries/bank.ts
ethanfrey: Very nice!

webmaster128:
> So I wonder if QueryClient should do proof verification at all.

I think it should because of the extensibility of QueryClient. Even if the two types of queries are unrelated, we can have one extension per module grouping verified and unverified queries.

packages/stargate/src/queries/bank.ts
```diff
@@ -0,0 +1,57 @@
...
+        // ("balances", binAddress, denom)
+        // it seem like prefix stores just do a dumb concat with the keys (no tricks to avoid overlap)
+        // https://github.com/cosmos/cosmos-sdk/blob/2879c0702c87dc9dd828a8c42b9224dc054e28ad/store/prefix/store.go#L61-L64
+        // https://github.com/cosmos/cosmos-sdk/blob/2879c0702c87dc9dd828a8c42b9224dc054e28ad/store/prefix/store.go#L37-L43
+        const binAddress = Bech32.decode(address).data;
+        const bankKey = Uint8Array.from([...toAscii("balances"), ...binAddress, ...toAscii(denom)]);
+        const responseData = await base.queryVerified("bank", bankKey);
+        return responseData.length ? toObject(cosmos.Coin.decode(responseData)) : null;
+      },
+      unverified: {
```
ethanfrey, line 49: I like the grouping.

Can you also add a few more queries to unverified? There are 2 supply queries: https://github.com/cosmos/cosmos-sdk/blob/master/proto/cosmos/bank/v1beta1/query.proto#L18-L22

(uh, oh they moved the protobuf files and changed the package name `v1beta1`)

packages/stargate/src/queryclient.ts
```diff
@@ -0,0 +1,183 @@
+/* eslint-disable no-dupe-class-members, @typescript-eslint/ban-types, @typescript-eslint/naming-convention */
+import { toHex } from "@cosmjs/encoding";
+import { Client as TendermintClient } from "@cosmjs/tendermint-rpc";
+import { arrayContentEquals, assert, isNonNullObject } from "@cosmjs/utils";
+
+type QueryExtensionSetup<P> = (base: QueryClient) => P;
+
+export class QueryClient {
```
ethanfrey, line 8: Nice, just like Launchpad composition

packages/stargate/src/queries/bank.ts
```diff
@@ -0,0 +1,57 @@
...
+export interface BankExtension {
+  readonly bank: {
+    readonly balance: (address: string, denom: string) => Promise<cosmos.ICoin | null>;
+    readonly unverified: {
+      readonly allBalances: (address: string) => Promise<cosmos.ICoin[]>;
+    };
+  };
+}
+
+export function setupBankExtension(base: QueryClient): BankExtension {
```
ethanfrey, line 23: Loving this closure encapsulation.

packages/stargate/src/stargateclient.ts
```diff
@@ -94,6 +96,7 @@ export interface PrivateStargateClient {
 
 export class StargateClient {
   private readonly tmClient: TendermintClient;
+  private readonly queryClient: QueryClient & BankExtension;
```
ethanfrey, line 99: Why is this private?

The QueryClient is beautifully extensible, but now we need to explicitly add mappings like `getBalance` below to use it?

I guess we need that for the typings? QueryClient gives us raw data from grpc, and this converts it to some types we control?

packages/stargate/src/queryclient.ts
```diff
@@ -0,0 +1,183 @@
...
+    return client;
+  }
+
+  private readonly tmClient: TendermintClient;
+
+  public constructor(tmClient: TendermintClient) {
+    this.tmClient = tmClient;
+  }
+
+  public async queryVerified(store: string, key: Uint8Array): Promise<Uint8Array> {
```
ethanfrey, line 147: Cool to add this here, and yes, I didn't forger the proof verification

packages/stargate/src/queries/bank.ts
webmaster128: It normalizes all protobuf classes to objects with the same properties and values. We need it for testing but it is nice in general to get consist runtime types.

packages/stargate/src/queries/bank.ts
webmaster128: This was work in progress. Now all bank and all auth queries are implemented

packages/stargate/src/stargateclient.ts
webmaster128: Exactly, the types from the autogenerated protobuf are a pain to work with. You don't want to throw `number | Long | null | undefined` on the user for a field that is non-nullable.

packages/stargate/src/queries/auth.ts
```diff
@@ -0,0 +1,60 @@
...
+    const path = `/cosmos.auth.Query/${method.name}`;
+    base
+      .queryUnverified(path, requestData)
+      .then((response) => callback(null, response))
+      .catch((error) => callback(error));
+  });
+
+  return {
+    auth: {
+      account: async (address: string) => {
```
ethanfrey, line 30: I like this... unverified is more or less autogen, and all the other proven, handl-built ones are top-level.

Good stuff

packages/stargate/src/queries/auth.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+        assert(account);
+        expect(account).toEqual({
+          address: toAccAddress(unused.address),
+          // pubKey not set
+          accountNumber: Long.fromNumber(unused.accountNumber, true),
+          // sequence not set
+        });
+      });
+
+      it("works for account with pubkey and non-zero sequence", async () => {
```
ethanfrey, line 72: Nice to test both cases (unused and w/ pubkey) :100:

packages/stargate/src/queries/bank.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+          },
+          {
+            amount: jasmine.stringMatching(nonNegativeIntegerMatcher),
+            denom: simapp.denomStaking,
+          },
+        ]);
+      });
+    });
+
+    describe("totalSupply", () => {
```
ethanfrey, line 141: s/totalSupply/supplyOf/

(code good, describe name off)

packages/stargate/src/queries/bank.ts
```diff
@@ -0,0 +1,67 @@
...
+        // https://github.com/cosmos/cosmos-sdk/blob/2879c0702c87dc9dd828a8c42b9224dc054e28ad/store/prefix/store.go#L61-L64
+        // https://github.com/cosmos/cosmos-sdk/blob/2879c0702c87dc9dd828a8c42b9224dc054e28ad/store/prefix/store.go#L37-L43
+        const key = Uint8Array.from([...toAscii("balances"), ...toAccAddress(address), ...toAscii(denom)]);
+        const responseData = await base.queryVerified("bank", key);
+        return responseData.length ? toObject(cosmos.Coin.decode(responseData)) : null;
+      },
+      unverified: {
+        balance: async (address: string, denom: string) => {
+          const { balance } = await queryService.balance({ address: toAccAddress(address), denom: denom });
+          assert(balance);
```
ethanfrey, line 48: Does this work if they have no balance for the denom?
assert will fail on undefined, but succeed on 0?

packages/stargate/src/queries/bank.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+        });
+      });
+
+      it("returns zero for non-existent balance", async () => {
+        pendingWithoutSimapp();
+        const client = await makeBankClient(simapp.tendermintUrl);
+
+        const response = await client.bank.unverified.balance(unused.address, "gintonic");
+        expect(response).toEqual({
+          amount: "0",
```
ethanfrey, line 78: Ahh, we get a proper zero-ed object if missing. Good stuff.

packages/stargate/src/queries/bank.ts
webmaster128: `balance` is an `ICoin`, so it is always truthy no matter what the content is

PR #384, opened 2020-08-13
Internalize SignDoc and omitDefaults for sign bytes creation

webmaster128:
This is only the first step of an effort of keeping all types for signing in @cosmjs/proto-signing.

### DIFF
--- a/packages/proto-signing/src/index.ts
+++ b/packages/proto-signing/src/index.ts
@@ -1,4 +1,3 @@
-export { omitDefaults } from "./adr27";
 export { Coin } from "./msgs";
 export { cosmosField } from "./decorator";
 export { Registry } from "./registry";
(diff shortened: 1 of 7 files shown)


packages/proto-signing/src/index.ts
```diff
@@ -1,4 +1,3 @@
-export { omitDefaults } from "./adr27";
```
ethanfrey, line 1: One less external exposed.

packages/proto-signing/src/signing.spec.ts
```diff
@@ -51,7 +50,7 @@ const testVectors = [
   },
 
   {
-    sequenceNumber: 2,
+    sequence: 2,
```
ethanfrey, line 53: I always love shorter names

packages/proto-signing/src/signing.ts
```diff
@@ -1,8 +1,24 @@
 /* eslint-disable @typescript-eslint/naming-convention */
+import { omitDefaults } from "./adr27";
 import { cosmos } from "./generated/codecimpl";
 
 const { SignDoc } = cosmos.tx;
 
-export function makeSignBytes(signDoc: cosmos.tx.ISignDoc): Uint8Array {
+export function makeSignBytes(
```
ethanfrey, line 7: This is super helpful. Then we just need helpers to build `txBody` and `authInfo`

PR #386, opened 2020-08-14
Verify ics23 proofs

### LINKED ISSUE #347: Integrate ICS23 proof validation for state queries in client
(not including header validation)

- Spec: https://github.com/cosmos/ics/tree/master/spec/ics-023-vector-commitments
- Implementation: https://github.com/confio/ics23

to be integrated in `packages/stargate/src/stargateclient.ts`:

```ts
  private async queryVerified(store: string, key: Uint8Array): Promise<Uint8Array> {
    const response = await this.tmClient.abciQuery({
      // we need the StoreKey for the module, not the module name
      // https://github.com/cosmos/cosmos-sdk/blob/8cab43c8120fec5200c3459cbf4a92017bb6f287/x/auth/types/keys.go#L12
      path: `/store/${store}/key`,
      data: key,
      prove: true,
    });

    if (response.code) {
      throw new Error(`Query failed with (${response.code}): ${response.log}`);
    }

    if (!arrayContentEquals(response.key, key)) {
      throw new Error(`Response key ${toHex(response.key)} doesn't match query key ${toHex(key)}`);
    }

    // TODO: implement proof verification
    // https://github.com/CosmWasm/cosmjs/issues/347

    return response.value;
  }
```

ethanfrey:
Closes #347 

- [x] Parses proof info in tendermint rpc
- [x] Validate proper sdk proof format in `queryVerified`
- [x] Validate the ics23 proof itself

### DIFF
--- a/packages/stargate/src/queries/queryclient.ts
+++ b/packages/stargate/src/queries/queryclient.ts
@@ -227,8 +227,12 @@ export class QueryClient {
     if (height == 0) {
       throw new Error("Query returned height 0, cannot prove it");
     }
-    // get the header for height+1
+
+    // get the header for height+1 with events
     const header = await firstEvent(this.tmClient.subscribeNewBlockHeader());
+    // alternate non-websocket query (fails with error as it is the future)
+    // const header = (await this.tmClient.blockchain(height+1, height+1)).blockMetas[0].header;
+
     if (header.height !== height + 1) {
       throw new Error(`Query returned height ${height}, but next header was ${header.height}`);
     }
--- a/packages/stargate/src/queries/queryclient.ts
+++ b/packages/stargate/src/queries/queryclient.ts
@@ -229,6 +229,7 @@ export class QueryClient {
       throw new Error("Query returned height 0, cannot prove it");
     }
 
+    const searchHeight = height + 1;
     let nextHeader: Header;
     let headersSubscription: Stream<NewBlockHeaderEvent> | undefined;
     try {
@@ -241,17 +242,25 @@ export class QueryClient {
       // get the header for height+1
       nextHeader = await firstEvent(headersSubscription); // TODO: fall back on polling if this returns a too high header
     } else {
-      let metas = (await this.tmClient.blockchain(height + 1, height + 1)).blockMetas;
-      while (metas.length === 0) {
+      // start from current height to avoid backend error for minHeight in the future
+      let header = (await this.tmClient.blockchain(height, searchHeight)).blockMetas
+        .map((meta) => meta.header)
+        .find((h) => h.height === searchHeight);
+      while (!header) {
         await sleep(1000);
-        metas = (await this.tmClient.blockchain(height + 1, height + 1)).blockMetas;
+        header = (await this.tmClient.blockchain(height, searchHeight)).blockMetas
+          .map((meta) => meta.header)
+          .find((h) => h.height === searchHeight);
       }
-      nextHeader = metas[0].header;
+      nextHeader = header;
     }
 
-    if (nextHeader.height !== height + 1) {
-      throw new Error(`Query returned height ${height}, but next header was ${nextHeader.height}`);
+    if (nextHeader.height !== searchHeight) {
+      throw new Error(
+        `Query requires header at height ${searchHeight} for proof verification, but next header was ${nextHeader.height}`,
+      );
     }
+
     return nextHeader;
   }
 }
--- a/packages/stargate/src/queries/queryclient.ts
+++ b/packages/stargate/src/queries/queryclient.ts
@@ -1,7 +1,43 @@
 /* eslint-disable no-dupe-class-members, @typescript-eslint/ban-types, @typescript-eslint/naming-convention */
-import { toHex, fromAscii } from "@cosmjs/encoding";
+import { ics23, verifyExistence } from "@confio/ics23";
+import { fromAscii, toHex } from "@cosmjs/encoding";
 import { Client as TendermintClient } from "@cosmjs/tendermint-rpc";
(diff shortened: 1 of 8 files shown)


ethanfrey:
@webmaster128 I am almost there.

The missing step is "get the header and verify with appHash". The issue is for a query that returns state after block H, I need the appHash from block H+1. There are 2 issues here:

1. We need a cleaner way than `sleep(5000)` to wait for the next header.
2. I don't see any header queries on tendermintRpc, just the block query.

Maybe I am missing something. With some tips on the tendermint rpc header stuff, I am pretty sure I can get this all finished - no significant roadblocks yet

ethanfrey:
The work is done. It can probably be refactored and use some more targeted tests.

Let me know what improvements you would suggest

packages/stargate/src/queries/queryclient.ts
```diff
@@ -1,8 +1,45 @@
 /* eslint-disable no-dupe-class-members, @typescript-eslint/ban-types, @typescript-eslint/naming-convention */
-import { toHex } from "@cosmjs/encoding";
+import { ics23, verifyExistence, verifyNonExistence } from "@confio/ics23";
+import { fromAscii, toAscii, toHex } from "@cosmjs/encoding";
+import { firstEvent } from "@cosmjs/stream";
 import { Client as TendermintClient } from "@cosmjs/tendermint-rpc";
 import { arrayContentEquals, assert, isNonNullObject } from "@cosmjs/utils";
 
+// TODO: import these from @confio/ics23 when exported
```
webmaster128, line 8: Do you create an update for this?

packages/stargate/src/queries/queryclient.ts
```diff
@@ -1,8 +1,45 @@
...
+export const IavlSpec: ics23.IProofSpec = {
```
webmaster128, line 9: This should be lower case

packages/stargate/src/queries/queryclient.ts
```diff
@@ -1,8 +1,45 @@
...
+  innerSpec: {
+    childOrder: [0, 1],
+    minPrefixLength: 4,
+    maxPrefixLength: 12,
+    childSize: 33,
+    hash: ics23.HashOp.SHA256,
+  },
+};
+
+export const TendermintSpec: ics23.IProofSpec = {
```
webmaster128, line 26: here too

packages/tendermint-rpc/src/v0-33/responses.ts
```diff
@@ -62,7 +82,7 @@ function decodeAbciQuery(data: RpcAbciQueryResponse): responses.AbciQueryRespons
   return {
     key: Base64.decode(optional(data.key, "" as Base64String)),
     value: Base64.decode(optional(data.value, "" as Base64String)),
-    // proof: may(Base64.decode, data.proof),
+    proof: may(decodeQueryProof, data.proof),
```
webmaster128, line 85: I think we should test this field in a plain Tendermint test. Is this possible with our testing `--proxy_app=kvstore`?

webmaster128:
> I don't see any header queries on tendermintRpc, just the block query.

`tmClient.blockchain` is probably what you are looking for

packages/tendermint-rpc/src/v0-33/responses.ts
ethanfrey: There are no proofs there. We can get undefined.

The actual format we parse is only from the cosmos-sdk

ethanfrey:
> tmClient.blockchain is probably what you are looking for

Ah, good stuff.  I will look there. The `nextHeader` subscription worked out quite well actually.

webmaster128:
> The `nextHeader` subscription worked out quite well actually.

This makes the whole Query client depend on a WebSocket connection. It would be better to avoid requiring that. It might be nice as a speed optimization.

packages/stargate/src/queries/queryclient.ts
ethanfrey: Published https://www.npmjs.com/package/@confio/ics23/v/0.6.3

ethanfrey:
I made all requested changes and some other cleanup.

> This makes the whole Query client depend on a WebSocket connection. It would be better to avoid requiring that. It might be nice as a speed optimization.

I tried with the other query and it doesn't work, I need to wait on the chain. Do you really want a polling loop there? (See https://github.com/CosmWasm/cosmjs/pull/386/commits/51d07b3f154d20097596b2e7a2e735db352e1d09)

webmaster128:
The CI fails because of https://github.com/CosmWasm/cosmjs/pull/394



> I tried with the other query and it doesn't work, I need to wait on the chain. Do you really want a polling loop there? (See [51d07b3](https://github.com/CosmWasm/cosmjs/commit/51d07b3f154d20097596b2e7a2e735db352e1d09))

What I really want i a query that can be verified without waiting seconds for the next block. But if we don't want to give up HTTP support, we need a polling fallback. I implemented one in #395. Once this is merged in here, this is good to go from my point of view.

ethanfrey:
> What I really want i a query that can be verified without waiting seconds for the next block. 

This is not possible with the sdk / tendermint design. Block H includes a list of transactions *to be executed* and the AppHash resulting from *executing the transactions in H-1*.

The only way to do this without a sleep/wait is to query the previous block. So, if we get header H as the most recent, then query on height H-1. This may be a bit surprising to a user, as it may give a different result than an unverified query, but also is much faster (no waiting for a block). We just need to document the approach and then do it.

ethanfrey:
Rebased on #394 to fix test error

webmaster128:
> This is not possible with the sdk / tendermint design.

This is what I suspected. Let's just keep subscribe or poll depending on what is available

webmaster128:
I'll finalize and merge this now

PR #389, opened 2020-08-17
Improve docs and tests of TendermintClient.blockchain

webmaster128:
Code 💅 to support #386

### DIFF
--- a/packages/tendermint-rpc/src/client.spec.ts
+++ b/packages/tendermint-rpc/src/client.spec.ts
@@ -14,8 +14,12 @@ import * as responses from "./responses";
 import { HttpClient, RpcClient, WebsocketClient } from "./rpcclients";
 import { TxBytes } from "./types";
 
+function tendermintEnabled(): boolean {
+  return !!process.env.TENDERMINT_ENABLED;
+}
+
 function pendingWithoutTendermint(): void {
-  if (!process.env.TENDERMINT_ENABLED) {
+  if (!tendermintEnabled()) {
     pending("Set TENDERMINT_ENABLED to enable tendermint-based tests");
   }
 }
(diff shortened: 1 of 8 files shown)


packages/tendermint-rpc/src/client.spec.ts
```diff
@@ -124,118 +129,217 @@ function defaultTestSuite(rpcFactory: () => RpcClient, adaptor: Adaptor): void {
...
-    const client = new Client(rpcFactory(), adaptor);
+      client.disconnect();
+    });
+
+    it("can limit by maxHeight", async () => {
+      pendingWithoutTendermint();
+      const client = new Client(rpcFactory(), adaptor);
+
+      const height = (await client.status()).syncInfo.latestBlockHeight;
+      const blockchain = await client.blockchain(undefined, height - 1);
```
ethanfrey, line 172: Ah, this will get any sequence of blocks. And from the future?

packages/tendermint-rpc/src/client.spec.ts
```diff
@@ -124,118 +129,217 @@ function defaultTestSuite(rpcFactory: () => RpcClient, adaptor: Adaptor): void {
...
-    expect(block.block.txs[0]).toEqual(tx);
+      // Run in parallel to increase chance there is no block between the calls
+      const [status, blockchain] = await Promise.all([client.status(), client.blockchain()]);
+      const height = status.syncInfo.latestBlockHeight;
 
-    client.disconnect();
-  });
+      expect(blockchain.lastHeight).toEqual(height);
+      expect(blockchain.blockMetas.length).toBeGreaterThanOrEqual(3);
+      expect(blockchain.blockMetas[0].header.height).toEqual(height);
```
ethanfrey, line 160: No args is "get all blocks up to the current one?"

packages/tendermint-rpc/src/client.spec.ts
ethanfrey: > I would ask for a test that given status queries blockchain(height, height+1).

I tested this in my proof code. It throws an error like:

`Error: {"code":-32603,"message":"Internal error","data":"min height 1005 can't be greater than max height 1004"}`

webmaster128: I'll test future blocks. But I'm pretty sure this is query+filtering.

The min > max case should return an empty result instead of an error in my opinion. But the error is more helpful for debugging since this is not intended usually.

packages/tendermint-rpc/src/client.spec.ts
webmaster128: No args defaults to minHeight=0 and maxHeight=Infinity. I wonder if we should make the arguments non-optional to enforce setting them explicitly (from `blockchain(minHeight?: number, maxHeight?: number)` to `blockchain(minHeight: number | undefined, maxHeight: number | undefined)`)

packages/tendermint-rpc/src/client.spec.ts
ethanfrey: The error was not a `min > max` case (which can throw error), but `max > current height` case

webmaster128: `max > current height` is not the problem (added a test here). But I found the problem: `min > current height`

webmaster128: In #395 I worked around it using

```ts
      // start from current height to avoid backend error for minHeight in the future
      let header = (await this.tmClient.blockchain(height, searchHeight)).blockMetas
        .map((meta) => meta.header)
        .find((h) => h.height === searchHeight);
```

PR #398, opened 2020-08-18
Fallback on header query if subscription event is the wrong one

### LINKED ISSUE #397: Fix cases where the header we get from the subscription is too high
Errors like

```
  1 AuthExtension

    1.1 account
      ✗ works for unused account (1 sec)
        - Error: Query requires header at height 190 for proof verification, but next header was 191

**************************************************
*                    Failures                    *
**************************************************

1) AuthExtension account works for unused account
  - Error: Query requires header at height 190 for proof verification, but next header was 191

  /home/circleci/project/packages/stargate/src/queries/queryclient.ts:19:2
    }
    return ics23.CommitmentProof.decode(op.data);
  }
   ~
```

(https://app.circleci.com/pipelines/github/CosmWasm/cosmjs/1198/workflows/848f8abe-ea7c-49fc-9ce6-ed069e1553e6/jobs/4706)

can happen when the query happens at height n, the subscription is established and the first event is we get is not n+1 but n+2 or higher.

This is already documented in

```ts
    if (headersSubscription) {
      // get the header for height+1
      nextHeader = await firstEvent(headersSubscription); // TODO: fall back on polling if this returns a too high header
```

webmaster128:
Closes #397

### DIFF
--- a/packages/stargate/src/queries/queryclient.ts
+++ b/packages/stargate/src/queries/queryclient.ts
@@ -246,26 +246,19 @@ export class QueryClient {
       }
     }
 
-    if (!nextHeader) {
+    while (!nextHeader) {
       // start from current height to avoid backend error for minHeight in the future
-      let header = (await this.tmClient.blockchain(height, searchHeight)).blockMetas
+      const correctHeader = (await this.tmClient.blockchain(height, searchHeight)).blockMetas
         .map((meta) => meta.header)
         .find((h) => h.height === searchHeight);
-      while (!header) {
+      if (correctHeader) {
+        nextHeader = correctHeader;
+      } else {
         await sleep(1000);
-        header = (await this.tmClient.blockchain(height, searchHeight)).blockMetas
-          .map((meta) => meta.header)
-          .find((h) => h.height === searchHeight);
       }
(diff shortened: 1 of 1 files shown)


packages/stargate/src/queries/queryclient.ts
```diff
@@ -239,28 +239,26 @@ export class QueryClient {
...
+      // The first header we get might not be n+1 but n+2 or even higher. In such cases we fall back on a query.
+      if (firstHeader.height === searchHeight) {
+        nextHeader = firstHeader;
+      }
+    }
+
+    while (!nextHeader) {
       // start from current height to avoid backend error for minHeight in the future
-      let header = (await this.tmClient.blockchain(height, searchHeight)).blockMetas
+      const correctHeader = (await this.tmClient.blockchain(height, searchHeight)).blockMetas
```
willclarktech, line 251: You could just directly assign this to `nextHeader` since in the fail case `find()` returns `undefined`.

webmaster128: hmm, yeah, that could be done. However, I think it is easier to assign `nextHeader` only once when found.

PR #399, opened 2020-08-18
Update Fee tables

### LINKED ISSUE #396: Update Fee tables to take a global gas price + gas limit per operation
Right now the fee tables looks like

```ts
const defaultFees: FeeTable = {
  upload: {
    amount: coins(25000, "ucosm"),
    gas: "1000000", // one million
  },
  init: {
    amount: coins(12500, "ucosm"),
    gas: "500000", // 500k
  },
  migrate: {
    amount: coins(5000, "ucosm"),
    gas: "200000", // 200k
  },
  exec: {
    amount: coins(5000, "ucosm"),
    gas: "200000", // 200k
  },
  send: {
    amount: coins(2000, "ucosm"),
    gas: "80000", // 80k
  },
  changeAdmin: {
    amount: coins(2000, "ucosm"),
    gas: "80000", // 80k
  },
};
```

The values are a bit arbitrary and work for testnets where everything is free. Due to the data structure we use, the price per gas differs for every operation.

What we want instead is a single gas price that is then multiplied by the gas limit. Ideally this is implemented by a `Coin` with `Decimal` multiplication.

This is inspired by [buildFeeTable](https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw20-base/helpers.ts#L66-L84)

willclarktech:
Closes #396

Opening this draft PR now to get feedback on the design. Still to do:

- Update and use `Decimal` class
- Apply same design to `SigningCosmWasmClient`

### DIFF
--- a/packages/cosmwasm/src/signingcosmwasmclient.ts
+++ b/packages/cosmwasm/src/signingcosmwasmclient.ts
@@ -7,7 +7,7 @@ import {
   BroadcastTxResult,
   buildFeeTable,
   Coin,
-  FeeTable,
+  CosmosFeeTable,
   GasLimits,
   GasPrice,
   isBroadcastTxFailure,
@@ -36,12 +36,11 @@ import {
 /**
  * These fees are used by the higher level methods of SigningCosmWasmClient
  */
-export interface CosmWasmFeeTable extends FeeTable {
+export interface CosmWasmFeeTable extends CosmosFeeTable {
   readonly upload: StdFee;
   readonly init: StdFee;
   readonly exec: StdFee;
   readonly migrate: StdFee;
-  readonly send: StdFee;
   /** Paid when setting the contract admin to a new address or unsetting it */
   readonly changeAdmin: StdFee;
 }
@@ -153,7 +152,7 @@ export class SigningCosmWasmClient extends CosmWasmClient {
   public readonly senderAddress: string;
 
   private readonly signer: OfflineSigner;
-  private readonly fees: FeeTable;
+  private readonly fees: CosmWasmFeeTable;
 
   /**
    * Creates a new client with signing capability to interact with a CosmWasm blockchain. This is the bigger brother of CosmWasmClient.
--- a/packages/cosmwasm/src/signingcosmwasmclient.ts
+++ b/packages/cosmwasm/src/signingcosmwasmclient.ts
@@ -5,8 +5,11 @@ import {
   BroadcastMode,
   BroadcastTxFailure,
   BroadcastTxResult,
+  buildFeeTable,
   Coin,
-  coins,
+  FeeTable,
+  GasLimits,
+  GasPrice,
   isBroadcastTxFailure,
   makeSignBytes,
   Msg,
@@ -31,9 +34,9 @@ import {
 } from "./msgs";
 
 /**
- * Those fees are used by the higher level methods of SigningCosmWasmClient
+ * These fees are used by the higher level methods of SigningCosmWasmClient
  */
-export interface FeeTable {
+export interface CosmWasmFeeTable extends FeeTable {
   readonly upload: StdFee;
   readonly init: StdFee;
   readonly exec: StdFee;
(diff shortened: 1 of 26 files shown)


packages/launchpad/src/signingcosmosclient.ts
```diff
@@ -14,13 +14,38 @@ export interface FeeTable {
...
 
-const defaultFees: FeeTable = {
-  send: {
-    amount: coins(2000, "ucosm"),
-    gas: "80000", // 80k
-  },
+export interface FeeOptions {
+  readonly feeToken: string;
+  readonly gasPrice: number;
+}
```
webmaster128, line 20: This can be changed to a `GasPrice` class with a `denom` and an `amount` (similar to coin, but decimal). Then we can create a `GasPrice.fromString()` parser that taked the same represenation as in the network config:

```toml
# The minimum gas prices a validator is willing to accept for processing a
# transaction. A transaction's fees must meet the minimum of any denomination
# specified in this config (e.g. 0.25token1;0.0001token2).
minimum-gas-prices = "0.025ushell"
```

We only need one denom per network.

packages/launchpad/src/signingcosmosclient.spec.ts
```diff
@@ -27,6 +27,68 @@ describe("SigningCosmosClient", () => {
...
     });
+
+    it("can be constructed with custom fees", async () => {
+      const wallet = await Secp256k1Wallet.fromMnemonic(faucet.mnemonic);
+      const feeOptions = {
+        feeToken: "utest",
+        gasPrice: 3.14,
+      };
+      const client = new SigningCosmosClient(httpUrl, faucet.address, wallet, feeOptions);
+      expect((client as any).fees).toEqual({
```
webmaster128, line 38: Goo to test this. You can create a `PrivateSigningCosmosClient` similar to `PrivateCosmWasmClient` that allows you to access this in a type safe way.

packages/launchpad/src/signingcosmosclient.ts
```diff
@@ -14,13 +14,38 @@ export interface FeeTable {
...
+
+export type GasLimits = {
+  readonly [key in keyof FeeTable]: number;
 };
 
+function stdFee(gasLimit: number, denom: string, price: number): StdFee {
```
webmaster128, line 26: rename to `calculateFee`?

packages/launchpad/src/signingcosmosclient.ts
```diff
@@ -43,15 +68,21 @@ export class SigningCosmosClient extends CosmosClient {
     apiUrl: string,
     senderAddress: string,
     signer: OfflineSigner,
-    customFees?: Partial<FeeTable>,
+    feeOptions: FeeOptions = defaultFeeOptions,
+    gasLimits?: Partial<GasLimits>,
```
webmaster128, line 72: Ah nice split. Makes changing the token only very simple

packages/launchpad/src/signingcosmosclient.ts
willclarktech: Do you know if there's a spec for that format somewhere?

webmaster128: Hmm, no. The only thing I am aware of is this change: https://github.com/cosmos/cosmos-sdk/pull/7027. So we can use any `[a-z]` as the first character for the split.

webmaster128: Once you have the split position, you can just use `Decimal.fromUserInput("0.025", 18)`, which is the Cosmos SDK precision for decimaps afaik.

packages/cosmwasm/src/signingcosmwasmclient.ts
```diff
@@ -31,9 +34,9 @@ import {
 } from "./msgs";
 
 /**
- * Those fees are used by the higher level methods of SigningCosmWasmClient
+ * These fees are used by the higher level methods of SigningCosmWasmClient
  */
-export interface FeeTable {
+export interface CosmWasmFeeTable extends FeeTable {
```
webmaster128, line 39: If we use extend here (assuming the CosmWasmClient can do everything the CosmosClient can), then we can remove the `send` field here, right?

packages/launchpad/src/index.ts
```diff
@@ -29,6 +29,7 @@ export {
   isSearchByTagsQuery,
 } from "./cosmosclient";
 export { makeSignBytes } from "./encoding";
+export { buildFeeTable, FeeTable, GasLimits, GasPrice } from "./gas";
```
webmaster128, line 32: Do we need to export `buildFeeTable`? My impression was this becomes a fully internal calculation with the new API

packages/cosmwasm/src/signingcosmwasmclient.ts
willclarktech: Yes. Do you think we have any need for the completely generic `FeeTable` (e.g. to support clients which turn off some Cosmos SDK functionality)?

packages/launchpad/src/index.ts
willclarktech: I import it in CosmWasm. Should I be copying it over into that package instead?

packages/cosmwasm/src/signingcosmwasmclient.ts
webmaster128: No, this is good for now.

I wanted to have CosmWasmClient extending CosmosClient but that does not work because there is no multi inheritance in JS and we need inheritance for SigningCosmWasmClient and SigningCosmosClient.

packages/launchpad/src/index.ts
webmaster128: I see, good.

webmaster128 (approved):
Beautiful

packages/cosmwasm/src/signingcosmwasmclient.ts
willclarktech: OK, well we have `FeeTable` anyway. 😄

PR #401, opened 2020-08-18
Add IbcExtension with unverified queries

webmaster128:


packages/stargate/src/queries/ibc.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+async function makeClientWithIbc(rpcUrl: string): Promise<[QueryClient & IbcExtension, TendermintClient]> {
+  const tmClient = await TendermintClient.connect(rpcUrl);
+  return [QueryClient.withExtensions(tmClient, setupIbcExtension), tmClient];
+}
+
+describe("IbcExtension", () => {
+  describe("unverified", () => {
+    describe("channel", () => {
+      it("can be called", async () => {
+        pending("Fails with 'Query failed with (1): internal'. Make it work.");
```
ethanfrey, line 16: Ugly. Can you run `--trace` on the `simd` binary to get better output there, so we can make issues on cosmos-sdk?

ethanfrey: I just checked the code and this does turn off redact and add stack trace on master as well as launchpad

packages/stargate/src/queries/ibc.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+      });
+    });
+
+    describe("packetAcknowledgement", () => {
+      it("can be called", async () => {
+        pending("Fails with 'Query failed with (1): internal'. Make it work.");
+        pendingWithoutSimapp();
+        const [client, tmClient] = await makeClientWithIbc(simapp.tendermintUrl);
+
+        const response = await client.ibc.unverified.packetAcknowledgement();
```
ethanfrey, line 84: A lot of these need arguments, and need data to properly test. I will reach out to Informal Systems on this one

webmaster128:
> A lot of these need arguments, and need data to properly test. I will reach out to Informal Systems on this one

Yeah, this is the main issue I have here. It compiles but all of this has no meaning to me. I have no clue what to expect and what to pass. It would be good if you can take over this to show some initial use cases.

packages/stargate/src/queries/ibc.spec.ts
webmaster128: This is done already (see `scripts/simapp/template/run_simd.sh`)

packages/stargate/src/queries/ibc.spec.ts
webmaster128: > A lot of these need arguments

I added all query parameters except the pagination for lists (which I hope work with unset for the first data)

ethanfrey:
I looked at what Informal Systems had for tests in their ibc-rs crate. Seems like none. It also referenced the GoZ relayer, so rather confused which version they are targetting.

I found the PR they made to add random ibc data to genesis: https://github.com/cosmos/cosmos-sdk/pull/6637/files but it doesn't seem complete.

What we need is either:

1. A set of commands to run on the start of the chain to generate ibc client/connection/packets for a simple ibctransfer connection OR
2. A dump of such commands to pre-populated genesis file that we can import on start. This would allow us to use go scripts/tests to generate it and not just `simd` cli and likely more feasible.

@alpe I would appreciate your help very much in generating such data for these tests, I am happy to work on the cosmjs tests and queries themselves.

alpe:
Added a PR with some testdata: https://github.com/CosmWasm/cosmjs/pull/405

webmaster128:
Merging this now to
1. ensure what compiles now keeps compiling when updating simapp
2. show where the IBC queries belong
3. show an example of a module with more than one query service

PR #402, opened 2020-08-18
Test QueryClient directly

webmaster128:
Before, QueryClient was only tested via extensions. This adds tests for its two function via both WS and http transport (which matters for the next block header logic)

### DIFF
--- a/scripts/tendermint/start.sh
+++ b/scripts/tendermint/start.sh
@@ -26,7 +26,7 @@ docker run --rm \
 
 # make sure we allow cors origins, only possible by modifying the config file
 # https://github.com/tendermint/tendermint/issues/3216
-sed -ie 's/^cors_allowed_origins.*$/cors_allowed_origins = ["*"]/' "${TMP_DIR}/config/config.toml"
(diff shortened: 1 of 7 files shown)


scripts/tendermint/start.sh
```diff
@@ -26,7 +26,7 @@ docker run --rm \
 
 # make sure we allow cors origins, only possible by modifying the config file
 # https://github.com/tendermint/tendermint/issues/3216
-sed -ie 's/cors_allowed_origins.*$/cors_allowed_origins = ["*"]/' "${TMP_DIR}/config/config.toml"
+sed -ie 's/^cors_allowed_origins.*$/cors_allowed_origins = ["*"]/' "${TMP_DIR}/config/config.toml"
```
willclarktech, line 29: Might want to use something like this for future-proofing?
```
sed -ie 's/^cors_allowed_origins = .*$/cors_allowed_origins = ["*"]/' "${TMP_DIR}/config/config.toml"
```

webmaster128: Indeed. The space and `=` works fine for the `timeout_*` variables in `scripts/simapp/generate_template.sh`

PR #405, opened 2020-08-19
Seed simd with some ibc data

alpe:
Some IBC seed data for testing purpose

🚧  This works with a custom simd docker artifact built of cosmos-sdk master ('f02b0b5')
Latest nightly does not work as there were some fixes merged meanwhile

The IBC testdata is also in `genesis_ibc_testdata.json`. You may want to merge that in during template generation.


ethanfrey:
Tried  `scripts/simapp/start.sh` and got

`Unable to find image 'alpetest/simd:f02b0b5' locally`

I will try with current nightly and see what happens

alpe:
Sorry, the repo is public now

PR #407, opened 2020-08-19
Use denom in faucet

### LINKED ISSUE #391: Give up display tokens in the faucet
The faucet is the only place in CosmJS where we still have the BCP token tickers supported (e.g. request SHELL instead of ushell). This causes a lot of effort to get right in the faucet and it also seems to confuse testnet users trying to request the base denom.

What about removing

```ts
export interface BankTokenMeta {
  readonly denom: string;
  /**
   * The token ticker symbol, e.g. ATOM or ETH.
   */
  readonly tickerSymbol: string;
  /**
   * The number of fractional digits the token supports.
   *
   * A quantity is expressed as atomic units. 10^fractionalDigits of those
   * atomic units make up 1 token.
   *
   * E.g. in Ethereum 10^18 wei are 1 ETH and from the quantity 123000000000000000000
   * the last 18 digits are the fractional part and the rest the wole part.
   */
}
```

from the faucet entirely and leave it up to UI applications to convert between the base denom and a display token (where the logic needs to exist anyways for all the other APIs).

So instead of

```sh
JSON=$(jq -n --arg addr $(coral keys show -a fred) '{"ticker":"COSM", "address":$addr}') \
  && curl -X POST --header "Content-Type: application/json" --data "$JSON" https://faucet.coralnet.cosmwasm.com/credit
```

you'd

```sh
JSON=$(jq -n --arg addr $(coral keys show -a fred) '{"denom":"ucosm", "address":$addr}') \
  && curl -X POST --header "Content-Type: application/json" --data "$JSON" https://faucet.coralnet.cosmwasm.com/credit
```

@ethanfrey @orkunkl

willclarktech:
Closes #391

I wasn't sure about the fractional digits for Coralnet, so this definitely needs to be checked before merging.

### DIFF
--- a/packages/faucet/package.json
+++ b/packages/faucet/package.json
@@ -2,7 +2,7 @@
   "name": "@cosmjs/faucet",
   "version": "0.22.2",
   "description": "The faucet",
-  "contributors":[
+  "contributors": [
     "Ethan Frey <ethanfrey@users.noreply.github.com>",
     "Simon Warta <webmaster128@users.noreply.github.com>"
   ],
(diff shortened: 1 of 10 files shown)


packages/faucet/package.json
```diff
@@ -36,8 +36,8 @@
...
+    "start-dev": "FAUCET_CREDIT_AMOUNT_UCOSM=10000000 FAUCET_CREDIT_AMOUNT_USTAKE=5000000 FAUCET_CONCURRENCY=3 FAUCET_MNEMONIC=\"economy stock theory fatal elder harbor betray wasp final emotion task crumble siren bottom lizard educate guess current outdoor pair theory focus wife stone\" ./bin/cosmos-faucet start \"http://localhost:1317\"",
+    "start-coralnet": "FAUCET_ADDRESS_PREFIX=coral FAUCET_TOKENS=\"SHELL=10^6ushell, REEF=10^6ureef\" FAUCET_CREDIT_AMOUNT_USHELL=10000000 FAUCET_CREDIT_AMOUNT_UREEF=2000000 FAUCET_CONCURRENCY=3 FAUCET_MNEMONIC=\"economy stock theory fatal elder harbor betray wasp final emotion task crumble siren bottom lizard educate guess current outdoor pair theory focus wife stone\" ./bin/cosmos-faucet start \"https://lcd.coralnet.cosmwasm.com\""
```
webmaster128, line 40: This looks good. And it is only a demo command not used in production.

webmaster128 (approved):

PR #410, opened 2020-08-24
Add some provable queries to IBC querier

ethanfrey:
All unverified queries than can be read from a single key are now read from it.

### DIFF
--- a/packages/stargate/src/queries/ibc.ts
+++ b/packages/stargate/src/queries/ibc.ts
@@ -1,4 +1,5 @@
 import { toAscii } from "@cosmjs/encoding";
+import { Uint64 } from "@cosmjs/math";
 import Long from "long";
 
 import { ibc } from "../codec";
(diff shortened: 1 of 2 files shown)


packages/stargate/src/queries/ibc.ts
```diff
@@ -74,9 +90,39 @@ export function setupIbcExtension(base: QueryClient): IbcExtension {
...
+        const responseData = await base.queryVerified("ibc", key);
+        // keeper code doesn't parse, but returns raw
+        return responseData;
+      },
+      nextSequenceReceive: async (portId: string, channelId: string) => {
+        // keeper: https://github.com/cosmos/cosmos-sdk/blob/3bafd8255a502e5a9cee07391cf8261538245dfd/x/ibc/04-channel/keeper/keeper.go#L92-L101
+        // key: https://github.com/cosmos/cosmos-sdk/blob/ef0a7344af345882729598bc2958a21143930a6b/x/ibc/24-host/keys.go#L133-L136
+        const key = toAscii(`seqAcks/ports/${portId}/channels/${channelId}/nextSequenceAck`);
+        const responseData = await base.queryVerified("ibc", key);
+        return responseData.length ? parseUint64BigEndian(responseData) : null;
```
webmaster128, line 121: `Uint64.fromBytesBigEndian(responseData).toNumber()` should do that for you

ethanfrey: Thank you. Wish I found that in my brief search earlier. It would have saved me about 20 minutes
(I did learn about DataViews and TypedArrays however)

webmaster128: Yeah, encoding was split into multiple packages. All integers are in @cosmjs/math now.

PR #417, opened 2020-09-08
Add LedgerWallet

### LINKED ISSUE #268: Call into Cosmos Ledger app via OfflineSigner interface
Depends on #264 

Many apps want to sign with the cosmos ledger app. @faboweb has writen the excellent https://github.com/luniehq/cosmos-ledger library to integrate with the ledger from js clients.

Let's integrate this behind the wallet interface (and modify the interface if needed) to give more, secure options for key management to users of dApps built on CosmJS

willclarktech:
Closes #268

So far I just hosted the demo with `python3 -m http.server` in the `launchpad` directory.

Opening as a draft because of these outstanding questions:
- Should the demo code be tidied away more than it is? (E.g. subdirectory, README, separate webpack config if we add tests etc.)
- Should the demo include user-friendlier ways to construct the message?
- Should the demo include a `Msg` instead of an empty array? (I just used the simplest thing that worked.)
- Should the demo broadcast to a blockchain to demonstrate the full process?
- Should we have automated tests for this? (A big task because requesting access to USB devices requires a user interaction (e.g. clicking on a button.)

Note that the build is failing because the `typings` field of `@lunie/cosmos-ledger`'s `package.json` incorrectly points to `lib/types/cosmos-ledger.d.ts` instead of `lib/cosmos-ledger.d.ts`. (See https://github.com/luniehq/cosmos-ledger/pull/133)


packages/launchpad-ledger/src/launchpadledger.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  async sign(message: Uint8Array): Promise<Uint8Array> {
+    await this.connect();
+    assert(this.cosmosApp, "Cosmos Ledger App is not connected");
+
+    const response = await this.cosmosApp.sign(this.hdPath, fromUtf8(message));
+    this.handleLedgerErrors(response, {
+      rejectionMessage: "Transaction signing request was rejected by the user",
+    });
+    const { r: rRaw, s: sRaw } = bip66.decode((response as SignResponse).signature);
```
webmaster128, line 155: Could you try `Secp256k1Signature.fromDer((response as SignResponse).signature).toFixedLength()` with

```ts
export class Secp256k1Signature {
  // ...

  public toFixedLength(): Uint8Array {
    return new Uint8Array([...this.r(32), ...this.s(32)]);
  }
}
```

? Then we can remove the bip66 dependency.

packages/launchpad/package.json
```diff
@@ -45,6 +45,7 @@
   "dependencies": {
     "@cosmjs/crypto": "^0.22.2",
     "@cosmjs/encoding": "^0.22.2",
+    "@cosmjs/launchpad-ledger": "^0.22.2",
```
webmaster128, line 48: I'd prefer to make Ledger support optional, as it comes with dependencies that can be annoying sometimes (compiling libusb C code for Node.js). Can we swap the dependencies and more the demo to @cosmjs/launchpad-ledger?

webmaster128 (commented):
🐎

packages/launchpad-ledger/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "test-node": "node jasmine-testrunner.js",
+    "test-firefox": "yarn pack-web && karma start --single-run --browsers Firefox",
+    "test-chrome": "yarn pack-web && karma start --single-run --browsers ChromeHeadless",
+    "test": "yarn build-or-skip && yarn test-node",
+    "coverage": "nyc --reporter=text --reporter=lcov yarn test --quiet",
+    "pack-web": "yarn build-or-skip && webpack --mode development --config webpack.web.config.js"
+  },
+  "dependencies": {
+    "@cosmjs/utils": "^0.22.2",
+    "@ledgerhq/hw-transport-webhid": "^5.23.0",
```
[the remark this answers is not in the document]
willclarktech: So far! It's used by Lunie for Windows support. Do we need that?

packages/launchpad-ledger/src/launchpadledger.ts
```diff
@@ -0,0 +1,232 @@
...
+}
+
+function verifyBrowserIsSupported(platform: string, userAgent: string): void {
+  if (isWindows(platform)) {
+    throw new Error("Windows is not currently supported.");
+  }
+
+  const ua = userAgent.toLowerCase();
+  const isChrome = /chrome|crios/.test(ua) && !/edge|opr\//.test(ua);
+  if (!isChrome) {
```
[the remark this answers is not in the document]
willclarktech: Yes, but I can name this better. See https://github.com/luniehq/cosmos-ledger/blob/develop/src/cosmos-ledger.ts#L300-L301

packages/launchpad-ledger/src/launchpadledger.ts
willclarktech: Brilliant, I didn't find that method.

packages/launchpad-ledger/package.json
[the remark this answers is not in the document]
webmaster128: Windows support is blocked by

```ts
  if (isWindows(platform)) {
    throw new Error("Windows is not currently supported.");
  }
```

I'd do

1. Support WebUSB only
2. Figure out if we can use WebUSB with Windows
3. Add Node.js support as a 2nd transport
4. Add additional transports on demand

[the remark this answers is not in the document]
willclarktech: Lunie has code for Windows support but they use webhid. I just put that error in to simplify things and because I have no way to test it.

https://github.com/luniehq/cosmos-ledger/blob/develop/src/cosmos-ledger.ts#L85-L96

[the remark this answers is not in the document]
webmaster128: I see, thanks. However, it is well possible that the lack of Windows support for WebUSB is outdated.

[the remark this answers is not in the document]
willclarktech: https://github.com/LedgerHQ/ledgerjs/tree/master/packages/hw-transport-webusb#support-status

> Support status
WebUSB is currently only supported on Google Chrome / Chromium.
In Linux, user need to install the specific udev rules
In Mac, it should work.
In Windows, WebUSB does not work out of the box but you can fix it with Zadig.
In Android Chrome it works.

packages/launchpad-ledger/README.md
```diff
@@ -0,0 +1,12 @@
+# @cosmjs/launchpad-ledger
+
+[![npm version](https://img.shields.io/npm/v/@cosmjs/launchpad-ledger.svg)](https://www.npmjs.com/package/@cosmjs/launchpad-ledger)
+
+```
```
webmaster128, line 5: Thoe "```" look broken

packages/launchpad-ledger/src/ledgerwallet.ts
```diff
@@ -0,0 +1,45 @@
+import { AccountData, encodeSecp256k1Signature, OfflineSigner, StdSignature } from "@cosmjs/launchpad";
+
+import { LaunchpadLedger } from "./launchpadledger";
+
+interface LedgerWalletOptions {
+  readonly testModeAllowed: boolean;
+}
+
+export class LedgerWallet implements OfflineSigner {
```
webmaster128, line 9: What do you think about renaming this to `LedgerSigner`?

packages/launchpad-ledger/src/ledgerwallet.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  constructor(options?: LedgerWalletOptions) {
+    this.ledger = new LaunchpadLedger(options);
+  }
+
+  public async getAccounts(): Promise<readonly AccountData[]> {
+    await this.ledger.connect();
+
+    const address = (this.address = this.address || (await this.ledger.getCosmosAddress()));
+    const pubkey = (this.pubkey = this.pubkey || (await this.ledger.getPubKey()));
```
webmaster128, line 22: So here we could potentially return multiple accounts but there is no state available telling us which account exists, right?

packages/launchpad-ledger/src/ledgerwallet.ts
```diff
@@ -0,0 +1,45 @@
+import { AccountData, encodeSecp256k1Signature, OfflineSigner, StdSignature } from "@cosmjs/launchpad";
+
+import { LaunchpadLedger } from "./launchpadledger";
+
+interface LedgerWalletOptions {
+  readonly testModeAllowed: boolean;
```
webmaster128, line 6: I think we should support different paths here

webmaster128:
> Should the demo code be tidied away more than it is? (E.g. subdirectory, README, separate webpack config if we add tests etc.)

I think it would help to have the demo files in a `src/demo` but using .demo.x probably works as well. Instructions how to run the demo in the package's README.md would help.

> Should the demo include user-friendlier ways to construct the message?

I don't think this is necessary

> Should the demo include a Msg instead of an empty array? (I just used the simplest thing that worked.)

Yes, any simple send would do the job

> Should the demo broadcast to a blockchain to demonstrate the full process?

Mixed feeling. Broadcasting is the primary way of telling whether or not the signature is correct. But on the other hand, every Ledger tester must use a pre-defined Ledger mnemonic such that the account can be created.

> Should we have automated tests for this? (A big task because requesting access to USB devices requires a user interaction (e.g. clicking on a button.)

No

packages/launchpad-ledger/src/ledgerwallet.ts
willclarktech: What do you mean by "exists"? Eg exists where?

webmaster128: "exists" as in _has be created by the user before_.

A Ledger app can give us any address `m/44'/118'/0'/0/a` and from an `OfflineSigner.getAccounts` implementation I would expect to get all used accounts. However, I assume this is not possible without online [account discovery](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#Account_discovery). So we probably have to live with a single-account implementation.

What we could do is to accept a list of paths in the constructor instead of a single one.

packages/launchpad-ledger/src/launchpadledger.ts
```diff
@@ -0,0 +1,232 @@
...
+        throw new Error(
+          `Your Cosmos Ledger App is not up to date. Please update to version ${requiredCosmosAppVersion}.`,
+        );
+      case "No errors":
+        break;
+      default:
+        throw new Error(`Ledger Native Error: ${errorMessage}`);
+    }
+  }
+  /* eslint-enable */
```
webmaster128, line 231: Can we reduce the scope of this eslint-disable by writing some of

```ts
  private handleLedgerErrors(
    {
      error_message: errorMessage,
      device_locked: deviceLocked = false,
    }: { error_message: string; device_locked?: boolean },
    { rejectionMessage = "Request was rejected by the user" } = {},
  ): void {
```

as interfaces?

webmaster128 (approved):
Nice. One more comment and a few CHANGELOG entries missing. Other than that, good to merge

packages/launchpad-ledger/src/launchpadledger.ts
willclarktech: Doesn't seem so 😞 
https://github.com/typescript-eslint/typescript-eslint/issues/2250#issuecomment-670055096

PR #421, opened 2020-09-15
Add Decimal.minus() to math package

abefernan:
Add subtraction support to `Decimal` class so that allowances can be subtracted in the frontend code


webmaster128 (approved):

webmaster128:
Beautiful

PR #423, opened 2020-09-15
Add Node.js support to LedgerSigner

### LINKED ISSUE #422: Add Node.js support for @cosmjs/launchpad-ledger
We should have an alternative Transport to WebUSB that works in Node.js. Ideally the transport is selected automatically internally.

willclarktech:
Closes #422

I'm not sure what the best way to detect the environment is, but this is quite direct in terms of what the browser version needs and it seems to work fine.

Also note this very ugly hack to get webpack to bundle only `@ledgerhq/hw-transport-webusb` but not `@ledgerhq/hw-transport-node-hid`:
https://github.com/CosmWasm/cosmjs/blob/fcd659b204137ea52b3cfc62e7777357ad0a3503/packages/launchpad-ledger/src/launchpadledger.ts#L145-L151


packages/launchpad-ledger/demo/node.js
```diff
@@ -0,0 +1,14 @@
+const demo = require("../build/demo/node");
+
+async function run() {
+  const accounts = await demo.getAccounts();
+  console.info("Accounts from Ledger device:");
+  console.table(accounts);
```
webmaster128, line 6: Could this demo show a bunch of addresses (let's say account index 0, 1, 2) and the user selects which one to use?

packages/launchpad-ledger/src/demo/node.ts
```diff
@@ -0,0 +1,47 @@
...
+    readonly pubkey: string;
+  }>
+> {
+  const accounts = await signer.getAccounts();
+  return accounts.map((account) => ({ ...account, pubkey: toHex(account.pubkey) }));
+}
+
+export async function sign(fromAddress: string, toAddress: string): Promise<StdSignature> {
+  const rawMessage = createMessage(fromAddress, toAddress);
+  const message = JSON.stringify(JSON.parse(rawMessage));
```
webmaster128, line 45: Better use `makeSignBytes` from `@cosmjs/launchpad` to ensure the sorting of the JSON fields is correct

packages/launchpad-ledger/demo/node.js
willclarktech: Good we checked because Ledger doesn't like more than one request at a time (e.g. from `Promise.all`)!

packages/launchpad-ledger/src/ledgersigner.ts
```diff
@@ -1,18 +1,33 @@
...
+  encodeSecp256k1Signature,
+  makeCosmoshubPath,
+  OfflineSigner,
+  StdSignature,
+} from "@cosmjs/launchpad";
+
+import { LaunchpadLedger } from "./launchpadledger";
+
+export interface LedgerSignerOptions {
+  readonly accountNumbers?: readonly number[];
```
webmaster128, line 12: Instead of account numbers we should allow `HdPath[]` here (falling back to `[makeComsoshubPath(0)]`). Then the demo code can set a bunch of paths.

webmaster128 (commented):
Cool, just the same path flexibility for `LaunchpadLedger` and `LedgerSigner`

webmaster128 (approved):

PR #425, opened 2020-09-17
Remove ticker/fractional digits from faucet

### LINKED ISSUE #391: Give up display tokens in the faucet

webmaster128:
Closes #391

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,9 +23,9 @@
 - @cosmjs/faucet: Environmental variable `FAUCET_FEE` renamed to
   `FAUCET_GAS_PRICE` and now only accepts one token. Environmental variable
   `FAUCET_GAS` renamed to `FAUCET_GAS_LIMIT`.
-- @cosmjs/faucet: `/credit` API now accepts either `denom` (base token) or as
-  before `ticker` (unit token). Environmental variables specifying credit
-  amounts now need to use uppercase denom.
+- @cosmjs/faucet: `/credit` API now expects `denom` (base token) instead of
+  `ticker` (unit token). Environmental variables specifying credit amounts now
+  need to use uppercase denom.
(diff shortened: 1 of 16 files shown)


willclarktech (approved):
Do we need `BankTokenMeta` at all anymore?

webmaster128:
> Do we need `BankTokenMeta` at all anymore?

Nope, and we can remove a lot of code by removing the type (see the last 2 commits). Thanks.

willclarktech (approved):

PR #432, opened 2020-09-23
Allow OfflineSigner implementation to override fields

webmaster128:
This feature was requested by Keplr to allow the user to adjust things like gas price. This basically implements https://github.com/CosmWasm/cosmjs/pull/403#issuecomment-683477227. See `signer.ts` for the updated interface.

The reason why I chose to return `{ signedDoc: StdSignDoc, signature: StdSignature }` instead of `StdTx` is that it works better when collecting multiple signatures from multiple sign operations. In this case the `StdTx` returned from one signer is not really something you can submit as a signed transaction. Also returning `StdTx` does not allow the caller to detect changes in `account_number` and `sequence`.

~Based on #433~

- [x] Base implementation
- [x] Pull out Ledger demo commits
- [x] Create a helper that combines an StdSignDoc and a single StdSignature into a signed transaction

### DIFF
--- a/packages/launchpad-ledger/src/ledgersigner.ts
+++ b/packages/launchpad-ledger/src/ledgersigner.ts
@@ -48,7 +48,7 @@ export class LedgerSigner implements OfflineSigner {
     const hdPath = this.hdPaths[accountIndex];
     const signature = await this.ledger.sign(message, hdPath);
     return {
-      signedDoc: signDoc,
+      signed: signDoc,
       signature: encodeSecp256k1Signature(accountForAddress.pubkey, signature),
     };
   }
--- a/packages/launchpad-ledger/src/ledgersigner.ts
+++ b/packages/launchpad-ledger/src/ledgersigner.ts
@@ -4,10 +4,9 @@ import {
   encodeSecp256k1Signature,
   makeCosmoshubPath,
   OfflineSigner,
-  StdSignature,
   StdSignDoc,
 } from "@cosmjs/launchpad";
-import { serializeSignDoc } from "@cosmjs/launchpad";
+import { serializeSignDoc, SignResponse } from "@cosmjs/launchpad";
 
 import { LaunchpadLedger, LaunchpadLedgerOptions } from "./launchpadledger";
 
@@ -36,7 +35,7 @@ export class LedgerSigner implements OfflineSigner {
     return this.accounts;
   }
 
-  public async sign(signerAddress: string, signDoc: StdSignDoc): Promise<StdSignature> {
+  public async sign(signerAddress: string, signDoc: StdSignDoc): Promise<SignResponse> {
     const accounts = this.accounts || (await this.getAccounts());
     const accountIndex = accounts.findIndex((account) => account.address === signerAddress);
 
@@ -48,6 +47,9 @@ export class LedgerSigner implements OfflineSigner {
     const accountForAddress = accounts[accountIndex];
     const hdPath = this.hdPaths[accountIndex];
     const signature = await this.ledger.sign(message, hdPath);
-    return encodeSecp256k1Signature(accountForAddress.pubkey, signature);
+    return {
+      signedDoc: signDoc,
+      signature: encodeSecp256k1Signature(accountForAddress.pubkey, signature),
+    };
   }
 }
--- a/packages/launchpad-ledger/src/ledgersigner.ts
+++ b/packages/launchpad-ledger/src/ledgersigner.ts
@@ -5,7 +5,9 @@ import {
   makeCosmoshubPath,
   OfflineSigner,
   StdSignature,
+  StdSignDoc,
 } from "@cosmjs/launchpad";
+import { serializeSignDoc } from "@cosmjs/launchpad";
 
 import { LaunchpadLedger, LaunchpadLedgerOptions } from "./launchpadledger";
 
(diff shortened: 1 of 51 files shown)


packages/launchpad-ledger/src/ledgersigner.ts
```diff
@@ -34,17 +35,21 @@ export class LedgerSigner implements OfflineSigner {
...
+      throw new Error(`Address ${signerAddress} not found in wallet`);
     }
 
+    const message = serializeSignDoc(signDoc);
     const accountForAddress = accounts[accountIndex];
     const hdPath = this.hdPaths[accountIndex];
     const signature = await this.ledger.sign(message, hdPath);
-    return encodeSecp256k1Signature(accountForAddress.pubkey, signature);
+    return {
+      signedDoc: signDoc,
```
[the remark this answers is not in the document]
webmaster128: Whatr about just `signed: StdSignDoc`?

willclarktech:
CLI tests are failing. It's pretty annoying we can't use our normal tools to detect that in advance.

packages/launchpad-ledger/src/ledgersigner.ts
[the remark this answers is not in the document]
willclarktech: That works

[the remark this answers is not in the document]
webmaster128: Done in https://github.com/CosmWasm/cosmjs/pull/432/commits/602f68f667729a23a7eb0dfc5ec1f95c954d9c41

willclarktech (approved):

packages/launchpad/src/signer.ts
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * Request signature from whichever key corresponds to provided bech32-encoded address. Rejects if not enabled.
+   *
+   * The signer implementation may offer the user the ability to override parts of the signDoc. It must
+   * return the doc that was signed in the response.
+   *
+   * @param signerAddress The address of the account that should sign the transaction
+   * @param signDoc The content that should be signed
+   */
+  readonly sign: (signerAddress: string, signDoc: StdSignDoc) => Promise<SignResponse>;
```
webmaster128, line 37: @Thunnini can you have a look in this file? This allows the signer implementation to override any field in the SignDoc. The flow is:

1. Application sends a `StdSignDoc` to be signed
2. The signer shows it tot he and (optionally) allows the user to modify it
3. User approves
4. Signer serializes `StdSignDoc` (e.g. using `serializeSignDoc`) and creates a signature
5. Signer returns the signature and the final document it signed at the end of the day
6. Application can now double check what was signed (or just trust the signer) and broadcast

This should cover most of your requested flexibility, just in one call instead of multiple calls.

packages/launchpad/src/signer.ts
```diff
@@ -0,0 +1,38 @@
...
+export type Algo = "secp256k1" | "ed25519" | "sr25519";
+
+export interface AccountData {
+  /** A printable address (typically bech32 encoded) */
+  readonly address: string;
+  readonly algo: Algo;
+  readonly pubkey: Uint8Array;
+}
+
+export interface SignResponse {
```
ethanfrey, line 13: Nice response and good docstring

packages/launchpad/src/signingcosmosclient.ts
```diff
@@ -88,8 +88,8 @@ export class SigningCosmosClient extends CosmosClient {
   public async signAndBroadcast(msgs: readonly Msg[], fee: StdFee, memo = ""): Promise<BroadcastTxResult> {
     const { accountNumber, sequence } = await this.getSequence();
     const chainId = await this.getChainId();
-    const signBytes = makeSignBytes(msgs, fee, chainId, memo, accountNumber, sequence);
-    const signature = await this.signer.sign(this.senderAddress, signBytes);
+    const signDoc = makeSignDoc(msgs, fee, chainId, memo, accountNumber, sequence);
```
ethanfrey, line 91: If I understood this correctly above, the `sign` method can now modify the SignDoc and returns the updated one. In particular, it could adjust fee and memo to be different than what was passed in.

Right below this, when building `signedTx`, we should use the returned (signed) fee and memo. 

This should fail if we test it with an OfflineSigner implementation that eg. doubles the fees and sets memo to "Hello World" before signing. It would be good to test that case and ensure that all code handles updated signdocs results properly.

ethanfrey: I am fine with only passing the original messages in here, as if the wallet changed those when signing, we probably want this to fail. But `fee` and `memo` should definitely be under the control of the OfflineSigner

ethanfrey (commented):
Very nice interface, I like it. Simple, clean, and powerful. :rocket: 
Thank you for taking this over, much nicer than my approach.

I think the implementation is not complete and doesn't properly handle the case the doc is modified by the signer. Left a comment at one location where that is demo'd - the same is true for SigningCosmWasmClient

webmaster128:
> I think the implementation is not complete and doesn't properly handle the case the doc is modified by the signer.

Well spotted. This was a test ;)

webmaster128:
@willclarktech can you have another look at the commits starting with "Remove makeSignBytes"

@ethanfrey https://github.com/CosmWasm/cosmjs/pull/432/commits/9d1508e1eeeff05f13f6256ba58717b565ff7134 should address your main point I guess

webmaster128:
Merging now in order to pre-release 0.23.0. Any further comments are still welcome.

PR #433, opened 2020-09-23
Create DirectSecp256k1Wallet

webmaster128:
This creates a minimal wallet (copy of Secp256k1Wallet without serialization) for the use in protobuf tests. This makes protobuf code independent of the OfflineSigner interface.

This PR should be merged before #432

### DIFF
--- a/packages/proto-signing/src/directsecp256k1wallet.spec.ts
+++ b/packages/proto-signing/src/directsecp256k1wallet.spec.ts
@@ -1,0 +1,70 @@
+import { Secp256k1, Secp256k1Signature, Sha256 } from "@cosmjs/crypto";
+import { fromBase64, fromHex, toAscii } from "@cosmjs/encoding";
+
+import { DirectSecp256k1Wallet } from "./directsecp256k1wallet";
+
+describe("DirectSecp256k1Wallet", () => {
+  // m/44'/118'/0'/0/0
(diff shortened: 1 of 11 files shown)


packages/proto-signing/src/directsecp256k1wallet.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+      expect((await DirectSecp256k1Wallet.generate(12)).mnemonic.split(" ").length).toEqual(12);
+      expect((await DirectSecp256k1Wallet.generate(15)).mnemonic.split(" ").length).toEqual(15);
+      expect((await DirectSecp256k1Wallet.generate(18)).mnemonic.split(" ").length).toEqual(18);
+      expect((await DirectSecp256k1Wallet.generate(21)).mnemonic.split(" ").length).toEqual(21);
+      expect((await DirectSecp256k1Wallet.generate(24)).mnemonic.split(" ").length).toEqual(24);
+    });
+  });
+
+  describe("getAccounts", () => {
+    it("resolves to a list of accounts if enabled", async () => {
```
willclarktech, line 37: "if enabled"?

willclarktech (approved):

packages/proto-signing/src/directsecp256k1wallet.spec.ts
webmaster128: Oh, this neither makes sense here nor in Secp256k1Wallet since both wallets cannot be enabled/disabled. Will remove.

PR #436, opened 2020-09-24
Add sortJson tests

webmaster128:


packages/launchpad/src/encoding.spec.ts
```diff
@@ -1 +1,55 @@
...
+    });
+
+    it("sorts objects by key", () => {
+      // already sorted
+      expect(sortJson({})).toEqual({});
+      expect(sortJson({ a: 3 })).toEqual({ a: 3 });
+      expect(sortJson({ a: 3, b: 2, c: 1 })).toEqual({ a: 3, b: 2, c: 1 });
+
+      // not yet sorted
+      expect(sortJson({ b: 2, a: 3, c: 1 })).toEqual({ a: 3, b: 2, c: 1 });
```
willclarktech, line 22: `toEqual` doesn't test the order of keys. Is that what you wanted?

webmaster128: Arg, good point. We probably have to stringify both sides just to get ordered object equalty checks.

willclarktech: Probably the best way. We could also cycle through `Object.entries` recursively...?

webmaster128: Okay I think those test revealed that an object sorting API is not the most clever thing to have. I changed the interface a bit to combine sording and serialization which is what we really want: sorted JSON serialization.

packages/launchpad/src/encoding.ts
```diff
@@ -5,24 +5,30 @@ import { Uint53 } from "@cosmjs/math";
 import { Msg } from "./msgs";
 import { StdFee } from "./types";
 
-function sortJson(json: any): any {
-  if (typeof json !== "object" || json === null) {
-    return json;
+function sortedObject(obj: any): any {
```
willclarktech, line 8: Suggest naming this `sortObject`.

packages/launchpad/src/encoding.spec.ts
```diff
@@ -1 +1,50 @@
...
+      expect(sortedJsonStringify(true)).toEqual(`true`);
+      expect(sortedJsonStringify(false)).toEqual(`false`);
+      expect(sortedJsonStringify("aabbccdd")).toEqual(`"aabbccdd"`);
+      expect(sortedJsonStringify(75)).toEqual(`75`);
+      expect(sortedJsonStringify(null)).toEqual(`null`);
+      expect(sortedJsonStringify([5, 6, 7, 1])).toEqual(`[5,6,7,1]`);
+      expect(sortedJsonStringify([5, ["a", "b"], true, null, 1])).toEqual(`[5,["a","b"],true,null,1]`);
+    });
+
+    it("sorts objects by key", () => {
```
willclarktech, line 15: I'm wondering if we need to worry about ordering of eg special characters, capital letters etc?

webmaster128: So far we never came across such a case for transaction signing. Since all of this is not specified but defined by _do what Go does (what we don't know exactly but let's see)_, I'd avoid adding more logic or tests here.

packages/launchpad/src/encoding.ts
webmaster128: This follows the idea that "sort" is a mutating operation while "sorted" returns a sorted copy of the object. This way it is done [in Python](https://discuss.codecademy.com/t/what-is-the-difference-between-sort-and-sorted/349679) and [in Swift](https://stackoverflow.com/questions/40621076/whats-the-difference-between-sorted-and-sort-in-array) and I liked the concept. In JavaScript "sort" is a mutating operation as well, which also returns the sorted object for no clear reason.

willclarktech: Fine with me. For the record I quite like the approach of using the same name with (non-mutating) or without (mutating) a trailing underscore which kind of hints that mutation is weird and bad.
Eg https://pytorch.org/docs/stable/tensors.html#torch.Tensor.abs

willclarktech (approved):

packages/launchpad/src/encoding.ts
webmaster128: :) better than the "i" prefix of bn.js for sure

PR #437, opened 2020-09-29
Enable explicit-member-accessibility and no-parameter-properties

webmaster128:
Two linter rules we used to have in the IOV Core days


willclarktech (approved):
💇

PR #442, opened 2020-09-30
Add Uint32.fromBytes/Uint64.fromBytes/Uint32.fromString

### LINKED ISSUE #411: Unify naming of Uint64.fromBytesBigEndian/Uint32.fromBigEndianBytes
It would be good to have a static interface, ensuring the naming is consistent.

Since we almost always need big endian, we can simply use

```ts
public static fromBytes(bytes: ArrayLike<number>, endianess: "be" | "le" = "be"): Uint32 {
```

leading to much shorter caller code on average.

webmaster128:
Closes #411

### DIFF
--- a/packages/math/src/integers.spec.ts
+++ b/packages/math/src/integers.spec.ts
@@ -80,52 +80,52 @@ describe("Integers", () => {
       });
     });
 
-    describe("fromBigEndianBytes", () => {
+    describe("fromBytes", () => {
       it("can be constructed from to byte array", () => {
-        expect(Uint32.fromBigEndianBytes([0, 0, 0, 0]).toNumber()).toEqual(0);
-        expect(Uint32.fromBigEndianBytes([0, 0, 0, 1]).toNumber()).toEqual(1);
-        expect(Uint32.fromBigEndianBytes([0, 0, 0, 42]).toNumber()).toEqual(42);
-        expect(Uint32.fromBigEndianBytes([0x3b, 0x9a, 0xca, 0x00]).toNumber()).toEqual(1000000000);
-        expect(Uint32.fromBigEndianBytes([0x7f, 0xff, 0xff, 0xff]).toNumber()).toEqual(2147483647);
-        expect(Uint32.fromBigEndianBytes([0x80, 0x00, 0x00, 0x00]).toNumber()).toEqual(2147483648);
-        expect(Uint32.fromBigEndianBytes([0xff, 0xff, 0xff, 0xff]).toNumber()).toEqual(4294967295);
+        expect(Uint32.fromBytes([0, 0, 0, 0]).toNumber()).toEqual(0);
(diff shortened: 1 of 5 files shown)


packages/math/src/integers.ts
```diff
@@ -80,6 +105,8 @@ export class Uint32 implements Integer, WithByteConverters {
   }
 }
 
+const _uint32ClassConformsToStaticInterface: FixedLengthIntegerStatic<Uint32> = Uint32;
```
willclarktech, line 108: Are these just to check the types? Maybe they make more sense in a test file?

packages/math/src/integers.spec.ts
```diff
@@ -396,77 +431,77 @@ describe("Integers", () => {
...
-      expect(
-        Uint64.fromBytesBigEndian([0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]).toBytesBigEndian(),
-      ).toEqual(new Uint8Array([0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]));
-      expect(
-        Uint64.fromBytesBigEndian([0xab, 0x22, 0xbc, 0x5f, 0xa9, 0x20, 0x4e, 0x0d]).toBytesBigEndian(),
-      ).toEqual(new Uint8Array([0xab, 0x22, 0xbc, 0x5f, 0xa9, 0x20, 0x4e, 0x0d]));
-      expect(
-        Uint64.fromBytesBigEndian([0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff]).toBytesBigEndian(),
-      ).toEqual(new Uint8Array([0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff]));
+      expect(Uint64.fromBytes([0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]).toBytesBigEndian()).toEqual(
```
willclarktech, line 434: Did you consider merging these methods too?

packages/math/src/integers.ts
webmaster128: Yes, in eb63957b08d16ddb2209d08407d464e27beda670 I clarified that a bit. It is the static version of the `implements` keyword. It could also be written as

I think this check belongs to the definition. If we move this to a test we need to make the interfaces public, which I would like to avoid if possible.

packages/math/src/integers.spec.ts
webmaster128: This could be done, yes. But those are not as important from my point of view because we have consistent naming due to the instance interface

```ts
interface WithByteConverters {
  readonly toBytesBigEndian: () => Uint8Array;
  readonly toBytesLittleEndian: () => Uint8Array;
}
```

willclarktech (approved):

PR #444, opened 2020-09-30
Fix timeout calculation and document why it does not matter

webmaster128:
Before this, the timeout was calucalted as 120 * 1000 * 1000. But at the end of the day, the value is unused.


willclarktech (approved):
🙃

PR #446, opened 2020-10-01
Pass transport to Ledger signer

### LINKED ISSUE #445: Let @cosmjs/launchpad-ledger take Transport as an argument
- [x] Do the same change as it was done here: https://github.com/iov-one/iov-ledger-iovns/pull/31:
       _Give up the idea of managing Transports internally. The transport handling is too environment-specific and should be done by the caller._
- [x] Remove `optionalDependencies` and only use `@ledgerhq/hw-transport-node-hid` and `@ledgerhq/hw-transport-webusb` in demos and tests

willclarktech:
Closes #445

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,9 @@ jobs:
           keys:
             - yarn-packages-{{ checksum "yarn.lock" }}
       - run:
+          name: Install libusb
+          command: sudo apt-get install libusb-1.0-0-dev
+      - run:
           name: Install Dependencies
           command: yarn install --frozen-lockfile
       - save_cache:
@@ -111,6 +114,11 @@ jobs:
           name: Restore Yarn Package Cache
           keys:
             - yarn-packages-{{ checksum "yarn.lock" }}
+      - run:
+          name: Install libusb
+          command: |
+            wget -O "$HOME/libusb.deb" http://de.archive.ubuntu.com/ubuntu/pool/main/libu/libusb-1.0/libusb-1.0-0-dev_1.0.20-1_amd64.deb
+            sudo dpkg -i "$HOME/libusb.deb"
       - run:
           name: Install Dependencies
           command: yarn install --frozen-lockfile
@@ -215,6 +223,11 @@ jobs:
           keys:
             - yarn-packages-{{ checksum "yarn.lock" }}
       - run:
+          name: Install libusb
+          command: |
+            wget -O "$HOME/libusb.deb" http://de.archive.ubuntu.com/ubuntu/pool/main/libu/libusb-1.0/libusb-1.0-0-dev_1.0.20-1_amd64.deb
+            sudo dpkg -i "$HOME/libusb.deb"
+      - run:
           name: Install Dependencies
           command: yarn install --frozen-lockfile
       - save_cache:
@@ -299,6 +312,11 @@ jobs:
           keys:
             - yarn-packages-{{ checksum "yarn.lock" }}
       - run:
+          name: Install libusb
+          command: |
+            wget -O "$HOME/libusb.deb" http://de.archive.ubuntu.com/ubuntu/pool/main/libu/libusb-1.0/libusb-1.0-0-dev_1.0.20-1_amd64.deb
+            sudo dpkg -i "$HOME/libusb.deb"
+      - run:
           name: Install Dependencies
           command: yarn install --frozen-lockfile
       - save_cache:
@@ -344,6 +362,9 @@ jobs:
           name: Restore Yarn Package Cache
           keys:
             - yarn-packages-{{ checksum "yarn.lock" }}
+      - run:
+          name: Install libusb
+          command: sudo apt-get install libusb-1.0-0-dev
       - run:
           name: Install Dependencies
           command: yarn install --frozen-lockfile
@@ -391,6 +412,9 @@ jobs:
           name: Restore Yarn Package Cache
           keys:
             - yarn-packages-{{ checksum "yarn.lock" }}
+      - run:
+          name: Install libusb
+          command: sudo apt-get install libusb-1.0-0-dev
       - run:
           name: Install Dependencies
           command: yarn install --frozen-lockfile
--- a/packages/launchpad-ledger/src/demo/node.ts
+++ b/packages/launchpad-ledger/src/demo/node.ts
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import { toBase64 } from "@cosmjs/encoding";
 import { makeCosmoshubPath, makeSignDoc, StdFee, StdSignature } from "@cosmjs/launchpad";
-import LedgerTransport from "@ledgerhq/hw-transport-node-hid";
+import TransportNodeHid from "@ledgerhq/hw-transport-node-hid";
 
 import { LedgerSigner } from "../ledgersigner";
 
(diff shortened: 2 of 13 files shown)


packages/launchpad-ledger/src/demo/node.ts
```diff
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import { toBase64 } from "@cosmjs/encoding";
 import { makeCosmoshubPath, makeSignDoc, StdFee, StdSignature } from "@cosmjs/launchpad";
+import LedgerTransport from "@ledgerhq/hw-transport-node-hid";
```
webmaster128, line 4: What about using the orgininal transport name form https://github.com/LedgerHQ/ledgerjs/tree/master/packages/hw-transport-node-hid/src?

packages/launchpad-ledger/src/demo/web.ts
```diff
@@ -2,6 +2,7 @@ import { toBase64 } from "@cosmjs/encoding";
 import { AccountData, makeCosmoshubPath, StdSignDoc } from "@cosmjs/launchpad";
 import { Uint53 } from "@cosmjs/math";
 import { assert } from "@cosmjs/utils";
+import LedgerTransport from "@ledgerhq/hw-transport-webusb";
```
webmaster128, line 5: same here

packages/launchpad-ledger/src/launchpadledger.ts
```diff
@@ -2,7 +2,7 @@ import { HdPath, Secp256k1Signature } from "@cosmjs/crypto";
 import { fromUtf8 } from "@cosmjs/encoding";
 import { makeCosmoshubPath } from "@cosmjs/launchpad";
 import { assert } from "@cosmjs/utils";
-import Transport from "@ledgerhq/hw-transport";
+import LedgerTransport from "@ledgerhq/hw-transport";
```
webmaster128, line 5: Why rename this interface?

packages/launchpad-ledger/src/launchpadledger.ts
```diff
@@ -125,82 +93,20 @@ export class LaunchpadLedger {
...
-      if (trimmedErrorMessage.startsWith("No device selected")) {
-        throw new Error(
-          "You did not select a Ledger device. If you did not see your Ledger, check if the Ledger is plugged in and unlocked.",
-        );
-      }
-
-      throw error;
+    if (this.app) {
+      await this.transport.close();
+      this.app = null;
```
webmaster128, line 109: When the transport is provided externally, do we even want to keep this methis? We could close the transport in the caller code for symmetry.

webmaster128 (commented):
Jap, much better. We now should remove the optionalDependencies and adjust the README accordingly.

packages/launchpad-ledger/src/launchpadledger.ts
willclarktech: `Transport` is already available via the DOM lib: https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L20083

webmaster128: And we cannot override it? If I recall correctly, importing types overrides globals

willclarktech: We can. Do we... want to?

willclarktech:
@webmaster128 These ledgerhq modules don't seem to install on CI. They were already failing, but because they were optional before the failure was OK: https://app.circleci.com/pipelines/github/CosmWasm/cosmjs/1369/workflows/ea95d013-ce87-41f9-a437-ae58bcecc77f/jobs/5558

Can we just skip the installation for this package since we won't be testing on CI?

webmaster128:
> Can we just skip the installation for this package since we won't be testing on CI?

I like the idea. Do you have an idea how to skip the installation of a `devDependency`?

packages/launchpad-ledger/src/launchpadledger.ts
webmaster128: I think we should. This is only a type and not a global object. `Transport` is undefined in JavaScript (node.js, Firefox and Chrome)

willclarktech:
@webmaster128 I thought I did: https://github.com/CosmWasm/cosmjs/pull/446/commits/27e0273390956e4362b64c7d6c48214f99f959fd
But it fails both locally and on CI (but for different reasons).

.circleci/config.yml
```diff
@@ -41,6 +41,9 @@ jobs:
           name: Restore Yarn Package Cache
           keys:
             - yarn-packages-{{ checksum "yarn.lock" }}
+      - run:
+        name: Install libusb
+        command: sudo apt-get install libusb
```
webmaster128, line 46: you probably need someting like libusb-dev to compile against libusb headers

packages/launchpad-ledger/src/ledgersigner.spec.ts
```diff
@@ -12,12 +12,29 @@ import {
   StdFee,
 } from "@cosmjs/launchpad";
 import { assert, sleep } from "@cosmjs/utils";
+import LedgerTransport from "@ledgerhq/hw-transport";
```
webmaster128, line 15:
```suggestion
import Transport from "@ledgerhq/hw-transport";
```

packages/launchpad-ledger/src/ledgersigner.spec.ts
```diff
@@ -12,12 +12,29 @@ import {
...
 
 import { LedgerSigner } from "./ledgersigner";
 import { pendingWithoutLedger, pendingWithoutWasmd, wasmd } from "./testutils.spec";
 
 const interactiveTimeout = 120_000;
 
+async function createLedgerTransport(): Promise<LedgerTransport> {
```
webmaster128, line 22:
```suggestion
async function createLedgerTransport(): Promise<Transport> {
```

packages/launchpad-ledger/src/ledgersigner.spec.ts
```diff
@@ -28,11 +45,20 @@ describe("LedgerSigner", () => {
   const defaultSequence = "0";
   const defaultAccountNumber = "42";
   const defaultRecipient = "cosmos1p6xs63q4g7np99ttv5nd3yzkt8n4qxa47w8aea";
+  let ledgerTransport: LedgerTransport;
```
webmaster128, line 48:
```suggestion
  let ledgerTransport: Transport;
```

packages/launchpad-ledger/src/ledgersigner.ts
```diff
@@ -7,6 +7,7 @@ import {
   StdSignDoc,
 } from "@cosmjs/launchpad";
 import { serializeSignDoc, SignResponse } from "@cosmjs/launchpad";
+import LedgerTransport from "@ledgerhq/hw-transport";
```
webmaster128, line 10:
```suggestion
import Transport from "@ledgerhq/hw-transport";
```

packages/launchpad-ledger/src/ledgersigner.ts
```diff
@@ -15,9 +16,9 @@ export class LedgerSigner implements OfflineSigner {
   private readonly hdPaths: readonly HdPath[];
   private accounts?: readonly AccountData[];
 
-  public constructor(options: LaunchpadLedgerOptions = {}) {
+  public constructor(transport: LedgerTransport, options: LaunchpadLedgerOptions = {}) {
```
webmaster128, line 19:
```suggestion
  public constructor(transport: Transport, options: LaunchpadLedgerOptions = {}) {
```

webmaster128 (approved):

PR #452, opened 2020-10-07
Add Secpk256k1 offline key signer

shanev:
An implementation of `OfflineSigner` that allows signing with a private key instead of a mnemonic.

### DIFF
--- a/packages/cli/examples/delegate.ts
+++ b/packages/cli/examples/delegate.ts
@@ -1,4 +1,4 @@
-const wallet = await Secp256k1HdWallet.fromMnemonic(
+const wallet = await Secp256k1Wallet.fromMnemonic(
   "enlist hip relief stomach skate base shallow young switch frequent cry park",
 );
 const [{ address: senderAddress }] = await wallet.getAccounts();
--- a/packages/cli/examples/delegate.ts
+++ b/packages/cli/examples/delegate.ts
@@ -1,4 +1,4 @@
-const wallet = await Secp256k1Wallet.fromMnemonic(
+const wallet = await Secp256k1HdWallet.fromMnemonic(
   "enlist hip relief stomach skate base shallow young switch frequent cry park",
 );
 const [{ address: senderAddress }] = await wallet.getAccounts();
--- a/packages/cli/examples/faucet_addresses.ts
+++ b/packages/cli/examples/faucet_addresses.ts
@@ -2,7 +2,7 @@ const mnemonic =
   "economy stock theory fatal elder harbor betray wasp final emotion task crumble siren bottom lizard educate guess current outdoor pair theory focus wife stone";
 
 for (let i of [0, 1, 2, 3, 4]) {
-  const wallet = await Secp256k1HdWallet.fromMnemonic(mnemonic, makeCosmoshubPath(i), "cosmos");
+  const wallet = await Secp256k1Wallet.fromMnemonic(mnemonic, makeCosmoshubPath(i), "cosmos");
   const [{ address, pubkey }] = await wallet.getAccounts();
   console.info(`Address ${i}: ${address}`);
   console.info(`Pubkey ${i}: ${toBase64(pubkey)}`);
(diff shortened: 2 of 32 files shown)


webmaster128 (commented):
This looks solid, thanks.

I wonder how to integrate this. The natural name for this signer would be `Secp256k1Wallet`, which we already use. I think the term "wallet" makes sense for a signer that stores keys. So I tend to like the renaming
- `Secp256k1Wallet` -> `Secp256k1HdWallet`
- `Secp256k1Key` -> `Secp256k1Wallet`
This makes the most common wallet 2 characters longer and break apps but would be clear.

I am curious to see where you use this signer implementation.

On the code side: could you run `yarn lint-fix` in the package (or repo root) to get the formatting right? In general we use "privkey" and "pubkey" as atomic words in this project. I know they aren't in the dictionary but treating them as words makes code very ergonomic. But I'm not sure what the best name for `Secp256k1Key.fromPrivateKey` is. Let's see …

shanev:
The wallet rename is much more clear. 

I left `Secp256k1DerivationJson` as is, or do you think it makes sense to also rename it to `Secp256k1HdDerivationJson`?

Renamed `fromPrivateKey` to `fromPrivkey`, though another option could be to call it something like `fromPrivateHexKey` and do the hex string conversion inside. Thoughts?

I'll add an example user case. Had a bit of trouble getting CLI examples running, but it can be used in https://github.com/CosmWasm/cosmjs/blob/master/packages/cli/examples/delegate.ts for example.

Note: Still fixing / renaming things, and have to run pending tests.

webmaster128:
> I'll add an example user case. Had a bit of trouble getting CLI examples running, but it can be used in https://github.com/CosmWasm/cosmjs/blob/master/packages/cli/examples/delegate.ts for example.

How the wallet works is clear. I think a demo together with tor.us would be interesting to see, but not in this repo. The CLI would not support this integration because torus seems to require browsers.

I'll do the renaming to `Secp256k1HdWallet` in a separate PR for a few logistic reasons:
- 0.23.0 that will come this week should include the breaking change, no matter when we merge the new wallet
- I want client code to break hard when upgrading from 0.22 to 0.23 instead of silently shipping a different wallet under the same name
- The diff here is extremely hard to review because of the re-use of the old class and file name

shanev:
Sure, can add a demo with Torus in the Dapps repo after we have it working with Stakebird. Then others can see how it works.

Reverted the rename and will submit a separate PR later. Didn't know where ya'll were in the development cycle.

webmaster128:
No worries, you cannot know that. In https://github.com/CosmWasm/cosmjs/pull/453 the renaming is done to free the name `Secp256k1Wallet`. [Once 0.23.0 is out](https://github.com/CosmWasm/cosmjs/milestone/5), we can get back to this PR.

webmaster128:
Alright, so the name `Secp256k1Wallet` is free now. I suggest
- Rename to `Secp256k1Wallet`
- Simplify constructor to just `Secp256k1Wallet.fromKey()`. There is only one type of key that makes sense here.

> though another option could be to call it something like `fromPrivateHexKey` and do the hex string conversion inside.

CosmJS uses `Uint8Array` very consistently for binary data, whih I think we should keep. Users can easily compose the building blocks with whetever encoding they have available, like `Secp256k1Wallet.fromKey(fromHex(hexStringFromSomewhere))`.

webmaster128:
@shanev do you think you can [update](https://github.com/cosmos/cosmjs/pull/452#issuecomment-709051714) this PR at some point this month? Otherwise I can take over this work in order to include it in 0.24.0.

shanev:
@webmaster128 Feel free to take it over. Won't be able to get to it this month unfortunately.

webmaster128:
Closing in favour of https://github.com/cosmos/cosmjs/pull/488

webmaster128:
Secp256k1Wallet will be released as part of 0.24.x. Thanks again for bringing this up.

PR #454, opened 2020-10-08
Add ERC20_ENABLED to make the contract optional

### LINKED ISSUE #450: Make erc20 contract optional in tests
Introduce `ERC20_ENABLED`, which indicates that on top of `WASMD_ENABLED` an [erc20 contract](https://github.com/CosmWasm/cosmwasm-examples/tree/master/erc20) is deployed to the local test chain.

webmaster128:
Closes #450

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -136,6 +136,7 @@ jobs:
       - run:
           environment:
             WASMD_ENABLED: 1
+            ERC20_ENABLED: 1
             SIMAPP_ENABLED: 1
             TENDERMINT_ENABLED: 1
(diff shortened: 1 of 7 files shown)


willclarktech (commented):
Looks good, but don't we want some/all of the search tests to use hackatom?

webmaster128:
> Looks good, but don't we want some/all of the search tests to use hackatom?

That would be nice indeed. The fewer dependencies we have on ERC20, the better. I only commented on the easiest/most critical ones.

willclarktech (approved):

Issue #412, opened 2020-09-02
Defense against prototype pollution and supply chain attacks


kriskowal:
At @agoric, we’ve begun investigating how compatible CosmJS is with [SES](https://github.com/Agoric/SES-shim/tree/master/packages/ses#secure-ecmascript-ses) such that projects using CosmJS can use tools like SES and [LavaMoat](https://github.com/LavaMoat/lavamoat) to mitigate supply chain and prototype pollution attacks.

To evaluate whether a module and its transitive dependencies can be initialized under SES, I’ve locally run the following script in each of the CosmJS library packages. These are the packages that all have a single entry-module named `./build/index.js`.

```js
require("ses/lockdown");
lockdown();
console.log(require("./build/index.js"));
```

This provides some basic assurance that these libraries could be used in a project using SES. Likely incompatibilities usually arise when a library depends on a shim or polyfill, and these can usually be fixed by using the corresponding [“ponyfill”](https://ponyfoo.com/articles/polyfills-or-ponyfills), like `colors/safe` instead of `colors`. (Though I would note, CosmJS depends much more heavily on the equivalent `ansi-colors` module that does not perform any ~~prototype attacks~~ monkey-patching.)

- [x] encoding
- [x] math
- [x] proto-signing
- [x] utils
- [x] tendermint-rpc
- [x] stargate
- [x] launchpad
- [x] crypto
- [x] cosmwasm

The following depend on a broken ponyfill, `observable-symbol/ponyfill`, which we haved worked with Ben Lesh to fix. This fix comes in a new major version, so an explicit upgrade should get the remaining libraries working.
https://github.com/benlesh/symbol-observable/pull/48

- [ ] json-rpc
- [ ] socket
- [ ] stream

Of the two applications, `faucet` and `cli`, the faucet appears to work with the SES lockdown snippet at the head of its `bin`.

```js
#!/usr/bin/env node
require("ses/lockdown");
lockdown();
```

- [x] faucet
- [ ] cli

The CLI runs into this mysterious error, far far from any likely cause:

```
  cosmjs/node_modules/ast-types/lib/types.js:281
                      throw new Error("missing name");
                      ^

  [Error <Object <[Object: null prototype] {}>>: missing name
    at Object.from (cosmjs/node_modules/ast-types/lib/types.js:281:27)
    at cosmjs/node_modules/ast-types/lib/types.js:247:71
    at Array.map (<anonymous>)
[...]

PR #462, opened 2020-10-13
Update xstream

willclarktech:
Part of #412


webmaster128 (approved):

webmaster128:
@kriskowal FYI.

As you can see herer, we don't pull in @types/globalthis now :)

PR #463, opened 2020-10-13
Add convenience hash functions

### LINKED ISSUE #459: Create simple hash functions
I hate redundant APIs, I really do. But looking at this user facing code snippet: https://github.com/CosmWasm/cosmjs/wiki/What-can-CosmJS-do-for-me%3F#deriving-a-tendermint-address-from-a-consensus-pubkey I think it is time to make a compromise between redundancy and simple caller code. I can't even remember when we needed incremental hashing.

Let's create hash functions with a single input like `function sha256(in: Uint8Array): Uint8Array;` that is equivalent to `new Sha256(in).digest()` for
- sha1
- sha256
- sha512
- ripemd160
- keccak256

willclarktech:
Closes #459

### DIFF
--- a/packages/crypto/src/sha.ts
+++ b/packages/crypto/src/sha.ts
@@ -26,6 +26,7 @@ export class Sha1 implements HashFunction {
   }
 }
 
+/** Convenience function equivalent to `new Sha1(data).digest()` */
 export function sha1(data: Uint8Array): Uint8Array {
   return new Sha1(data).digest();
 }
@@ -53,6 +54,7 @@ export class Sha256 implements HashFunction {
   }
 }
 
+/** Convenience function equivalent to `new Sha256(data).digest()` */
 export function sha256(data: Uint8Array): Uint8Array {
   return new Sha256(data).digest();
 }
@@ -80,6 +82,7 @@ export class Sha512 implements HashFunction {
   }
 }
 
+/** Convenience function equivalent to `new Sha512(data).digest()` */
 export function sha512(data: Uint8Array): Uint8Array {
   return new Sha512(data).digest();
 }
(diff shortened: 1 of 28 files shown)


willclarktech:
@webmaster128 Do we need unit tests for these functions?

webmaster128:
> Do we need unit tests for these functions?

Let's have one test with one input each. Then we at least know nothing is broken and the rest is tested in the hasher class.

webmaster128 (commented):
Code is nice, thanks.

Could you add a CHANGELOG entry for 0.23.1 and short doc comments for the functions explaining those are queivalent to `new Hasher(in).digest()`?

packages/crypto/src/sha.ts
```diff
@@ -79,7 +81,7 @@ export class Sha512 implements HashFunction {
     return new Uint8Array(this.impl.digest());
   }
 }
-
+/** Convenience function equivalent to `new Sha512(data).digest()` */
```
webmaster128, line 84: Looks like an empty line is lost here

packages/crypto/types/sha.d.ts
```diff
@@ -6,17 +6,20 @@ export declare class Sha1 implements HashFunction {
...
 }
+export declare function sha256(data: Uint8Array): Uint8Array;
 export declare class Sha512 implements HashFunction {
   readonly blockSize: number;
   private readonly impl;
   constructor(firstData?: Uint8Array);
   update(data: Uint8Array): Sha512;
   digest(): Uint8Array;
 }
+export declare function sha512(data: Uint8Array): Uint8Array;
```
webmaster128, line 25: Could you rebuild the types to include the doc comment?

webmaster128 (commented):
Almost there :)

webmaster128 (approved):

PR #464, opened 2020-10-13
 🔥 Deduplicate launchpad logs module

### LINKED ISSUE #460: Deduplicate logs module between @cosmjs/launchpad and @cosmjs/cosmwasm
For historic reasons there is `packages/launchpad/src/logs.ts` and `packages/cosmwasm/src/logs.ts` that are more or less the same.

Log handling in @cosmjs/launchpad should be generic enough to support all modules. So most likely the entire `packages/cosmwasm/src/logs.ts` can be refactored away.

willclarktech:
Closes #460


packages/cosmwasm/src/index.ts
```diff
@@ -1,6 +1,3 @@
-import * as logs from "./logs";
-export { logs };
-
```
webmaster128, line 3: This is th point where we break a public API. Not sure anyone would notice, but still

webmaster128 (approved):
Nice, but let's put it on hold until for a while

webmaster128 (approved):
Could you rebase this to make CI happy?

PR #467, opened 2020-10-13
Remove unused axios dependency from cosmwasm

willclarktech:
Not sure why this was there


webmaster128:
> Not sure why this was there

One upon a time, @cosmjs/launchpad and @cosmjs/cosmwasm were one package. I copied and remaned it, then removed stuff step by step.

webmaster128 (approved):

PR #474, opened 2020-10-15
Update and re-enable erc20 contract

webmaster128:


scripts/wasmd/start.sh
```diff
@@ -37,7 +37,7 @@ echo "wasmd running and logging into $WASMD_LOGFILE"
 # sleep 3 && cat "$WASMD_LOGFILE"
 
 # Use a large timeout because of potentially long image download in `docker run`
-if ! timeout 180 bash -c "until docker inspect -f '{{.State.Running}}' '$CONTAINER_NAME' &> /dev/null; do sleep 0.5; done"; then
+if ! timeout 180 bash -c "until [ \"\$( docker container inspect -f '{{.State.Status}}' \"$CONTAINER_NAME\" 2> /dev/null )\" = \"running\" ]; do sleep 0.5; done"; then
```
webmaster128, line 40: Before, `docker inspect` returned a failure code as long as the container was not available. Once it existed but was not yet running, `docker inspect` succeeded even when the text output was "false".

scripts/wasmd/deploy_erc20.js
```diff
@@ -137,7 +137,7 @@ async function main() {
   const wallet = await Secp256k1HdWallet.fromMnemonic(alice.mnemonic);
   const client = new SigningCosmWasmClient(httpUrl, alice.address0, wallet);
 
-  const wasm = fs.readFileSync(__dirname + "/contracts/cw-erc20.wasm");
+  const wasm = fs.readFileSync(__dirname + "/contracts/cw_erc20.wasm");
```
willclarktech, line 140: Why change the filename?

willclarktech (approved):

scripts/wasmd/deploy_erc20.js
webmaster128: The older rust-optimizer always created a `contract.wasm`, which we then manually renamed. The new rust-optimizer uses the output names from Rust directly to allow writing multiple contracts in one out directory. So next time the nameservices is recompiled, it gets the underscore too.

See also https://github.com/CosmWasm/cosmwasm-examples/commit/a13efad8e9c4a8fd009291f13b1f6912464fd26d

PR #476, opened 2020-10-15
Upgrade TS to 4.0

webmaster128:
because we can


packages/launchpad-ledger/tsconfig.json
```diff
@@ -5,7 +5,11 @@
     "outDir": "build",
     "declarationDir": "build/types",
     "rootDir": "src",
+    "typeRoots": [
+      "./custom_types",
+      "../../node_modules/@types"
```
willclarktech, line 10: What's the precedence of this list? Eg is this the right order if some official types are out of date and we need to override them?

webmaster128: Yes, this is what I researched long time ago when I first used this (first match is used). Will double check.

webmaster128: See https://www.typescriptlang.org/tsconfig#typeRoots. However, I could not find any documentation about precedence. Let's dig into this when we have a type package twice.

PR #482, opened 2020-10-20
Add SigningStargateClient

### LINKED ISSUE #272: Create CosmosClient/SigningCosmosClient in @cosmjs/stargate

### LINKED ISSUE #479: Research: Can we use OfflineSigner for Stargate?
Can we use the existing OfflineSigner interface for Stargate transactions?
- If yes, great, done.
- If no, can we adjust it slightly to only need one interface for Launchpad and Stargate?
- If yes, great, let's do that.
- If no, create a similar interfce for Stargate

willclarktech:
Closes #272
Closes #479

This is my suggestion for how to extend `OfflineSigner` for other sign modes (it can easily be
extended beyond `SIGN_MODE_DIRECT`).

There’s a weird abuse of the type system here but it removes the need for people to redundantly specify two types: https://github.com/cosmos/cosmjs/blob/272-signing-stargate-client/packages/launchpad/src/signer.ts#L64


packages/launchpad-ledger/src/ledgersigner.ts
```diff
@@ -1,22 +1,28 @@
...
+  serializeSignDoc,
+  SignMode,
 } from "@cosmjs/launchpad";
-import { serializeSignDoc, SignResponse } from "@cosmjs/launchpad";
 import Transport from "@ledgerhq/hw-transport";
 
 import { LaunchpadLedger, LaunchpadLedgerOptions } from "./launchpadledger";
 
-export class LedgerSigner implements OfflineSigner {
+export class LedgerSigner implements OfflineSigner<AminoSignSpec> {
```
ethanfrey, line 17: Very cool actually how this `OfflineSigner<T>` can set two unrelated variable types in `sign`

packages/launchpad/src/secp256k1hdwallet.ts
```diff
@@ -259,10 +263,11 @@ export class Secp256k1HdWallet implements OfflineSigner {
     ];
   }
 
-  public async sign(signerAddress: string, signDoc: StdSignDoc): Promise<SignResponse> {
+  public async sign(signerAddress: string, signJob: AminoSignJob): Promise<AminoSignResponse> {
```
ethanfrey, line 266: This should support sign mode direct, right? Or is that a separate type in stargate?

If the wallet can only sign *either* legacy *or* direct/proto (which is my current understanding with the type system), then that means they will only have exactly one sign mode. And you should replace `public readonly signModes: readonly SignMode[];` with `public readonly signMode: SignMode;`

packages/launchpad/src/signer.ts
```diff
@@ -10,7 +9,17 @@ export interface AccountData {
...
 
-export interface SignResponse {
+export interface AminoSignJob {
+  readonly mode: SignMode.LegacyAminoJson;
+  readonly signDoc: StdSignDoc;
+}
+
+export interface DirectSignJob {
+  readonly mode: SignMode.Direct;
+  readonly signBytes: Uint8Array;
```
ethanfrey, line 19: Why is this encoded bytes? There is no higher level objects we would encode? Or we assume the caller knows how to proto serialize?

We don't require the caller to json/amino serialize the StdSignDoc in AminoSignJob, so I wonder why we have to here. If this is a current technical limitation, that is fine, just add a `FIXME` to this type. If this is a conscious decision as to why this is the proper way to do it, can you please document why?

packages/launchpad/src/signer.ts
```diff
@@ -19,7 +28,25 @@ export interface SignResponse {
   readonly signature: StdSignature;
 }
 
-export interface OfflineSigner {
+export interface DirectSignResponse {
+  readonly signed: Uint8Array;
```
ethanfrey, line 32: Same question here as to `DirectSignJob.signBytes`

packages/launchpad/src/signer.ts
```diff
@@ -34,5 +61,5 @@ export interface OfflineSigner {
    * @param signerAddress The address of the account that should sign the transaction
    * @param signDoc The content that should be signed
    */
-  readonly sign: (signerAddress: string, signDoc: StdSignDoc) => Promise<SignResponse>;
+  readonly sign: (signerAddress: string, signDoc: T["signJob"]) => Promise<T["signResponse"]>;
```
ethanfrey, line 64: what happens to this signature if we implement `OfflineSigner<AminoSignSpec | DirectSignSpec>`?

Would it be:

`readonly sign: (signerAddress: string, signDoc: AminoSignJob | DirectSignJob) => Promise<AminoSignResponse | DirectSignResponse>;`
	
Or a compiler error?

packages/stargate/src/signingstargateclient.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+    it("can be constructed with custom gas price", async () => {
+      const wallet = await DirectSecp256k1Wallet.fromMnemonic(faucet.mnemonic);
+      const gasPrice = GasPrice.fromString("3.14utest");
+      const client = await SigningStargateClient.connectWithWallet(
+        simapp.tendermintUrl,
+        wallet,
+        undefined,
+        gasPrice,
+      );
+      const openedClient = (client as unknown) as PrivateSigningStargateClient;
```
ethanfrey, line 47: This is an odd construct. I will look at `SigningStargateClient` why you do this

packages/stargate/src/signingstargateclient.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    it("can be constructed with custom gas limits", async () => {
+      const wallet = await DirectSecp256k1Wallet.fromMnemonic(faucet.mnemonic);
+      const gasLimits = {
+        send: 160000,
+      };
+      const client = await SigningStargateClient.connectWithWallet(
+        simapp.tendermintUrl,
+        wallet,
+        undefined,
```
ethanfrey, line 69: These are lots of optional fields. I am happy with one `option?: T` at the end, which may be there. With two, where you need `(x, undefined, y)` it starts to get ugly. With 3 here even more.

Why not do `connectWithWallet(url, wallet, options?: Options)`, where

```ts
interface Options {
   registry: Registry,
   gasPrice: string
   gasLimits: T
}
```

usage would be simple as well, and much clearer:

`await SigningStargateClient.connectWithWallet(simapp.tendermintUrl, wallet, {gasLimits});`

packages/stargate/src/signingstargateclient.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+            },
+          ],
+          gas: "160000",
+        },
+      });
+    });
+  });
+
+  describe("sendTokens", () => {
+    it("works", async () => {
```
ethanfrey, line 116: Very nice flow demonstrated here.

packages/stargate/src/signingstargateclient.ts
```diff
@@ -0,0 +1,121 @@
...
+import { cosmos } from "./codec";
+import { BroadcastTxResponse, StargateClient } from "./stargateclient";
+
+const { TxRaw } = cosmos.tx.v1beta1;
+
+const defaultGasPrice = GasPrice.fromString("0.025ucosm");
+const defaultGasLimits: GasLimits<CosmosFeeTable> = { send: 80000 };
+
+/** Use for testing only */
+export interface PrivateSigningStargateClient {
```
ethanfrey, line 27: Ah, now I see... you use the casts above to expose private variables.

If any of these private variables are security critical, we should do something to make them really private (like wrap in some closure). This may be the case for some signer implementation.

packages/stargate/src/signingstargateclient.ts
```diff
@@ -0,0 +1,121 @@
...
+      messages: messages,
+      memo: memo,
+    };
+    const txBodyBytes = this.registry.encode({
+      typeUrl: "/cosmos.tx.v1beta1.TxBody",
+      value: txBody,
+    });
+    const gasLimit = Int53.fromString(fee.gas).toNumber();
+    const authInfoBytes = makeAuthInfo([pubkeyAny], fee.amount, gasLimit, sequence);
+    const signJob = makeDirectSignJob(txBodyBytes, authInfoBytes, chainId, accountNumber);
```
ethanfrey, line 111: Seeing this example, I can clarify my comment on `DirectSignJob.signBytes` above. Why not take an object:
`{txBody, authInfo, chainId, accountNumber}` instead of the bytes - it would be more like the amino style.

(This would just move the logic from here into the signer, but the plus is the signer would understand the bytes and be able to display them to the user).

packages/launchpad/src/signer.ts
```diff
@@ -19,7 +28,25 @@ export interface SignResponse {
...
+}
+
+export interface DirectSignSpec {
+  readonly signJob: DirectSignJob;
+  readonly signResponse: DirectSignResponse;
+}
+
+export type SignSpec = AminoSignSpec | DirectSignSpec;
+
+export interface OfflineSigner<T extends SignSpec> {
```
ethanfrey, line 48: After a bit more reflection, and reading through the code, I think the most straight-forward approach uses more simple types:

```typescript
export interface OfflineAminoSigner {
  readonly getAccounts: () => Promise<readonly AccountData[]>;
  readonly signAmino: (signerAddress: string, signJob: AminoSignJob) => Promise<AminoSignResponse>;
}

export interface OfflineAminoDirect {
  readonly getAccounts: () => Promise<readonly AccountData[]>;
  readonly signDirect: (signerAddress: string, signJob: DirectSignJob) => Promise<DirectSignResponse>;
} 
```

You can add some `isOfflineAminoSigner` and `isOfflineDirectSigner` helpers to detect this at runtime. And a wallet can easily implement both.

You heavily rely on compile-time type info here, which is not present at runtime. When we deal with eg. injected providers from Keplr, how can you determine what support it provides? Also, how to support both in one signer? Ideally a software Secp256k1Wallet could support both and just drop in to any signer - that is not really possible with the current system.

ethanfrey (changes requested):
Massive PR 25k lines! Even without auto-gen, this is pretty huge, though much is just renamings I guess.

Nice work getting the simple send and balance interfaces working in the SigningStargateClient. I would adjust the constructor a bit for ergonomics however. 

I think this OfflineSigner unification is very interesting with the type-system, but when we consider all type info is gone at runtime, this makes it quite fragile for reflection, and also doesn't allow us to create one signer that supports both interfaces. It is very clever how you unify two different objects into one type, but for practice, I think the two unique types (which overlap) would be easier to use.

packages/launchpad-ledger/src/ledgersigner.ts
willclarktech: I'm still not sure how I feel about this, but was inspired by @abefernan here: https://github.com/CosmWasm/dApps/pull/45#discussion_r500124760

packages/stargate/src/signingstargateclient.spec.ts
willclarktech: This is just so we can access a private property in the tests. I copied the pattern from what we have in Launchpad.

packages/stargate/src/signingstargateclient.ts
willclarktech: Yes. So far I don't believe this applies to our signers - we even expose `mnemonic` as a public property somewhere.

packages/launchpad/src/signer.ts
willclarktech: Yes that's fine. In fact I was hoping it would match up the inputs and outputs but this compiles:
```ts
  public async sign(
    _signerAddress: string,
    _signJob: AminoSignJob | DirectSignJob,
  ): Promise<DirectSignResponse> {
    throw new Error("xxx");
  }
```

packages/stargate/src/signingstargateclient.ts
```diff
@@ -120,8 +119,26 @@ export class SigningStargateClient extends StargateClient {
       value: txBody,
     });
     const gasLimit = Int53.fromString(fee.gas).toNumber();
-    const authInfoBytes = makeAuthInfoBytes([pubkeyAny], fee.amount, gasLimit, sequence);
 
+    if (!isOfflineDirectSigner(this.signer)) {
+      const signMode = cosmos.tx.signing.v1beta1.SignMode.SIGN_MODE_LEGACY_AMINO_JSON;
+      const authInfoBytes = makeAuthInfoBytes([pubkeyAny], fee.amount, gasLimit, sequence, signMode);
+      const msgs = messages.map((msg) => ({
+        type: "cosmos-sdk/MsgDelegate",
```
ethanfrey, line 127: I guess we need a whole lookup table of amino names to proto names? I searched through and there doesn't seem to be an explicit map. But you can look in the various `codec.go` files and see the amino names attached to a struct. The protoname should be the same as the struct. I guess a mapping of proto.Any name -> amino name would be sufficient for this to work?

https://github.com/cosmos/cosmos-sdk/blob/release/v0.40.x/x/staking/types/codec.go#L12-L18

ethanfrey: But other than that (WIP) point, this looks like a good approach

ethanfrey (commented):
Nice enhancements

packages/stargate/src/signingstargateclient.ts
```diff
@@ -0,0 +1,188 @@
...
+function snakifyMsgValue(obj: Msg): Msg {
+  return {
+    ...obj,
+    value: Object.entries(obj.value).reduce(
+      (snakified, [key, value]) => ({
+        ...snakified,
+        [key
+          .split(/(?=[A-Z])/)
+          .join("_")
+          .toLowerCase()]: value,
```
webmaster128, line 42: Can we extract the camel -> snake case conversion in a function that comes with unit tests? There will be some nasty edge cases and then we have a better time debugging it.

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@
   @cosmjs/launchpad instead.
 - @cosmjs/launchpad: Add `Secp256k1Wallet` to manage a single raw secp256k1
   keypair.
+- @cosmjs/launchpad: `OfflineSigner` type’s `sign` method renamed `signAmino`
+  and `SignResponse` type renamed `AminoSignResponse`.
+- @cosmjs/launchpad: `Secp256k1HdWallet.sign` method renamed `signAmino`.
```
webmaster128, line 11: This needs to be done for `Secp256k1Wallet` as well. I assume it was done already, otherwise we get compile fails.

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@
...
+- @cosmjs/launchpad-ledger: `LedgerSigner.sign` method renamed `signAmino`.
+- @cosmjs/proto-signing: Add new package for handling transaction signing with
+  protobuf encoding.
+- @cosmjs/stargate: Add new package for Cosmos-SDK Stargate support.
```
webmaster128, line 15:
```suggestion
- @cosmjs/stargate: Add new package for Cosmos SDK Stargate support.
```

packages/proto-signing/src/signer.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export interface OfflineDirectSigner {
+  readonly getAccounts: () => Promise<readonly AccountData[]>;
+  readonly signDirect: (
+    signerAddress: string,
+    signDoc: cosmos.tx.v1beta1.ISignDoc,
+  ) => Promise<DirectSignResponse>;
+}
+
+export type OfflineSigner = OfflineAminoSigner | OfflineDirectSigner;
```
webmaster128, line 22: Is this or exclusive? Can we have an `OfflineSigner` that implements both interfaces?

packages/proto-signing/src/signer.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function isOfflineDirectSigner(signer: OfflineSigner): signer is OfflineDirectSigner {
+  return (signer as any).signDirect !== undefined;
```
webmaster128, line 25:
```suggestion
  return (signer as OfflineDirectSigner).signDirect !== undefined;
```

This is the documented way to temporarily cast for type checks. It guves you type safety when the name or type of `signDirect` change over time.

packages/proto-signing/src/directsecp256k1wallet.spec.ts
```diff
@@ -54,15 +57,30 @@ describe("DirectSecp256k1Wallet", () => {
...
+      const chainId = "simd-testing";
+      const accountNumber = 1;
+      const signDoc = makeSignDoc(
+        fromHex(bodyBytes),
+        makeAuthInfoBytes([pubkey], fee, gasLimit, sequence),
+        chainId,
+        accountNumber,
+      );
+      const signDocBytes = makeSignBytes(signDoc);
+      const signResponse = await wallet.signDirect(faucet.address, signDoc);
```
webmaster128, line 79: In Lauchpad, we always use destruction here, like

```suggestion
      const { signature } = await wallet.signDirect(faucet.address, signDoc);
```

I think we can do this as well for all `signDirect` calls

webmaster128 (commented):
Beautiful

CHANGELOG.md
willclarktech: Yes, I updated it but didn't add a changelog entry since it's also new in 0.24.

packages/proto-signing/src/signer.ts
willclarktech: It's the union, so you can implement both but you don't have to. E.g. this compiles:
```ts
export class TestSigner implements OfflineAminoSigner, OfflineDirectSigner {
  public constructor() {}

  public getAccounts(): Promise<readonly AccountData[]> {
    throw new Error("not implemented");
  }

  public signDirect(): Promise<DirectSignResponse> {
    throw new Error("not implemented");
  }

  public signAmino(): Promise<AminoSignResponse> {
    throw new Error("not implemented");
  }
}
```

CHANGELOG.md
webmaster128: Oh, right, perfect

packages/proto-signing/src/signer.ts
webmaster128: Good to know, thanks

webmaster128 (approved):

PR #496, opened 2020-10-28
Support Stargate in faucet

### LINKED ISSUE #478: Make faucet work for Stargate chains
Based on https://github.com/cosmos/cosmjs/issues/272 we should have a Stargate faucet. Ideally a second client in the same faucet binary.

I don't know a great way to tell the faucet which client to use. Maybe we try to connect to Tendermint using the provided base URL and if that works, assume it is a stargate backend.

willclarktech:
Closes #478

I tried to make it easy to excise the Launchpad-related code when we need to.

### DIFF
--- a/packages/faucet/src/faucet.spec.ts
+++ b/packages/faucet/src/faucet.spec.ts
@@ -1,6 +1,7 @@
 import { Random } from "@cosmjs/crypto";
 import { Bech32 } from "@cosmjs/encoding";
 import { CosmosClient } from "@cosmjs/launchpad";
+import { StargateClient } from "@cosmjs/stargate";
 import { assert } from "@cosmjs/utils";
 
 import { Faucet } from "./faucet";
@@ -12,7 +13,12 @@ function pendingWithoutWasmd(): void {
   }
 }
 
-const httpUrl = "http://localhost:1317";
+function pendingWithoutSimapp(): void {
+  if (!process.env.SIMAPP_ENABLED) {
+    return pending("Set SIMAPP_ENABLED to enabled Stargate node-based tests");
+  }
+}
+
 const defaultTokenConfig: TokenConfiguration = {
   bankTokens: ["ucosm", "ustake"],
 };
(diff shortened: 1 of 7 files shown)


packages/faucet/src/faucet.ts
```diff
@@ -35,38 +43,28 @@ export class Faucet {
...
+  private readonly clients: { [senderAddress: string]: SigningCosmosClient | SigningStargateClient };
   private readonly logging: boolean;
   private creditCount = 0;
 
   private constructor(
-    apiUrl: string,
     addressPrefix: string,
     config: TokenConfiguration,
     wallets: ReadonlyArray<readonly [string, OfflineSigner]>,
+    clients: { [senderAddress: string]: SigningCosmosClient | SigningStargateClient },
```
webmaster128, line 55: If we pass in the clients here, we don't need `wallets` anymore. The addresses are available in the client's `senderAddress`. We just need to preserve the order of the clients.

packages/faucet/src/faucet.ts
```diff
@@ -89,8 +85,13 @@ export class Faucet {
    * Throws an error if the transaction failed.
    */
   public async send(job: SendJob): Promise<void> {
-    const result = await this.clients[job.sender].sendTokens(job.recipient, [job.amount], constants.memo);
-    assertIsBroadcastTxSuccess(result);
+    const client = this.clients[job.sender];
+    if (client instanceof CosmosClient) {
+      const result = await client.sendTokens(job.recipient, [job.amount], constants.memo);
```
webmaster128, line 90: Impressive that TS understands that if `client` is `CosmosClient`, it also is `SigningCosmosClient`.

But since `CosmosClient` has no `sendTokens`, I'd test `client instanceof SigningCosmosClient`

packages/faucet/src/profile.ts
```diff
@@ -1,19 +1,25 @@
...
+import { DirectSecp256k1Wallet, isOfflineDirectSigner, OfflineSigner } from "@cosmjs/proto-signing";
+import { SigningStargateClient } from "@cosmjs/stargate";
+
+import * as constants from "./constants";
 
 export async function createWallets(
   mnemonic: string,
   addressPrefix: string,
   numberOfDistributors: number,
+  direct = true,
```
webmaster128, line 12: Since we only use direct signing mode in stargate, I'd use a `stargate` flag here as well.

packages/faucet/src/faucet.spec.ts
```diff
@@ -26,143 +32,433 @@ const faucetMnemonic =
...
 
 describe("Faucet", () => {
-  describe("constructor", () => {
-    it("can be constructed", async () => {
-      pendingWithoutWasmd();
-      const faucet = await Faucet.make(httpUrl, defaultAddressPrefix, defaultTokenConfig, faucetMnemonic, 3);
-      expect(faucet).toBeTruthy();
+  describe("launchpad", () => {
+    const apiUrl = "http://localhost:1317";
+    const shouldUseDirectSigning = false;
```
webmaster128, line 37: As mentioned below, this can just be `const stargate = false;`

packages/faucet/src/faucet.spec.ts
```diff
@@ -26,143 +32,433 @@ const faucetMnemonic =
...
 
-  describe("configuredTokens", () => {
-    it("works", async () => {
-      pendingWithoutWasmd();
-      const faucet = await Faucet.make(httpUrl, defaultAddressPrefix, defaultTokenConfig, faucetMnemonic, 3);
-      const tickers = faucet.configuredTokens();
-      expect(tickers).toEqual(["ucosm", "ustake"]);
+  describe("stargate", () => {
+    const apiUrl = "localhost:26657";
+    const shouldUseDirectSigning = true;
```
webmaster128, line 245: As mentioned below, this can just be `const stargate = true;`

webmaster128 (approved):

Issue #494, opened 2020-10-28
Review IBC queries, create test data and improve testing


webmaster128:
In `packages/stargate/src/queries/ibc.spec.ts` we test Stargate IBC queries. However, the testing is very minimal because when we wrote them, we were not able to generate proper test fixtures.

The 6 `expect(response).toBeTruthy(); // TODO: implement checks` should be replaced with tests that check the data.

There was work from Alex and Ethan to generate test data (https://github.com/cosmos/cosmjs/pull/405 and https://github.com/cosmos/cosmjs/pull/406). Let's try to update this work.

Once we have the [IbcExtension](https://github.com/cosmos/cosmjs/blob/master/packages/stargate/src/queries/ibc.ts) tested, I assume we have everything we need for an IBC visualizer data source.

ethanfrey:
Let's not use those linked prs. Data format is also changing we need simapp v0.40.0-rc2

We can provide test vectors when that is ready

ethanfrey:
@willclarktech @webmaster128 

Here is my brief into as to what concepts are needed for the relayer:

`Client/connection` - are almost always 1:1 and refer to headers of another chain (to probe that something came from there). Let's list connections and each can "show details" for full client (in theory there can be multiple connections per client but I have not heard 1 example when this would be in reality, still listing connections means we would support it)

`Port` - each module/contract can bind onto a port. This is similar to TCP, but the same port is used for both sending and receiving.

`Channel` -  this is where information is sent. It involves a handshake between (connection, port) pair (one on each chain) and the channel gets it's own ID. (these ids are unique per connection)

`Packet` - this is a message sent over a channel and will be processed by an app on the other side. For ICS20 this packets says "move x tokens from me to you". There is an ack, which contains a result (success) or error, so the sender know if the money should be refunded locally when it fails

----

I have created https://github.com/CosmWasm/dApps/issues/61 to show the general workflow (3-4 pages we need to show). Here are the key queries (from the protobuf files) that we need working:

Connections:

`rpc Connections(QueryConnectionsRequest) returns (QueryConnectionsResponse)` - includes connection_id, client_id, state

Clients:

`rpc ClientState(QueryClientStateRequest) returns (QueryClientStateResponse)` - includes `google.protobuf.Any client_state`, which we should be able to parse. Currently defined client types are "tendermint", "localhost" and "solo_machine" (not sure where these are defined in code)

[...]

PR #504, opened 2020-11-03
simapp: add ibc example data

alpe:
See #494

This PR adds an IBC client, connection and channel via genesis. 

The IBC example data is  in `scripts/simapp/genesis-ibc.json` and is added to the genesis via `scripts/simapp/generate_template.sh`

To view the data via CLI:
```sh
simd q ibc client states
simd q ibc connection connections
simd q ibc channel channels
```

The data was generated for Stargate `RC1`


webmaster128:
Thank you @alpe! Do I understand correctly that this data works for RC1 and must be re-recreated for RC2?

willclarktech (approved):

PR #510, opened 2020-11-03
Improve IBC query tests

willclarktech:
Part 2 of #494 
~Merge after #509~

### DIFF
--- a/packages/stargate/src/queries/ibc.spec.ts
+++ b/packages/stargate/src/queries/ibc.spec.ts
@@ -1,7 +1,10 @@
 import { Client as TendermintClient } from "@cosmjs/tendermint-rpc";
+import Long from "long";
 
+import { cosmos, ibc } from "../codec";
 import { pendingWithoutSimapp, simapp } from "../testutils.spec";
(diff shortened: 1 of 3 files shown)


packages/stargate/src/queries/ibc.spec.ts
```diff
@@ -1,8 +1,11 @@
 import { Client as TendermintClient } from "@cosmjs/tendermint-rpc";
+import Long from "long";
 
+import { cosmos, ibc } from "../codec";
 import { pendingWithoutSimapp, simapp } from "../testutils.spec";
 import { IbcExtension, setupIbcExtension } from "./ibc";
 import { QueryClient } from "./queryclient";
+import * as ibcTest from "./testutils.spec";
```
webmaster128, line 8: I'm not convinced this is the best way to export/import and organize the `testutils.spec` module. We use this module name usually for all kinds of helper functions, not only data.

Can we either export all the ibc data as a single `ibcData` or rename the module to only contain ibc data?

webmaster128 (approved):

PR #518, opened 2020-11-08
Add DirectSecp256k1Wallet

JakeHartnell:
Adds a `DirectSecp256k1Wallet` for holding a single secp256k1 keypair. Renames the current "DirectSecp256k1Wallet" to "DirectSecp256k1HdWallet", making it consistent with [Secp256k1Wallet](https://cosmos.github.io/cosmjs/latest/launchpad/classes/secp256k1wallet.html) and [Secp256k1HdWallet](https://cosmos.github.io/cosmjs/latest/launchpad/classes/secp256k1hdwallet.html) in `@cosmjs/launchpad`.

Note, I did renaming DirectSecp256k1HdWallet and implementing the new DirectSecp256k1Wallet in the same commit and PR. Please let me know if you would like this separate. Open to other feedback or commits.

Being able to create a wallet that holds a single keypair is vary useful for integrating with things like [Torus](https://tor.us/). :pray:

PR #528, opened 2020-11-13
Export DirectSecp256k1Wallet from proto-signing package

JakeHartnell:
@willclarktech, forgot to export `DirectSecp256k1Wallet` when I separated #518 into two different commits. Noticed today when I was trying to use it in our app. :sweat_smile:


webmaster128 (approved):

webmaster128:
This is released as part of 0.24.0-alpha.7

Issue #501, opened 2020-10-29
Add subkey support


ethanfrey:
This is somewhat optional (it can be part of the ICF milestone, but is not required). After providing a [specification for cosmjs multisig support](https://github.com/cosmos/cosmjs/issues/416#issuecomment-718904679), I figured that this would be a simpler first step.

Add subkey support based on the [cw1 spec](https://github.com/CosmWasm/cosmwasm-plus/tree/master/packages/cw1), and in particular for the admin/testing, with the [cw1-subkeys implementation](https://github.com/CosmWasm/cosmwasm-plus/tree/master/contracts/cw1-subkeys). I would add two adapters there for each.

Ideal User flow (wallet) - submission. This should work for *any* cw1-compliant contract:

* Enter a contract address and "assume identity" (wallet checks if this is cw1 and you are a member)
* When using an dApp show all queries as the contract address (eg. wallet balance)  - this is a valid address from the signer
* When hitting eg. "send tokens" or "buy name" in a dApp, instead of signing the tx directly, this should wrap the transaction into a `CosmosMsg` and submit that to the cw1 contract. The contract will either reject the msg (insufficient rights) or execute it the same tx. 

This just requires the [`execute` message](https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw1-subkeys/src/msg.rs#L19). You could also use the [`can_send` query](https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw1-subkeys/src/msg.rs#L64-L67) as a pre-flight to see if it the sender has authority to send that.

Ideal User Flow (wallet) - administration This is only for `cw1-subkeys`. Just adding a contract-specific wrapper is plenty, but make this easy for use in tests or clients:

* Enter a cw1-subkey address and view it's state
* Check if your current key is [an admin of this contract](https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw1-subkeys/src/msg.rs#L54)
*  Check if you can [send bank tokens from the contract](https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw1-subkeys/src/msg.rs#L57)
[...]

PR #535, opened 2020-11-17
Add CW3 multisig support

### LINKED ISSUE #416: First-class multi-sig support
## Step 1: Decide on implementation

We can choose between
1. off-chain authorization aggregation (multiple signatures on one transaction) or
2. on-chain authorization aggregation (multiple independent transactions to trigger an action).

## 1. off-chain authorization aggregation

**Advantages:** compatibility with [the Comos multisig ](https://hub.cosmos.network/master/resources/gaiacli.html#multisig-transactions) method (multiple pubkey squeezed into one multisig "pubkey")
**Disadvantages:** sequence numbers must not change during the signing process (i.e. a user that signed must not broadcast a different transaction until the multisig transaction is comitted); starting with Stargate, the signing mode must be coordinated before composing the transaction; off-chain infrastructure requires to aggregate the signatures.

## 2. on-chain authorization aggregation

**Advantages:** signing process is independent (one single signature tx per authorization); can be implemented nicely in wallets that track the authorization state
**Disadvantages:** higher fees due to multiple transactions; not yet standardized as far as I know

Option 2. can either be implemented using groups accounts (do they exist yet?) or CosmWasm.

## Step 2: Implement

willclarktech:
Closes #416

Do we need more comprehensive query tests? They just use the normal `CosmWasmClient` query method.


packages/cosmwasm/src/multisigcosmwasmclient.ts
```diff
@@ -0,0 +1,76 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { BroadcastMode, GasLimits, GasPrice, OfflineSigner } from "@cosmjs/launchpad";
+
+import { CosmWasmFeeTable, ExecuteResult, SigningCosmWasmClient } from "./signingcosmwasmclient";
+
+export type Duration =
```
webmaster128, line 6: Why is this called "duration"? Isn't this a point in time, not a time span?

packages/cosmwasm/src/multisigcosmwasmclient.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+      const { logs } = await proposer.createMultisigProposal(
+        "My proposal",
+        "A proposal to propose proposing proposals",
+        [msg],
+      );
+      const wasmEvents = logs[0].events.find((event) => event.type === "wasm");
+      assert(wasmEvents, "Wasm events not found in logs");
+      const proposalIdAttribute = wasmEvents.attributes.find((log) => log.key === "proposal_id");
+      assert(proposalIdAttribute, "Proposal ID not found in logs");
+      const proposalId = parseInt(proposalIdAttribute.value, 10);
```
webmaster128, line 124: This requires the proposer to manually inform the voter about the proposal ID to vote on.

Could the voter get the open proposals from chain?

packages/cosmwasm/src/multisigcosmwasmclient.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+      pendingWithoutLaunchpad();
+      pendingWithoutCw3();
+      const wallet = await Secp256k1HdWallet.fromMnemonic(alice.mnemonic);
+      const client = new MultisigCosmWasmClient(
+        launchpad.endpoint,
+        alice.address0,
+        wallet,
+        deployedCw3.instances[0],
+      );
+      const result = await client.queryContractSmart(deployedCw3.instances[0], { threshold: {} });
```
webmaster128, line 40: I think it would be nice to expose query functions from `MultisigCosmWasmClient`, which allow doing the most important CW3 queries directly. This would internalize the query message format `{ threshold: {} }`.

packages/cosmwasm/src/multisigcosmwasmclient.ts
```diff
@@ -0,0 +1,76 @@
...
+    };
+
+export enum Vote {
+  Yes = "yes",
+  No = "no",
+  Abstain = "abstain",
+  Veto = "veto",
+}
+
+export class MultisigCosmWasmClient extends SigningCosmWasmClient {
```
webmaster128, line 21: What about renaming this to `Cw3CosmWasmClient`? Looking at https://github.com/cosmos/cosmjs/issues/416#issuecomment-718904679 and #501 there seems to be some variety of multisig solution upcoming and right now I don't see how to magically abstract away all of this. But we can have a bunch of this wrapper classes, which can be used directly or serve as a template for creating a similar but incompatible one.

packages/cosmwasm/src/multisigcosmwasmclient.ts
willclarktech: Sorry, this should be `Expiration`.

webmaster128 (approved):
👏

packages/cosmwasm/src/cw3cosmwasmclient.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+            ],
+          },
+        },
+      };
+      await proposer.createMultisigProposal("My proposal", "A proposal to propose proposing proposals", [
+        msg,
+      ]);
+      const { proposals } = await voter.reverseProposals({ limit: 1 });
+      const proposalId = proposals[0].id;
+      const voteResult = await voter.voteMultisigProposal(proposalId, Vote.Yes);
```
ethanfrey, line 260: I would test that trying to execute before this second vote is falsy. To ensure we are covering this case.

packages/cosmwasm/src/cw3cosmwasmclient.ts
```diff
@@ -0,0 +1,197 @@
...
+  readonly startAfter?: number;
+  readonly limit?: number;
+}
+
+interface StartAfterStringPaginationOptions {
+  readonly startAfter?: string;
+  readonly limit?: number;
+}
+
+export class Cw3CosmWasmClient extends SigningCosmWasmClient {
```
ethanfrey, line 73: It would be good to easily query the balance of the contract. When using this client, we assume we are acting on behalf of the multisig and our senderAddress is used to sign our txs, but otherwise unimportant.

However, `getBalances()` will be returning info about `senderAddress`, not `cw3ContractAddress`, right? I think we should use that contract address as the default account to query. (And/or expose it somehow from the client, the senderAddress is `public readonly`, correct?)

packages/cosmwasm/src/cw3cosmwasmclient.ts
```diff
@@ -0,0 +1,197 @@
...
+        start_after: startAfter,
+        limit: limit,
+      },
+    });
+  }
+
+  public createMultisigProposal(
+    title: string,
+    description: string,
+    msgs: ReadonlyArray<Record<string, unknown>>,
```
ethanfrey, line 161: This could be more tightly specified to the set of all CosmosMsg types, but that would exclude custom message types, so fine as this I guess. Maybe nice to add better typing though, and then allow `custom: ReadonlyArray<Record<string, unknown>>`, as we know the shape of `bank`, `staking`, `wasm` queries and it is easy to make typos when creating those.

This should be a simple enhancement

packages/cosmwasm/src/cw3cosmwasmclient.ts
```diff
@@ -0,0 +1,197 @@
...
+    const handleMsg = {
+      propose: {
+        title: title,
+        description: description,
+        msgs: msgs,
+        earliest: earliest,
+        latest: latest,
+      },
+    };
+    return this.execute(this.cw3ContractAddress, handleMsg, memo);
```
ethanfrey, line 175: This came to me as an idea for a future iteration. I guess we could/should wait til stargate to do this to avoid wasted work:

We could create a helper function to convert from CosmosSDK message types to CosmosMsg types. Thus, you could take a normal bank/send message (or transaction) and then convert it to the corresponding CosmosMsg types and embed it in a proposal. This would be a stepping stone for apps that were unaware of the mutlisig - they request you to sign a staking message, and the wallet would convert that to the proper CosmosMsg type and embed it in a proposal.

You could click on the normal UI without modifying the dApp and the wallet would just turn this into a multisig action. Of course this would fail for anything but bank, staking, and wasm calls, but that is still a big and important space to cover.

What do you think? If so, let's make an issue for it

ethanfrey (commented):
Some belated feedback. A few requested enhancements, and some ideas for the future

packages/cosmwasm/src/cw3cosmwasmclient.ts
willclarktech: https://github.com/cosmos/cosmjs/issues/548

PR #536, opened 2020-11-17
Upgrade simapp to rc3

### LINKED ISSUE #524: Upgrade simapp from 0.40.0-rc2 to 0.40.0-rc3
and adapt Stargate client
~Depends on #515~

webmaster128:
Closes #524

The upgrade seems to be simple, but unfortunately our Tendermint RPC client cannot determine the backend's Tendermint version (https://github.com/cosmos/cosmos-sdk/issues/7963).


willclarktech (commented):
Looks good to me, except for the version bug

webmaster128:
@willclarktech I added a workaround in the last commit. It hurts. But we need to unblock ourselves to make progress in https://github.com/cosmos/cosmjs/issues/491

PR #538, opened 2020-11-18
Reduce dependency on tendermint version detection

webmaster128:
Auto-detecting Tendermint versions is a nice idea in general, but is not that important anymore. We know the Tendermint version for each Cosmos SDK version.

This is an attempt to work around https://github.com/cosmos/cosmos-sdk/issues/7963 by not having to rely on the value in the version field in cases where we know what version the backend has.


packages/tendermint-rpc/types/adaptors/v0-33/hasher.spec.d.ts
```diff
@@ -0,0 +1 @@
+export {};
```
willclarktech, line 1: Is this here because of the additional nesting?

webmaster128: I think so, yes. Unfortunately the `/**/` path component does not work well in shx. Any ideas how to best fix this?

packages/tendermint-rpc/src/client.spec.ts
```diff
@@ -42,32 +42,34 @@ function randomString(): string {
...
 
 function defaultTestSuite(rpcFactory: () => RpcClient, adaptor: Adaptor, expected: ExpectedValues): void {
-  it("can connect to tendermint with known version", async () => {
-    pendingWithoutTendermint();
-    const client = new Client(rpcFactory(), adaptor);
-    expect(await client.abciInfo()).toBeTruthy();
-    client.disconnect();
-  });
+  describe("create", () => {
+    it("can auto-discover Tendermint version and communicate", async () => {
```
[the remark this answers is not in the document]
webmaster128: Right. The empty version field only happens when Tendermint is embedded in Cosmos SDK builds. Then we run those Tendermint tests against the Tendermint backends, we always get a version.

packages/tendermint-rpc/types/adaptors/v0-33/hasher.spec.d.ts
willclarktech: I mean, manually add additional levels? It's not pretty but we're not likely to get much deeper than this any time soon.

willclarktech: If we had all our tests in a `__tests__` directory this wouldn't be an issue.

webmaster128:
Okay, two more commits addressing the comments

willclarktech (approved):

PR #550, opened 2020-11-24
Fix omitDefaults/omitDefault for Long

### LINKED ISSUE #549: Proto signing fails when accountNumber = 0
All the tests for proto signing use `1` as the `accountNumber`:

https://github.com/cosmos/cosmjs/blob/c0b714bc5fb527dddf36e7f579a26e281268a7ab/packages/proto-signing/src/signing.spec.ts#L94

https://github.com/cosmos/cosmjs/blob/c0b714bc5fb527dddf36e7f579a26e281268a7ab/packages/proto-signing/src/directsecp256k1hdwallet.spec.ts#L71

https://github.com/cosmos/cosmjs/blob/c0b714bc5fb527dddf36e7f579a26e281268a7ab/packages/proto-signing/src/directsecp256k1wallet.spec.ts#L47

We had a situation where we tried to send tokens from `accountNumber: 0`, which gets converted to a `Long` of `{low: 0, high: null, unsigned: null}` here:

https://github.com/cosmos/cosmjs/blob/c0b714bc5fb527dddf36e7f579a26e281268a7ab/packages/proto-signing/src/signing.ts#L44

`omitDefaults` then recursively converts this to a plain JS object of `{low: null, high: null, unsigned: null}` here:

https://github.com/cosmos/cosmjs/blob/c0b714bc5fb527dddf36e7f579a26e281268a7ab/packages/proto-signing/src/adr27.ts#L43

This leads to an invalid signature (ref: https://github.com/tendermint/vue/issues/39)

@webmaster128 @jgimeno @vince19972 @fadeev

webmaster128:
Closes #549

### DIFF
--- a/packages/proto-signing/src/adr27.spec.ts
+++ b/packages/proto-signing/src/adr27.spec.ts
@@ -1,4 +1,5 @@
 import { fromHex } from "@cosmjs/encoding";
+import Long from "long";
 import { parse } from "protobufjs";
 
 import { omitDefault, omitDefaults } from "./adr27";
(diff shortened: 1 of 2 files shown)


willclarktech (approved):
💆

PR #554, opened 2020-11-26
CosmosMsg types for multisig client

### LINKED ISSUE #547: Specify CW3 Multisig message types
We currently allow a very loose type for msgs in `Cw3CosmWasmClient`:
https://github.com/cosmos/cosmjs/blob/246da8f4d79d8ced403660df88b714c6584098d3/packages/cosmwasm/src/cw3cosmwasmclient.ts#L32

We should restrict the types according to https://github.com/CosmWasm/cosmwasm/blob/v0.12.0/packages/std/src/results/cosmos_msg.rs#L13-L23 (and allow anything under `custom`.

willclarktech:
Closes #547


packages/cosmwasm/src/msgs.ts
```diff
@@ -144,3 +144,75 @@ export interface MsgMigrateContract extends Msg {
...
+    readonly send: readonly Coin[];
+    readonly label?: string;
+  };
+}
+
+export interface WasmMsg {
+  readonly wasm: WasmExecuteMsg | WasmInstantiateMsg;
+}
+
+export type CosmosMsg = BankMsg | CustomMsg | StakingMsg | WasmMsg;
```
webmaster128, line 218: Can you link the definition of the type in a doc comment (https://github.com/CosmWasm/cosmwasm/blob/v0.12.0/packages/std/src/results/cosmos_msg.rs#L10-L23)?

packages/cosmwasm/src/msgs.ts
```diff
@@ -144,3 +144,75 @@ export interface MsgMigrateContract extends Msg {
 export function isMsgMigrateContract(msg: Msg): msg is MsgMigrateContract {
   return (msg as MsgMigrateContract).type === "wasm/MsgMigrateContract";
 }
+
```
webmaster128, line 147: All those messages 👇 live on a different level than the existing ones 👆. What we have above is sent to the chain. The new types are sent to the contract. I have no great naming suggestion that is better than the current naming, but let's put them in a different file to indicate those are different things.

packages/cosmwasm/src/cosmosmsg.ts
```diff
@@ -0,0 +1,77 @@
...
+  };
+}
+
+export interface WasmMsg {
+  readonly wasm: WasmExecuteMsg | WasmInstantiateMsg;
+}
+
+/** These definitions are derived from CosmWasm:
+ * https://github.com/CosmWasm/cosmwasm/blob/v0.12.0/packages/std/src/results/cosmos_msg.rs#L10-L23
+ * */
```
webmaster128, line 76:
```suggestion
 */
```

packages/cosmwasm/types/cosmosmsg.d.ts
```diff
@@ -0,0 +1,65 @@
...
+    readonly send: readonly Coin[];
+    readonly label?: string;
+  };
+}
+export interface WasmMsg {
+  readonly wasm: WasmExecuteMsg | WasmInstantiateMsg;
+}
+/** These definitions are derived from CosmWasm:
+ * https://github.com/CosmWasm/cosmwasm/blob/v0.12.0/packages/std/src/results/cosmos_msg.rs#L10-L23
+ * */
```
webmaster128, line 63:
```suggestion
 */
```

webmaster128 (approved):

ethanfrey (commented):
Nice one

PR #556, opened 2020-11-26
Add support for CW1 contracts

### LINKED ISSUE #501: Add subkey support

willclarktech:
Closes #501

A few things to note:
1. I didn't add support for the freeze handle msg. Do we want it?
1. I made setting an admin list private and exposed only add/remove as in the issue, but maybe we just want the ability to directly set admins instead?


packages/cosmwasm/src/cw1cosmwasmclient.ts
```diff
@@ -0,0 +1,52 @@
...
+        sender: address,
+        msg: msg,
+      },
+    });
+  }
+
+  public executeSubkey(
+    msgs: readonly CosmosMsg[],
+    memo = "",
+    transferAmount: readonly Coin[] = [],
```
webmaster128, line 43: I'd remove this for simplicity since I don't see any use of `MessageInfo::sent_funds` in the contract. So this is the same as executing and sending to the contract in two operations.

packages/cosmwasm/src/cw1cosmwasmclient.ts
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { Account, BroadcastMode, Coin, GasLimits, GasPrice, OfflineSigner } from "@cosmjs/launchpad";
+
+import { CosmosMsg } from "./cosmosmsg";
+import { CosmWasmFeeTable, ExecuteResult, SigningCosmWasmClient } from "./signingcosmwasmclient";
+
+export interface CanSendResult {
+  readonly can_send: boolean;
+}
```
webmaster128, line 9: We could internalize this response type and just return a boolean: `  public canSend(msg: CosmosMsg, address = this.senderAddress): Promise<boolean> {`

webmaster128:
The current diff only covers the "Ideal User flow (wallet) - submission. This should work for any cw1-compliant contract:" part of #501. I think we at least need the admin part "Ideal User Flow (wallet) - administration This is only for cw1-subkeys." as well.

Once we have those two building blocks, we can close the ticket and pull out all the integration into applications and higher level interfaces.

packages/cosmwasm/src/cw1subkeycosmwasmclient.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+        launchpad.endpoint,
+        alice.address0,
+        wallet,
+        deployedCw1.instances[0],
+      );
+      const spender = makeRandomAddress();
+      expect(await client.getAllowance(spender)).toEqual({ balance: [], expires: { never: {} } });
+
+      const increaseAmount = coin(100, "ucosm");
+      const increaseResult = await client.increaseAllowance(spender, increaseAmount);
```
ethanfrey, line 236: :+1: finally seeing the non-empty/admin results

packages/cosmwasm/src/cw1subkeycosmwasmclient.ts
```diff
@@ -0,0 +1,123 @@
...
+    return admins;
+  }
+
+  public async isAdmin(address = this.senderAddress): Promise<boolean> {
+    const admins = await this.getAdmins();
+    return admins.includes(address);
+  }
+
+  public async getAllAllowances(): Promise<readonly AllowanceResult[]> {
+    const { allowances } = await this.queryContractSmart(this.cw1ContractAddress, {
```
ethanfrey, line 41: All allowances supports pagination: https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw1-subkeys/src/msg.rs#L70-L73

This will only ever get the first page (or default length 10).

Please expose pagination in the API, or internally loop to read all allowances in multiple queries before returning.

ethanfrey (approved):
Good to merge as is.
But missing pagination support for all_allowances and all_permissions queries (this can be added in a separate PR... and only affects contracts with lots of subkey allowances)

PR #564, opened 2020-12-01
Use signed data in stargate client signAndBroadcast

### LINKED ISSUE #561: Need to handle the new sign doc of the `OfflineDirectSigner`
https://github.com/chainapsis/keplr-extension/issues/43
https://github.com/chainapsis/keplr-example/blob/stargate/src/main.js

I am implementing the protobuf tx support for Keplr extension.
For now, signing generally does well.
However, because the current @cosmjs/stargate ignores the `SignDoc` that is given back by the `OfflineSigner`, it is impossible to change the tx config such as the fee, gas from the wallet side.

https://github.com/cosmos/cosmjs/blob/master/packages/stargate/src/signingstargateclient.ts#L126

willclarktech:
Closes #561

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - @cosmjs/launchpad-ledger: `LedgerSigner.sign` method renamed `signAmino`.
 - @cosmjs/proto-signing: Add new package for handling transaction signing with
   protobuf encoding.
+- @cosmjs/proto-signing: Expose `DirectSignResponse` interface.
 - @cosmjs/stargate: Add new package for Cosmos SDK Stargate support.
 - @cosmjs/tendermint-rpc: Make `Client.detectVersion` private and let it return
   a version instead of a client.
--- a/packages/stargate/src/encoding.ts
+++ b/packages/stargate/src/encoding.ts
@@ -1,27 +1,36 @@
+const typeRegister: Record<string, string> = {
+  "/cosmos.bank.v1beta1.MsgSend": "cosmos-sdk/MsgSend",
+  "/cosmos.bank.v1beta1.MsgMultiSend": "cosmos-sdk/MsgMultiSend",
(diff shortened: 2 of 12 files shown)


packages/stargate/src/encoding.ts
```diff
@@ -1,27 +1,36 @@
...
   }
   return type;
 }
+
+export function getMsgTypeUrl(type: string): string {
+  const typeUrlEntry = Object.entries(typeRegister).find(([_, value]) => value === type);
+  if (!typeUrlEntry) {
+    throw new Error("Type not known");
+  }
+  return typeUrlEntry[0];
```
webmaster128, line 35: Just an idea to help explaining what is going on here:

```suggestion
  const [typeUrl, _type] = typeUrlEntry;
  return typeUrl;
```

webmaster128 (commented):
Any chance we can test this new functionality? Do we have a signer at hand that can change the signed content?

willclarktech:
> Any chance we can test this new functionality? Do we have a signer at hand that can change the signed content?

@webmaster128 All of our signers just pass the sign doc straight through to the return value. I could add a class that wraps a signer but modifies the fees?

webmaster128:
> @webmaster128 All of our signers just pass the sign doc straight through to the return value. I could add a class that wraps a signer but modifies the fees?

Yeah sounds good. Something that lives in tests only.

CHANGELOG.md
```diff
@@ -11,7 +11,8 @@
 - @cosmjs/launchpad: `Secp256k1HdWallet.sign` method renamed `signAmino`.
 - @cosmjs/launchpad-ledger: `LedgerSigner.sign` method renamed `signAmino`.
 - @cosmjs/proto-signing: Add new package for handling transaction signing with
-  protobuf encoding.
+- @cosmjs/proto-signing: Expose `DirectSignResponse` interface. protobuf
+  encoding.
```
webmaster128, line 15: This looks broken

webmaster128 (approved):

PR #566, opened 2020-12-08
Add docs to HdPath

webmaster128:


packages/crypto/src/slip10.ts
```diff
@@ -49,6 +49,32 @@ export class Slip10RawIndex extends Uint32 {
   }
 }
 
+/**
+ * An array of raw SLIP10 indices.
+ *
+ * This can be construced by string parsing:
```
willclarktech, line 55: Suggest: "constructed via string parsing"

willclarktech (approved):

Issue #491, opened 2020-10-28
Add support for x/wasm messages in stargate


ethanfrey:
Requires: tagged version of wasmd (https://github.com/CosmWasm/wasmd/pull/293)

Helpful: Test against `simapp` `v0.40.0-rc1` first (https://github.com/cosmos/cosmjs/issues/475)

Let's provide the same support as we have in launchpad. This may switch cosmwasm package to only support starport (wasmd 0.12) or it may add a second cosmwasm package with starport (wasmd 0.12) support while maintaining the launchpad version.

webmaster128:
Let's create a private `@cosmjs/cosmwasm-stargate` package that we develop to the point where we are happy with it. Then we can rename `cosmwasm` -> `cosmwasm-launchpad` and `cosmwasm-stargate` to `cosmwasm` and decide if and how long we want to maintain the old one.

ethanfrey:
cosmwasm/wasmd:v0.12.0-alpha1` is now published and available with `docker pull` .

This should be drop-in compatible with `confio/simapp:v0.40.0-rc1` for bank/staking/ibc modules (please test this if we need to bugfix on the wasmd side). It also exposes a number of new protobuf message types and grpc queries for `x/wasm`. 

Adding these would not only be great for actually providing us dApp clients on 0.12, but also to help define a workflow that allows us to add non-simapp (cosmos-sdk repo) messages and queries for a client.

webmaster128:
#515 should be done first

webmaster128:
https://github.com/cosmos/cosmjs/issues/524 should be done first

willclarktech:
To-do list:

- [x] Prepare `cosmwasm` package
- [x] Prepare `stargate` package
- [x] Prepare `wasmd` scripts
- [ ] Add basic `cosmwasm-stargate` package eg `CosmWasmClient` and `SigningCosmWasmClient` #571 
- [ ] Add full `searchTx` functionality and tests
- [ ] Add verified queries?

webmaster128:
What about this:
1. Rename @cosmjs/cosmwasm to @cosmjs/cosmwasm-launchpad
1. Create a new wrapper @cosmjs/cosmwasm around @cosmjs/cosmwasm-launchpad, which re-exports everything
1. Publish @cosmjs/cosmwasm with a deprecation warning "This package has been renamed to @cosmjs/cosmwasm-launchpad"
1. Wait a release or two
[...]

PR #568, opened 2020-12-10
Prepare Stargate package for CosmWasm support

### LINKED ISSUE #376: Create a log parser for Stargate client
I.e. for parsing raw logs in `.broadcastTx()` results.

willclarktech:
Part 1 of #491
Closes #376

### DIFF
--- a/packages/stargate/src/queries/queryclient.ts
+++ b/packages/stargate/src/queries/queryclient.ts
@@ -3,7 +3,7 @@ import { iavlSpec, ics23, tendermintSpec, verifyExistence, verifyNonExistence } 
 import { toAscii, toHex } from "@cosmjs/encoding";
 import { firstEvent } from "@cosmjs/stream";
 import { Client as TendermintClient, Header, NewBlockHeaderEvent, ProofOp } from "@cosmjs/tendermint-rpc";
-import { arrayContentEquals, assert, isNonNullObject, sleep } from "@cosmjs/utils";
+import { arrayContentEquals, assert, assertDefined, isNonNullObject, sleep } from "@cosmjs/utils";
 import { Stream } from "xstream";
 
 type QueryExtensionSetup<P> = (base: QueryClient) => P;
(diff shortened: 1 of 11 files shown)


packages/stargate/scripts/define-proto.sh
```diff
@@ -1,13 +1,12 @@
...
 set -o errexit -o nounset -o pipefail
-command -v shellcheck > /dev/null && shellcheck "$0"
+command -v shellcheck >/dev/null && shellcheck "$0"
 
 TMP_DIR="./tmp"
 JS_SOURCE_FILE="$TMP_DIR/codecimpl.js"
 DEFINITIONS_FILE="$TMP_DIR/codecimpl.d.ts"
 OUTPUT_DIR="./src/codec/generated/"
 
-
```
webmaster128, line 10: Whiy this diff? This is already done in https://github.com/cosmos/cosmjs/pull/567

packages/stargate/src/queries/queryclient.ts
```diff
@@ -234,7 +234,7 @@ export class QueryClient {
   // throws an error if height is 0 or undefined
   private async getNextHeader(height?: number): Promise<Header> {
     assert(height);
-    if (height == 0) {
+    if (height === 0) {
```
webmaster128, line 237: The `assert` already guarantees this is not 0. We could use `assertDefined(height);` instead to keep the specific error message for 0.

packages/stargate/src/stargateclient.ts
```diff
@@ -54,14 +60,14 @@ export interface BroadcastTxFailure {
   readonly code: number;
   readonly transactionHash: string;
   readonly rawLog?: string;
-  readonly data?: Uint8Array;
+  readonly data?: readonly IMsgData[];
```
webmaster128, line 63: Ideally we has this, such that there is always one entry per message. But I see that the protobuf type system makes this hard to get.

```suggestion
  readonly data: readonly IMsgData[];
```

webmaster128 (commented):
Does this close https://github.com/cosmos/cosmjs/issues/376?

packages/stargate/scripts/define-proto.sh
willclarktech: Ah, need to rebase

webmaster128 (approved):

willclarktech:
> Does this close #376?

I think so!

PR #571, opened 2020-12-10
Add cosmwasm-stargate package

willclarktech:
Part 4 of #491

Merge after:
- [x] #568
- [x] #569
- [x] #570

PR #575, opened 2020-12-11
Deploy more contracts in Wasmd

willclarktech:
Part 6 of #491

~Merge after #571~

### DIFF
--- a/scripts/wasmd/deploy_cw3.js
+++ b/scripts/wasmd/deploy_cw3.js
@@ -1,0 +1,108 @@
+#!/usr/bin/env node
+
+/* eslint-disable @typescript-eslint/naming-convention */
+const { DirectSecp256k1HdWallet } = require("@cosmjs/proto-signing");
+const { SigningCosmWasmClient } = require("@cosmjs/cosmwasm-stargate");
+const fs = require("fs");
+
+const endpoint = "http://localhost:26659";
+const alice = {
+  mnemonic: "enlist hip relief stomach skate base shallow young switch frequent cry park",
+  address0: "wasm14qemq0vw6y3gc3u3e0aty2e764u4gs5lndxgyk",
+  address1: "wasm1hhg2rlu9jscacku2wwckws7932qqqu8xm5ca8y",
+  address2: "wasm1xv9tklw7d82sezh9haa573wufgy59vmwnxhnsl",
+  address3: "wasm17yg9mssjenmc3jkqth6ulcwj9cxujrxxg9nmzk",
+  address4: "wasm1f7j7ryulwjfe9ljplvhtcaxa6wqgula3nh873j",
+};
+
+const codeMeta = {
+  source: "https://crates.io/api/v1/crates/cw3-fixed-multisig/0.3.1/download",
+  builder: "cosmwasm/rust-optimizer:0.10.4",
+};
+
+const initData = [
+  {
+    admin: alice.address0,
+    label: "Multisig (1/3)",
+    initMsg: {
+      voters: [
+        { addr: alice.address0, weight: 1 },
+        { addr: alice.address1, weight: 1 },
+        { addr: alice.address2, weight: 1 },
+      ],
+      required_weight: 1,
+      max_voting_period: { height: 12345 },
+    },
+  },
+  {
+    admin: alice.address0,
+    label: "Multisig (2/3)",
+    initMsg: {
+      voters: [
+        { addr: alice.address0, weight: 1 },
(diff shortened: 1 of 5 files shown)


scripts/wasmd/deploy_cw3.js
```diff
@@ -0,0 +1,108 @@
...
+  mnemonic: "enlist hip relief stomach skate base shallow young switch frequent cry park",
+  address0: "wasm14qemq0vw6y3gc3u3e0aty2e764u4gs5lndxgyk",
+  address1: "wasm1hhg2rlu9jscacku2wwckws7932qqqu8xm5ca8y",
+  address2: "wasm1xv9tklw7d82sezh9haa573wufgy59vmwnxhnsl",
+  address3: "wasm17yg9mssjenmc3jkqth6ulcwj9cxujrxxg9nmzk",
+  address4: "wasm1f7j7ryulwjfe9ljplvhtcaxa6wqgula3nh873j",
+};
+
+const codeMeta = {
+  source: "https://crates.io/api/v1/crates/cw3-fixed-multisig/0.3.1/download",
```
ethanfrey, line 19: Nice. There is a v0.4.0 coming out in the next week or two with plenty of changes to multisigs and groups.

I think minimal breaking API changes, but a lot more power.

scripts/wasmd/deploy_erc20.js
```diff
@@ -0,0 +1,168 @@
...
+};
+const unused = {
+  address: "wasm1cjsxept9rkggzxztslae9ndgpdyt240842kpxh",
+};
+const guest = {
+  address: "wasm17d0jcz59jf68g52vq38tuuncmwwjk42us8fnse",
+};
+
+const codeMeta = {
+  source: "https://crates.io/api/v1/crates/cw-erc20/0.7.0/download",
```
ethanfrey, line 25: I know we have been testing this for a while, but it is not actively developed or used. I would consider using cw20 in cosmwasm-plus instead.

The main bonus I see for erc20 is it uses a binary state encoding, which would test raw query support for binary data (if you do those tests). Otherwise, I don't see a good reason not to switch.

ethanfrey (commented):
Nice

scripts/wasmd/deploy_erc20.js
webmaster128: The problem we have here is dependencies. If we use cw20 we had to upgrade the eintire cosmwasm-plus to a new CosmWasm version before we can test the CosmJS client for that version. erc20 is already a problem, but a small one as we can update the single contract quickly and independent of the rest of the examples repo.

ethanfrey: Valid point, but the above cw3 is also using cosmwasm-plus. (So if you use one, might as well use them all).

Plus, we should be able to update faster now that the cosmwasm api is stabilizing more :pray:

webmaster128: > but the above cw3 is also using cosmwasm-plus.

Jupp, which shows that the multisig client does not belong in CosmJS but a higher level place.

In an ideal world, all tests in this repo only depend on [CosmWasm development contracts](https://github.com/CosmWasm/cosmwasm/tree/master/contracts) that are available automatically once CosmWasm is tagged.

ethanfrey: > Jupp, which shows that the multisig client does not belong in CosmJS but a higher level place.

There will be an increasing amount of app-specific logic we will want to support, we should find a good place for it.

You suggest a separate repo entirely? Where `stargate-cosmwasm` just supports generic `x/wasm` and not any other extension we have built?

I am fine with that, but we do want these helpers, so good to think about splitting this out soonish (maybe after the 0.24 release if it adds too much overhead?)

webmaster128: I don't think there is an urgent need to do such a change. I just see separating out components cleanly as the direction we should work towards instead of blocking ourselves because everything depends on each other.

I think we should be able to ship in this order
1. CosmWasm
2. wasmvm
3. x/wasm
4. CosmJS
5. all contracts and dapps in any order

We often found x/wasm or wasmd bugs when working on the client. And there are CosmJS users that do not care about the apps we build. So I think 3./4. should not need to wait for application specifics.

ethanfrey: Agreed. 3+4 are very important to work in parallel as there are many code segments that need a solid client to test properly

webmaster128 (approved):

PR #587, opened 2020-12-20
Remove tagged string types

### LINKED ISSUE #580: Remove tagged string types
Eg from `tendermint-rpc`:
```
export type Base64String = string & As<"base64">;
export type HexString = string & As<"hex">;
export type IntegerString = string & As<"integer">;
```
This doesn't seem to be considered best practice in TypeScript.

webmaster128:
Closes #580

### DIFF
--- a/packages/tendermint-rpc/src/adaptors/v0-33/hasher.ts
+++ b/packages/tendermint-rpc/src/adaptors/v0-33/hasher.ts
@@ -1,4 +1,4 @@
-import { Sha256 } from "@cosmjs/crypto";
+import { Sha256, sha256 } from "@cosmjs/crypto";
 
 import {
   encodeBlockId,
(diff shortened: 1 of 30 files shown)


willclarktech (approved):

packages/tendermint-rpc/src/adaptors/v0-33/hasher.ts
```diff
@@ -1,4 +1,4 @@
-import { Sha256 } from "@cosmjs/crypto";
+import { Sha256, sha256 } from "@cosmjs/crypto";
```
willclarktech, line 1: Not strictly part of this issue, but since we're tidying things up, maybe you want to refactor the other place where the `Sha256` class is used.

webmaster128: The other use cases in this file require incremental hashing, which is why `Sha256` is still imported here

PR #593, opened 2020-12-22
Actually use AminoTypes custom type URLs

willclarktech:


webmaster128 (commented):
Could you also test overrides, not only additions? Thanks!

packages/stargate/src/aminotypes.ts
```diff
@@ -27,11 +27,16 @@ export class AminoTypes {
   private readonly register: Record<string, string>;
 
   public constructor(additions: Record<string, string> = {}) {
-    this.register = { ...defaultTypes, ...additions };
+    const additionalAminoTypes = Object.values(additions);
+    const filteredDefaultTypes = Object.entries(defaultTypes).reduce(
+      (acc, [key, value]) => (additionalAminoTypes.includes(value) ? acc : { ...acc, [key]: value }),
+      {},
+    );
```
webmaster128, line 34: Oh, why is this needed?

willclarktech: Otherwise, it might be many-to-one and `.find` will find the wrong one.

willclarktech: (I.e. the third test for `fromAmino` will fail.

webmaster128: Great, thanks

webmaster128 (approved):

PR #606, opened 2021-01-08
Update wasmd docker image to v0.14.0-rc2

### LINKED ISSUE #602: Update wasmd to 0.14.0
- [x] Update wasmd to 0.14.0-rc2
- [ ] Update wasmd to 0.14.0 (not yet released, see https://github.com/CosmWasm/wasmd/tags)
- [ ] Update code (if needed)

ethanfrey:
Closes #602 

This passed tests locally, with no code changes needed from 0.13. Let's see what the CI says

### DIFF
--- a/packages/cosmwasm-stargate/src/queries/wasm.spec.ts
+++ b/packages/cosmwasm-stargate/src/queries/wasm.spec.ts
@@ -387,8 +387,12 @@ describe("WasmExtension", () => {
         expect(codeId).toBeGreaterThanOrEqual(1);
         expect(codeId).toBeLessThanOrEqual(200);
         expect(result.data!.length).toEqual(1);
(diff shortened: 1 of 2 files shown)


ethanfrey:
wasmd `v0.14.0-rc2` changed the result.data field returns from the `x/wasm` module.

Higher level functions are using reported events to get data, rather than parsing the data field. Only one test fails that explicitly tests the data field (and init.sh works fine).

I added references to the new protobuf types that can be used to parse the data field. @willclarktech or @webmaster128 could you take this over and get the last test to pass? I provide links to the protobuf definitions used here, but you know the tooling to support, so I'll leave it to you.

The `v0.14.0` release should be the same code, but adding a few unit tests and a doc fix.

PR #608, opened 2021-01-08
Bump simapp to v0.40.0-rc7

ethanfrey:
Closes #607 

This doesn't work as the genesis file has changed - a number of IBC changes that need to be adapted.

PR #611, opened 2021-01-11
Upgrade simapp to 0.40.0

### LINKED ISSUE #607: Update simapp to v0.40.0-rc7 (or v0.40.0 final)
Following up on #602 to make sure we are testing support for final released.

I updated the image and the startup failed due to different genesis format for IBC data. I guess @alpe will need to help here to re-generate a proper genesis file with the IBC data.

webmaster128:
Closes #607
Closes #608
Closes #610


packages/stargate/src/queries/ibc.spec.ts
```diff
@@ -125,12 +130,8 @@ describe("IbcExtension", () => {
           1,
           2,
           3,
-          4,
-          5,
-          6,
-          7,
         ]);
-        expect(response.sequences).toEqual([4, 5, 6, 7].map((n) => Long.fromInt(n, true)));
+        expect(response.sequences).toEqual([1, 2, 3].map((n) => Long.fromInt(n, true)));
```
webmaster128, line 134: This test got bording. We "unreceived" all of the requested packages

ethanfrey: bording? broken?

I will look in more. If it is an error in the relayer/genesis, or in our code, or in the Cosmos SDK :grimacing:

webmaster128: "Boring", sorry.

ethanfrey (commented):
Generally looks good to me and happy to see IBC working so quickly.

It would be nice to have a little more data to ensure unreceived/acknowledged queries work properly.

webmaster128:
Merging now since we only have this set of test data

PR #613, opened 2021-01-12
Some API polishing

webmaster128:


packages/cosmwasm-stargate/src/signingcosmwasmclient.spec.ts
```diff
@@ -25,19 +25,19 @@ const { MsgDelegate } = codec.cosmos.staking.v1beta1;
 const { Tx } = codec.cosmos.tx.v1beta1;
 
 describe("SigningCosmWasmClient", () => {
-  describe("connectWithWallet", () => {
+  describe("connectWithSigner", () => {
     it("can be constructed", async () => {
       pendingWithoutWasmd();
       const wallet = await DirectSecp256k1HdWallet.fromMnemonic(alice.mnemonic, undefined, wasmd.prefix);
```
willclarktech, line 31: Do we still like `Wallet` in these class names?

willclarktech (approved):
Looking good. I guess the Launchpad change needs a CHANGELOG entry?

packages/cosmwasm-stargate/src/signingcosmwasmclient.spec.ts
webmaster128: The idea is this: signer is the interface. There are many ways to implement the signer. One way is having a wallet (term used for key storages here).

packages/cosmwasm-launchpad/src/signingcosmwasmclient.ts
```diff
@@ -149,7 +149,7 @@ export interface PrivateSigningCosmWasmClient {
 }
 
 export class SigningCosmWasmClient extends CosmWasmClient {
-  public readonly senderAddress: string;
+  public readonly signerAddress: string;
```
webmaster128, line 152: @willclarktech ah right, I did not notice this is public. Will add a CHANGELOG entry for this change.

Issue #594, opened 2020-12-23
SigningCosmWasmClient: test Amino signing mode with CosmWasm message types


webmaster128:
Right now, the CosmWasm types are missing in

```ts
export class SigningCosmWasmClient extends CosmWasmClient {
  // …
  private readonly aminoTypes = new AminoTypes();
```

ethanfrey:
I think we can reopen this and try again with the wasmd 0.14.0 release, that should have legacy amino support working. It would be nice to have ledger working if it isn't a huge ask, and I am guessing that the code that failed tests before may work with the new wasmd.

webmaster128:
> I am guessing that the code that failed tests before may work with the new wasmd

I don't see any point in untilizing any kind of auto-conversion code as we did before. If we want to have this, let's switch over message types and add support for some of them by creating the Amino representation manually.

Let's say we want to create a sign for of a `/cosmwasm.wasm.v1beta1.MsgExecuteContract` – how do we know what it's Amino JSON represenation looks like?

ethanfrey:
I agree completely with manually adding these.

I would support bank, staking (most common used by wallets) and x/wasm (for cosmwasm testnet)

That should be sufficient bridge until ledger supports protobuf/Stargate.

PR #614, opened 2021-01-12
Implement Amino signing for bank/staking in Stargate

willclarktech:
Part 1 of #594

Two questions raised here:
- What should be in the default registry exported from `@cosmjs/proto-signing`. Should `MsgSend` be in there? If so should all the others from `bank`/`staking` also be there?
- ~~How to handle PubKeys that aren't Secp256k1 (see https://github.com/cosmos/cosmjs/pull/614/files#diff-0beea075ac1216ee252587198c2533fe8dc426da803a316d6536df72b7fbb628R218-R222)? Amino puts them in this weird Bech32-ish format but that doesn't include the scheme.~~

### DIFF
--- a/packages/stargate/src/aminotypes.ts
+++ b/packages/stargate/src/aminotypes.ts
@@ -1,61 +1,386 @@
-const defaultTypes: Record<string, string> = {
-  "/cosmos.bank.v1beta1.MsgSend": "cosmos-sdk/MsgSend",
-  "/cosmos.bank.v1beta1.MsgMultiSend": "cosmos-sdk/MsgMultiSend",
-  "/cosmos.crisis.v1beta1.MsgVerifyInvariant": "cosmos-sdk/MsgVerifyInvariant",
-  "/cosmos.distribution.v1beta1.MsgSetWithdrawAddress": "cosmos-sdk/MsgSetWithdrawAddress",
-  "/cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward": "cosmos-sdk/MsgWithdrawDelegatorReward",
-  "/cosmos.distribution.v1beta1.MsgWithdrawValidatorComission": "cosmos-sdk/MsgWithdrawValidatorComission",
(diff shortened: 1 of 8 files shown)


packages/stargate/src/aminotypes.ts
```diff
@@ -1,61 +1,381 @@
...
+          maxRate: commission.max_rate,
+          maxChangeRate: commission.max_change_rate,
+        },
+        minSelfDelegation: min_self_delegation,
+        delegatorAddress: delegator_address,
+        validatorAddress: validator_address,
+        pubkey: {
+          // TODO: How do we detect/handle other kinds of PubKey?
+          type_url: "/cosmos.crypto.secp256k1.PubKey",
+          value: fromBase64(decodeBech32Pubkey(pubkey).value),
```
webmaster128, line 221: The pubkey type is encoded in `pubkey` and available in `ecodeBech32Pubkey(pubkey).type`. I'd switch over that and throw an error for everything that is not `tendermint/PubKeySecp256k1`. This makes it safe and easy to extend if needed.

webmaster128 (commented):
Beautiful 😍

webmaster128:
>  What should be in the default registry exported from `@cosmjs/proto-signing`. Should `MsgSend` be in there?

Is this question related to Amino signing (and the class `AminoTypes`)? As far as I can tell, `@cosmjs/proto-signing` does not know about the existence of Amino signing.

willclarktech:
> Is this question related to Amino signing (and the class AminoTypes)?

Not really, I just thought about it because in the PR I added a bunch of Msg types to the default registry used by `SigningStargateClient` and I wasn't sure if the more sensible place would be to include it in all registries by default.

webmaster128:
> I wasn't sure if the more sensible place would be to include them in all registries by default

I'm not sure as well, but the current setup seems good enough for me. We'll learn more when people start using it.

webmaster128 (approved):

PR #616, opened 2021-01-14
Investigate ts-proto

### LINKED ISSUE #586: Evaluate ts-proto for code generation
After a talk with @okwme and @amaurymartiny it turned out that [ts-proto](https://github.com/stephenh/ts-proto) might be a better, more modern way to generate TypeScript definitions from .proto files than protobuf.js's .d.ts generator is. ts-proto is based on protobuf.js but provides a TS layer that produces more modern output.

Regen uses ts-proto in their client lib here: https://github.com/regen-network/regen-js

It would be nice if we find a common best practice for JS/TS code generation to improve the experience for developers working with CosmJS but have their own types.

willclarktech:
Closes #586 

~~Spike relating to #586, seemed easier to get a concrete idea of what would be involved in switching to ts-proto by just making some code changes.~~

Some things to say:
- It doesn't seem possible to generate types except via `.proto` files, so we'd be giving up the decorator approach. We also can't use the more manual approach from `demo.ts`.
- It doesn't generate any index files, so we'd either have to do that ourselves or stick to the nested imports pattern.
- Default values seem to be handled differently to what we're used to, so one of the tests in `signing.spec.ts` is failing.
- It seems ts-proto doesn't support Node <v12: https://github.com/stephenh/ts-proto/issues/97
- I definitely prefer the ts-proto approach. It feels much less magical and more straightforward. The generated code is much easier to read.
- ts-proto is 2 years old and still has the feel of a young project. I'm a little nervous about that.


packages/proto-signing/src/pubkey.ts
```diff
@@ -1,38 +1,37 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import { fromBase64 } from "@cosmjs/encoding";
-import { encodeSecp256k1Pubkey, PubKey } from "@cosmjs/launchpad";
+import { encodeSecp256k1Pubkey, PubKey as LaunchpadPubKey } from "@cosmjs/launchpad";
```
webmaster128, line 3: Urg, here it is annoying.

packages/proto-signing/src/pubkey.ts
```diff
@@ -1,38 +1,37 @@
...
 
-import { cosmos, google } from "./codec";
+import { PubKey } from "./codec/cosmos/crypto/secp256k1/keys";
+import { Any } from "./codec/google/protobuf/any";
```
webmaster128, line 6: This seems as good as the old version. I know there was a linter rule in tslint which discouraged the use of sub-folder imports. But in other places this is recommended to import a subset of a module.

packages/proto-signing/src/registry.spec.ts
```diff
@@ -51,34 +49,35 @@ describe("registry demo", () => {
     expect(msgSendDecoded.amount).toEqual(msgSend.amount);
   });
 
-  it("works with a custom msg", () => {
-    const typeUrl = "/demo.MsgDemo";
-    const registry = new Registry([[typeUrl, MsgDemoType]]);
-    const MsgDemo = registry.lookupType(typeUrl)!;
+  // TODO: Can't autogenerate types from TS code using ts-proto
```
webmaster128, line 52: `MsgDemoType` is created at runtime using `protobuf` only:

```ts
import { Field, Root, Type } from "protobufjs";

export const MsgDemo = new Type("MsgDemo").add(new Field("example", 1, "string"));

const root = new Root().define("demo").add(MsgDemo);

export default root;
```

This is used by the Starport team: https://twitter.com/fadeev/status/1345006725586440192

Shouldn't it be possible to use them the same way as before via one of those constructors?

```ts
/** Reflected message type. */
export class Type extends NamespaceBase {
    // …

    /**
     * Creates a message type from a message type descriptor.
     * @param name Message name
     * @param json Message type descriptor
     * @returns Created message type
     */
    public static fromJSON(name: string, json: IType): Type;

    // …

    /**
     * Creates a new message of this type using the specified properties.
     * @param [properties] Properties to set
     * @returns Message instance
     */
    public create(properties?: { [k: string]: any }): Message<{}>;

    // …
}
```

webmaster128:
One additional positive note:
- The generated code does not uses classes. This probably makes our life easier when it compes to equality testing. No `toObject` calls needed anymore

packages/proto-signing/src/codec/cosmos/bank/v1beta1/tx.ts
```diff
@@ -0,0 +1,324 @@
...
+  },
+  decode(input: Uint8Array | Reader, length?: number): MsgSend {
+    const reader = input instanceof Uint8Array ? new Reader(input) : input;
+    let end = length === undefined ? reader.len : reader.pos + length;
+    const message = { ...baseMsgSend } as MsgSend;
+    message.amount = [];
+    while (reader.pos < end) {
+      const tag = reader.uint32();
+      switch (tag >>> 3) {
+        case 1:
```
ethanfrey, line 102: Interesting, it checks the field number, but doesn't check if the type matches. (last 3 bytes)
So, if I encoded an int with field 1, this would not detect the mis-match.

(Or does this happen magically in `reader.string()`? I assume these calls consume bytes from the reader)

packages/proto-signing/src/codec/cosmos/bank/v1beta1/bank.ts
```diff
@@ -0,0 +1,650 @@
+/* eslint-disable */
```
amaurymartiny, line 1: (comment not related to the actual line, but to Will's original comment, I think having threads make things clearer).

> It doesn't seem possible to generate types except via .proto files, so we'd be giving up the decorator approach. We also can't use the more manual approach from demo.ts.

This was what I was initially proposing, and what I'm exploring inside regen-js. Happy to discuss more about it during our next call.

packages/proto-signing/src/codec/cosmos/bank/v1beta1/tx.ts
```diff
@@ -0,0 +1,324 @@
...
+          message.fromAddress = reader.string();
+          break;
+        case 2:
+          message.toAddress = reader.string();
+          break;
+        case 3:
+          message.amount.push(Coin.decode(reader, reader.uint32()));
+          break;
+        default:
+          reader.skipType(tag & 7);
```
ethanfrey, line 112: Okay, here if actually uses the field info (`tag & 7`) to determine how to skip the unknown. Odd this is not enforced above. eg. 

```ts
case 2:
   if tag & 7 != 2 {
       throw new Error(`Unexpected type ${tag & 7} for field 1`);
    }
    message.fromAddress = reader.string();
    break;
```

This is due to `proto-ts`, but seems like it could use improvement there
```

File: packages/proto-signing/src/codec/cosmos/bank/v1beta1/bank.ts [hunk 7]
```diff
@@ -0,0 +1,650 @@
...
+import { Coin } from "../../../cosmos/base/v1beta1/coin";
```
amaurymartiny, line 2: > It doesn't generate any index files, so we'd either have to do that ourselves or stick to the nested imports pattern.

Yeah, I also wanted that, I added a bunch of `index.ts` manually, until I stumbled upon https://github.com/stephenh/ts-proto/issues/167#issuecomment-737936654. I proposed a solution in the thread. I believe generating `index.ts` is a desirable feature.

packages/proto-signing/src/codec/cosmos/bank/v1beta1/tx.ts
ethanfrey: Otherwise, the auto-gen code does seem more compact and less confusing. But maybe just cuz it misses some options/cases (like deciding whether to error on unknown fields or silently skip them)

ethanfrey (commented):
A few thoughts looking at the auto-gen code

packages/proto-signing/src/codec/cosmos/bank/v1beta1/bank.ts
```diff
@@ -0,0 +1,650 @@
...
+import { Writer, Reader } from "protobufjs/minimal";
```
amaurymartiny, line 3: > ts-proto is 2 years old and still has the feel of a young project. I'm a little nervous about that.

This is definitely something I share. Adding on top of that, ts-proto is heavily dependent on ts-poet (a string manipulation tool for code generation) by the same author. There is a non-zero risk these packages gets unmaintained down the road.

On the plus side, these two packages are not that big.

packages/proto-signing/src/codec/cosmos/bank/v1beta1/tx.ts
```diff
@@ -0,0 +1,324 @@
...
+const baseMsgMultiSendResponse: object = {};
+
+/**
+ *  Msg defines the bank Msg service.
+ */
+export interface Msg {
+  /**
+   *  Send defines a method for sending coins from one account to another account.
+   */
+  Send(request: MsgSend): Promise<MsgSendResponse>;
```
amaurymartiny, line 51: In general, it is my belief we should maintain a fork of ts-proto, so that we have flexibility on the exact output we want.

For example, let's take this auto-generated interface in x/bank (related to ADR-031). It doesn't make sense to me to return a promise here. Also, there should be a better integration with `SigningStargateClient` for tx building.



I started exploring this by forking ts-proto and playing around. A good UX example for me would be:

```ts
const signingClient = new SigningStargateClient(...);
const bankMsgClient = new MsgClientImpl(signingClient); // MsgClientImpl is imported from this file

// signingClient (or a wrapper around it) could hold an internal array of `Msg`s, and the following line would add one `Msg` to the array
bankMsgClient.Send(request1, (response) => { console.log(response); }); // Everything will be well-typed.
// We could add more `Msg`s to the tx, by importing `Msg` clients from other modules
stakingMsgClient.Delegate(request2, (response) => { console.log(response); });

// ...

signingClient.signAndBroadcast(...); // Signs and broadcasts tx with all internal `Msg`s.
```

For those interested, I forked ts-proto [here](https://github.com/amaurymartiny/ts-proto/pull/1/files) and it generated [this code](https://github.com/regen-network/regen-js/pull/14/files#diff-0b3e07c937587ad311e5fd37408afe77dcf3ba661f5d3c24851ee3282d1d191bR76-R82)

webmaster128: > It doesn't make sense to me to return a promise here.

Why? What would be the alternative?

amaurymartiny: Mostly because of multi-Msgs txs. For those txs, this method should add a `x/bank Send Msg` to the tx building client (here `SigningStargateClient`, or a wrapper of it), and that's a sync operation. The tx building client could then do something else, like adding other `Msg`s, or fees/memo.

The signature would look like:

```ts
Send(request: MsgSend, callback?: (response: MsgSendResponse) => void): void;
```

And see my message above for an example usage.

webmaster128: If it is a sync operation, then it is clearly not an implementation of the `service`.

But what you outlined there can easily be done with the Promise interface as well. Promises can take as long as you want until they resolve.

```
const signingClient = new SigningStargateClient(...);
const bankMsgClient = new MsgClientImpl(signingClient); // MsgClientImpl is imported from this file

// signingClient (or a wrapper around it) could hold an internal array of `Msg`s, and the following line would add one `Msg` to the array
const pendingSend = bankMsgClient.Send(request1); // Everything will be well-typed.
// We could add more `Msg`s to the tx, by importing `Msg` clients from other modules
const pendingDelegate = stakingMsgClient.Delegate(request2);

// ...

signingClient.signAndBroadcast(...); // Signs and broadcasts tx with all internal `Msg`s.

// bankMsgClient and stakingMsgClient observe chain and wait for execution

const [response1, response2] = await = Promise.all([pendingSend, pendingDelegate]);
```

amaurymartiny: That would make sense too! It also seemed to me like the obvious way to do it with ts-proto, however, I find it slightly less idiomatic:

```ts
const pendingSend = bankMsgClient.Send(request1);
```

I'm guessing unfamiliar users would think "ok this returns a promise, so I should await", and that would not work. This, + the final `Promise.all()`, is not something I find really UX-friendly, so would prefer the callback version here.

But happy to define this as the go-to UX if it's well-documented.

webmaster128: > so would prefer the callback version here.

Which is also easy to achieve without changing the code generator:

```ts
const signingClient = new SigningStargateClient(...);
const bankMsgClient = new MsgClientImpl(signingClient); // MsgClientImpl is imported from this file

// signingClient (or a wrapper around it) could hold an internal array of `Msg`s, and the following line would add one `Msg` to the array
bankMsgClient.Send(request1).then((response) => { console.log(response); }); // Everything will be well-typed.
// We could add more `Msg`s to the tx, by importing `Msg` clients from other modules
stakingMsgClient.Delegate(request2).then((response) => { console.log(response); });

// ...

signingClient.signAndBroadcast(...); // Signs and broadcasts tx with all internal `Msg`s.
```

ethanfrey: Those callbacks are much more annoying if you want to do anything with them (not console.log).

Let's say you want to use these results in a calculation. How do you get access to both in a callback? It is a nightmare (like async js coding 7 years ago). The Promise.all is one line to learn and then super easy to use the response.

Then again, I am a bit confused by your use-case here. If you show this hooked up to something, then maybe it is clearer

amaurymartiny: Sounds good 👍.

willclarktech:
Updating the `@cosmjs/stargate` package has hit up against an issue in `ts-proto`: https://github.com/stephenh/ts-proto/issues/187

willclarktech:
@webmaster128 This is probably a good point for another review. Things to mention:

- In general a lot of things are simpler. Eg using interfaces vs classes and the query clients.
- There's an issue with the generated import paths for the wasm codec, which I manually fixed [here](https://github.com/cosmos/cosmjs/pull/616/commits/0f1d930b201a95cbcdc9741e36488f4b0bb4108b) and haven't got to the bottom of yet.
- The outstanding issue of whether we want to encourage nested imports, and how to handle the lack of index files if not.
- I adjusted the structure of the `@cosmjs/stargate` package to output the type definitions in the `build` directory to enable the nested imports. They're still checked in, but I'm not sure if we want that, and it causes an error on CI which will need to be fixed.
- The tests which use a custom codec are much more complex, eg https://github.com/cosmos/cosmjs/pull/616/files#diff-5d4996df2348af5d070553055ea6e3b5d991e73d809cb6a125f25d6094ed0954R285

packages/cosmwasm-stargate/src/cosmwasmclient.searchtx.spec.ts
```diff
@@ -8,7 +8,8 @@ import {
   makeSignDoc,
   Registry,
 } from "@cosmjs/proto-signing";
-import { BroadcastTxResponse, codec, isBroadcastTxFailure, isBroadcastTxSuccess } from "@cosmjs/stargate";
+import { BroadcastTxResponse, isBroadcastTxFailure, isBroadcastTxSuccess } from "@cosmjs/stargate";
+import { Tx, TxRaw } from "@cosmjs/stargate/build/codec/cosmos/tx/v1beta1/tx";
```
webmaster128, line 12: Okay, this is really ugly. I thought the `"main": "build/index.js"` would define the base path for `@cosmjs/stargate`, which is why I thought we get `@cosmjs/stargate/codec/cosmos/tx/v1beta1/tx`. But that's not the case.

I wonder if this is still out problem, if we re-export the types in `@cosmjs/stargate` that are part of the package's API. But if I recall correctly this is already too many to be maintainable.

packages/cosmwasm-stargate/src/queries/wasm.ts
```diff
@@ -67,54 +63,80 @@ export interface WasmExtension {
...
+  const queryService = new QueryClientImpl(rpc);
+
   return {
     unverified: {
       wasm: {
         listCodeInfo: async (paginationKey?: Uint8Array) => {
-          const request = paginationKey ? { pagination: { key: paginationKey } } : {};
-          return queryService.codes(request);
+          const request = {
+            pagination: {
```
webmaster128, line 76: When `paginationKey` is unset, we can (and probably should) leave the whole `pagination` value `undefined`

packages/cosmwasm-stargate/types/codec/gogoproto/gogo.d.ts
```diff
@@ -0,0 +1 @@
+export declare const protobufPackage = "gogoproto";
```
webmaster128, line 1: Do we need this package?

packages/proto-signing/package.json
```diff
@@ -54,6 +53,7 @@
   },
   "devDependencies": {
     "@cosmjs/encoding": "^0.24.0-alpha.22",
-    "@cosmjs/utils": "^0.24.0-alpha.22"
+    "@cosmjs/utils": "^0.24.0-alpha.22",
+    "ts-proto": "^1.53.0"
```
webmaster128, line 57: As mentioned above, no need to duplicate the dependency here

packages/stargate/src/queries/index.ts
```diff
@@ -9,3 +9,4 @@ export { BankExtension, setupBankExtension } from "./bank";
 export { DistributionExtension, setupDistributionExtension } from "./distribution";
 export { IbcExtension, setupIbcExtension } from "./ibc";
 export { setupStakingExtension, StakingExtension } from "./staking";
+export { createRpc } from "./utils";
```
webmaster128, line 12: `createPagination` would be nice to have available in other packages as well. We created many of those objects by hand in cosmwasm-stargate.

packages/stargate/src/queries/utils.ts
```diff
@@ -9,11 +12,31 @@ export function toAccAddress(address: string): Uint8Array {
...
+} {
+  return {
+    key: paginationKey ?? new Uint8Array(),
+    offset: Long.fromNumber(0),
+    limit: Long.fromNumber(0),
+    countTotal: false,
+  };
+}
+
+interface Rpc {
```
webmaster128, line 31: Can we make this type public, since it is part of the public `createRpc` signature? Any maybe a little more verbose.

packages/stargate/src/queries/utils.ts
```diff
@@ -9,11 +12,31 @@ export function toAccAddress(address: string): Uint8Array {
...
-export function toObject<I extends object>(thing: I): Omit<I, never> {
-  return { ...thing };
+export function createPagination(
+  paginationKey?: Uint8Array,
+): {
+  readonly key: Uint8Array;
+  readonly offset: Long;
+  readonly limit: Long;
+  readonly countTotal: boolean;
+} {
```
webmaster128, line 22: As mentioned above, this can probably return `undefined` of `paginationKey` is unset

yarn.lock
```diff
@@ -7138,6 +7158,11 @@ prettier-linter-helpers@^1.0.0:
   dependencies:
     fast-diff "^1.1.2"
 
+prettier@^2.0.2:
+  version "2.2.1"
+  resolved "https://registry.yarnpkg.com/prettier/-/prettier-2.2.1.tgz#795a1a78dd52f073da0cd42b21f9c91381923ff5"
+  integrity sha512-PqyhM2yCjg/oKkFPtTGUojv7gnZAoG80ttl45O6x2Ug/rMJw4wcc9k6aaf2hibP7BGVCCM33gZoGjyvt9mm16Q==
+
 prettier@^2.0.5:
```
webmaster128, line 7166: Could you merge those two prettier versions?

webmaster128 (commented):
Regarding the subfolder imports: Even if can can avoid the need for them, I'm getting the impression we should move the types to build consistently for all packages. This allows users to access internal types in case we miss to export them. I think we can discourage that the use of internal types, but often it is a feasable workaround to unblock people when a type is missing. If I understand correctly, this does not work right now since the types and the .js files use a different folder.

packages/stargate/src/queries/utils.ts
willclarktech: Probably! I didn't because of this comment that I don't fully understand: https://github.com/stephenh/ts-proto/blob/v1.53.0/src/generate-services.ts#L262-L270

webmaster128: I wonder what `index.ts`s he's referring to. Maybe there are (optional) index files?

The interface is defined multiple times indeed, but it is always the same one.

packages/cosmwasm-stargate/types/codec/gogoproto/gogo.d.ts
willclarktech: No, and it seems there are a bunch of other packages we don't need. 🤔

packages/stargate/src/queries/utils.ts
willclarktech: > maybe a little more verbose

Any suggestions?

webmaster128: Unfortunately I hardly understand what this is and how it fits in the stack

willclarktech: I think this is the kind of `index.ts` which is meant: https://github.com/stephenh/ts-proto/blob/v1.53.0/integration/barrel-imports/index.ts
This one doesn't involve the generated `Rpc` types specifically, but shows what would go wrong if they were exported from both `foo.ts` and `bar.ts`. I don't believe this index file has been generated.

webmaster128:
I can reproduce the lint error. It happens when the types are not yet available. Then `codeInfo` is of type `any` and the exclamation marks in

```
      const { codeInfo, data } = await client.unverified.wasm.getCode(hackatomCodeId);
      expect(codeInfo!.codeId.toNumber()).toEqual(hackatomCodeId);
      expect(codeInfo!.creator).toEqual(alice.address0);
```

become redundant to the linter. In order to fix this we either need to check-in types or cannot run lint in parallel to build.

webmaster128 (approved):
Great job, thank you!

PR #618, opened 2021-01-14
Demonstrate Stargate Ledger compatibility via Amino signing

willclarktech:
I guess the package name is a bit misleading, it’s really amino-ledger.


webmaster128 (approved):

webmaster128:
> I guess the package name is a bit misleading, it’s really amino-ledger

Feel free to rename at any point in time

PR #622, opened 2021-01-19
Adapt to master -> main renaming

webmaster128:


willclarktech (approved):
Looks good. Will we also need to update references to other repos?

webmaster128:
> Will we also need to update references to other repos?

What kind of references? From out text/comments to other repos? I'd only update those is the target choses to rename their default branch. But even then GitHub implemented some nice redirects (check e.g. https://github.com/cosmos/cosmjs/tree/master/packages/cli), so I don't see any pressure.

willclarktech:
> What kind of references? From out text/comments to other repos?

Yeah, but in particular other repos we control.

webmaster128:
Yeah I think we should update other repos as well.

PR #625, opened 2021-01-20
Upgrade @ledgerhq/* dependencies

### LINKED ISSUE #624: Update launchpad-ledger dependencies
A Chrome update causes a "Unable to reset device" error when using the ledger to log into [dApps](https://github.com/CosmWasm/dApps).

It seems to be fixed in a newer version of hw-transport: https://github.com/polkadot-js/apps/issues/4049

webmaster128:
Closes #624

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -328,68 +328,69 @@
   dependencies:
     vary "^1.1.2"
 
-"@ledgerhq/devices@^5.25.0":
-  version "5.25.0"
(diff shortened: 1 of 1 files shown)


willclarktech (approved):
Looks good. Should I test it manually?

webmaster128:
> Should I test it manually?

We have semver guarantees, right? ;) Given that this does not affect our users, I think we can just assume it works.

Issue #629, opened 2021-01-25
Add support for staking and distribution to stargate


ethanfrey:
Both unverified queries and constructing messages. This should be exposed in the `CosmosSigningClient`.

Use case: creating a dApp that uses the native staking APIs to bond tokens, collect rewards, and unbond.

PR #630, opened 2021-01-26
Add support for staking/distribution queries

willclarktech:
Part 1 of #629. Also adds distribution msgs to default type registry in `SigningStargateClient`.


ethanfrey (approved):
DistributionExtension and StakingExtension APIs look good to me.

I can't speak to the implementation, but it adds all the missing query support. Thanks

webmaster128 (approved):
Nice

PR #631, opened 2021-01-26
Add Amino types for distribution msgs

willclarktech:
Part 2 of #629


packages/stargate/src/aminotypes.spec.ts
```diff
@@ -12,12 +12,22 @@ import {
   MsgSend,
   MsgUndelegate,
 } from "@cosmjs/launchpad";
+import {
+  MsgFundCommunityPool,
+  MsgSetWithdrawAddress,
+  MsgWithdrawDelegatorReward,
+  MsgWithdrawValidatorCommission,
+} from "@cosmjs/launchpad/types/msgs";
```
webmaster128, line 20: Can we import them from `@cosmjs/launchpad` as above?

webmaster128 (approved):

packages/stargate/src/aminotypes.spec.ts
willclarktech: Oops, thanks

Issue #632, opened 2021-01-26
Add functionality needed for relayer


ethanfrey:
A number of missing APIs that I found when sketching out the relayer. Most are available in lower-level, but it would be good to expose them.

Missing from `stargate/SigningStargateClient`:

- [ ] `getCommit` - Return header and all validator signatures. This is available as `this.tmClient.commit(height?: number)` PR #633 but likely not needed
- [x] `queryClient` doesn't include IBCExtension. (in `v0.24.0-alpha.19`)
- [ ] Helpers for creating IBC messages. Minimal set: CreateClient, UpdateClient, ConnOpenInit, ConnOpenTry, ConnOpenAck, ConnOpenConfirm, ChanOpenInit, ChanOpenTry, ChanOpenAck, ChanOpenConfirm, ReceivePacket, ReceiveAcknowledgement, ReceiveTImeout (see https://github.com/cosmos/cosmjs/issues/636)

(I will add more when I find them)

willclarktech:
Default `StargateClient` includes the IBC extension as of https://github.com/cosmos/cosmjs/pull/630

ethanfrey:
Awesome, I will pull in main and double-check

webmaster128:
As far as I can tell those issues have been resolved. Let's create specific follow-ups when needed.

Wehn it comes to accessing lower level Tendermint it makes sense to create a Tendermint34Client directly. As of https://github.com/cosmos/cosmjs/pull/1376 you can stick your manually created TendermintClient in a SigningStargateClient or SigningCosmWasmClient.

PR #633, opened 2021-01-26
Expose tmClient.commit in StargateClient

ethanfrey:
Address one point in #632

### DIFF
--- a/packages/stargate/src/stargateclient.ts
+++ b/packages/stargate/src/stargateclient.ts
@@ -16,6 +16,7 @@ import {
   adaptor34,
   broadcastTxCommitSuccess,
   Client as TendermintClient,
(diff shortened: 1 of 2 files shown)


ethanfrey:
Freezing this one, as I am investigating an alternate approach proposed by Simon.

ethanfrey:
Okay, closing this. We should just create our own TendermintClient and QueryClient if we want advanced usages. The copy inside the SigningStargateClient is for internal use only, and it is not hard to make our own.

webmaster128:
https://github.com/cosmos/cosmjs/issues/642 aims to make creating Tendermint clients easier. I think this will be needed for many semi-advanced use cases and is much more maintainable than going through StargateClient.

PR #641, opened 2021-01-27
Fill out Stargate query extension tests

### LINKED ISSUE #629: Add support for staking and distribution to stargate

willclarktech:
Closes #629

### DIFF
--- a/packages/stargate/src/queries/staking.ts
+++ b/packages/stargate/src/queries/staking.ts
@@ -20,7 +20,11 @@ type IQueryValidatorDelegationsResponse = cosmos.staking.v1beta1.IQueryValidator
 type IQueryValidatorsResponse = cosmos.staking.v1beta1.IQueryValidatorsResponse;
 type IQueryValidatorUnbondingDelegationsResponse = cosmos.staking.v1beta1.IQueryValidatorUnbondingDelegationsResponse;
 
+// This needs to be exported otherwise TS won’t let you export BondStatusString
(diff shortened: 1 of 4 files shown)


packages/stargate/src/queries/staking.ts
```diff
@@ -165,7 +167,7 @@ export function setupStakingExtension(base: QueryClient): StakingExtension {
           });
           return toObject(response);
         },
-        validators: async (status: string, paginationKey?: Uint8Array) => {
+        validators: async (status: keyof typeof BondStatus, paginationKey?: Uint8Array) => {
```
webmaster128, line 170: Nice type magic. So I guess we could also do `type BondStatusString = Exclude<keyof typeof BondStatus, "BOND_STATUS_UNSPECIFIED">`.

I think it would be nice to accept the same argument here and above. "BOND_STATUS_UNSPECIFIED" should never be used.

webmaster128 (approved):

packages/stargate/src/queries/staking.ts
willclarktech: Excluding `BOND_STATUS_UNSPECIFIED` makes sense. I actually thought I'd removed this from here because I couldn't use it in the other place for some obscure type reason I didn't investigate too thoroughly.

webmaster128 (approved):

PR #643, opened 2021-01-27
Add more types to cosmwasm-stargate

willclarktech:


packages/cosmwasm-stargate/src/cosmwasmclient.ts
```diff
@@ -56,7 +62,13 @@ export interface PrivateCosmWasmClient {
...
 export class CosmWasmClient {
   private readonly tmClient: TendermintClient;
-  private readonly queryClient: QueryClient & AuthExtension & BankExtension & WasmExtension;
+  private readonly queryClient: QueryClient &
+    AuthExtension &
+    BankExtension &
+    DistributionExtension &
+    IbcExtension &
+    StakingExtension &
+    WasmExtension;
```
webmaster128, line 71: Since `queryClient` is private and should remain that way, this only adds code but no feature, right?

packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
```diff
@@ -95,6 +101,10 @@ function createBroadcastTxErrorMessage(result: BroadcastTxFailure): string {
 function createDefaultRegistry(): Registry {
   return new Registry([
     ["/cosmos.bank.v1beta1.MsgMultiSend", MsgMultiSend],
+    ["/cosmos.distribution.v1beta1.MsgFundCommunityPool", MsgFundCommunityPool],
+    ["/cosmos.distribution.v1beta1.MsgSetWithdrawAddress", MsgSetWithdrawAddress],
+    ["/cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward", MsgWithdrawDelegatorReward],
+    ["/cosmos.distribution.v1beta1.MsgWithdrawValidatorCommission", MsgWithdrawValidatorCommission],
```
webmaster128, line 107: It would be nice if we could get this list from `@cosmjs/stargate` and add the wasm types to it. The types are coming from `@cosmjs/stargate` already via

 ```
import {
  AminoTypes,
  BroadcastTxFailure,
  BroadcastTxResponse,
  codec,
  isBroadcastTxFailure,
  parseRawLog,
} from "@cosmjs/stargate";
```

File: packages/cosmwasm-stargate/src/cosmwasmclient.ts (hunk 1 above)
Comment (willclarktech) [reply packages/cosmwasm-stargate/src/cosmwasmclient.ts:71]: Good point!

File: packages/stargate/src/signingstargateclient.ts [hunk 3]
```diff
@@ -21,6 +21,7 @@ import {
   OfflineSigner,
   Registry,
 } from "@cosmjs/proto-signing";
+import { GeneratedType } from "@cosmjs/proto-signing/types/registry";
```
webmaster128, line 24: This should be in the index of `@cosmjs/proto-signing` (as it is a type of the public `Registry` constructor)

webmaster128 (approved):
Looks good.

packages/stargate/src/signingstargateclient.ts
willclarktech: The trouble with overactive IDEs.

PR #645, opened 2021-02-03
Fix Ledger test condition

willclarktech:
Otherwise the tests fail if simapp but not ledger is enabled

### DIFF
--- a/packages/ledger-amino/src/ledgersigner.spec.ts
+++ b/packages/ledger-amino/src/ledgersigner.spec.ts
@@ -23,6 +23,7 @@ import { LedgerSigner } from "./ledgersigner";
 import {
   faucet,
   launchpad,
+  ledgerEnabled,
   pendingWithoutLaunchpad,
   pendingWithoutLedger,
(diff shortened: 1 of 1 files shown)


webmaster128 (approved):

packages/ledger-amino/src/ledgersigner.spec.ts
```diff
@@ -61,7 +62,7 @@ describe("LedgerSigner", () => {
   let transport: Transport;
 
   beforeAll(async () => {
-    if (simappEnabled()) {
+    if (simappEnabled() && ledgerEnabled()) {
```
webmaster128, line 65: Wait a second, why does this block require Ledger? Souldn't this go into

```ts
  beforeEach(async () => {
    transport = await createTransport();
  });
```

willclarktech: Yep. Weird that this change stopped the test failure for me though.

webmaster128 (approved):

Issue #646, opened 2021-02-04
Move type definitions into `./build` and out of VCS


willclarktech:
After #616

PR #652, opened 2021-02-09
Export SigningStargateClientOptions from @cosmjs/stargate

### LINKED ISSUE #650: Export SigningStargateClientOptions from @cosmjs/stargate
I am making a wrapper / helper that instantiates a `SigningStargateClient` and then some other related actions. I wanted to accept the same options to pass into the `SigningStargateClient`, and ended up having to do this for import:

```
// TODO: this is wrong, expose in top level
import { SigningStargateClientOptions } from '@cosmjs/stargate/types/signingstargateclient';
```

The type is part of the public API and should be exported top level.

ethanfrey:
Closes #650

### DIFF
--- a/packages/stargate/src/index.ts
+++ b/packages/stargate/src/index.ts
@@ -30,4 +30,8 @@ export {
   SequenceResponse,
   StargateClient,
 } from "./stargateclient";
-export { defaultRegistryTypes, SigningStargateClient } from "./signingstargateclient";
+export {
+  defaultRegistryTypes,
+  SigningStargateClient,
+  SigningStargateClientOptions,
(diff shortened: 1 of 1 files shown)


webmaster128 (approved):
Thanks.

Nice to see how we only have one diff when the types are excluded from git.

ethanfrey:
I kept looking for that types dir, sure the CI would reject me for forgetting to check it in. Guess that is gone now? :relaxed:

willclarktech:
> Guess that is gone now?

@ethanfrey It's gone in `@cosmjs/stargate`. I'm just working on #646 to remove it elsewhere.

PR #653, opened 2021-02-09
Query client proof helper

### LINKED ISSUE #651: Add QueryClient helper that returns a proof
Unverified uses the grpc format and doesn't even attempt to get a proof.

Verified will get the proof, do all the verification, and only return the valid data in the end.

While building the relayer, I found I need another case, which does the query and gives me the raw data and proof - but does not do the validation. This proof then can be used to send in ibc messages to be verified by the remote chain.

I would want a variant that works with the raw store queries, as well that uses the grpc interface for convenience (if possible)

ethanfrey:
Closes #651 

This will let us get the full proof data from a QueryClient helper, so we can use it in a QueryExtension.
This function does the basic format validation but not the actual merkle / light client proof. It is a useful primitive to expose and we can use it to build the more complex calls.

### DIFF
--- a/packages/stargate/src/queries/queryclient.ts
+++ b/packages/stargate/src/queries/queryclient.ts
@@ -167,60 +167,32 @@ export class QueryClient {
   }
 
   public async queryVerified(store: string, key: Uint8Array): Promise<Uint8Array> {
-    const response = await this.tmClient.abciQuery({
-      // we need the StoreKey for the module, not the module name
-      // https://github.com/cosmos/cosmos-sdk/blob/8cab43c8120fec5200c3459cbf4a92017bb6f287/x/auth/types/keys.go#L12
-      path: `/store/${store}/key`,
-      data: key,
-      prove: true,
-    });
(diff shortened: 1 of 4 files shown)


ethanfrey:
Interesting side effect.
I actually required `queryVerfied` to have a proof (before it trusted everything if the proof was missing), and got this CI error from `wasmd/init.sh`:

```
Waiting for blockchain and REST server to be available ...
Waiting for height to be >= 1 ...
Okay, thank you for your patience.
Error: Expected 2 proof ops, got 0. Are you using stargate?
    at QueryClient.queryRawProof (/home/circleci/project/packages/stargate/build/queries/queryclient.js:80:19)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

It seems it was abusing that earlier. Maybe we wait for height 2 (so there are proofs), or just use unverified for the setup work?

willclarktech:
@ethanfrey For the stargate package we use `simapp` not `wasmd`.

willclarktech (approved):

PR #654, opened 2021-02-09
Move type definitions into build dirs

### LINKED ISSUE #646: Move type definitions into `./build` and out of VCS

willclarktech:
Closes #646


packages/cosmwasm/package.json
```diff
@@ -29,11 +28,10 @@
     "format-text": "prettier --write --prose-wrap always --print-width 80 \"./*.md\"",
     "lint": "eslint --max-warnings 0 \"**/*.{js,ts}\"",
     "lint-fix": "eslint --max-warnings 0 \"**/*.{js,ts}\" --fix",
-    "move-types": "shx rm -rf ./types/* && shx mv build/types/* ./types",
-    "format-types": "prettier --write --loglevel warn \"./types/**/*.d.ts\"",
+    "remove-test-types": "shx rm -rf build/*.spec.d.ts build/*/*.spec.d.ts",
```
webmaster128, line 31: is this removal still needed/useful?

webmaster128 (commented):
Could you have a look into `packages/stargate/.gitignore`? There we have some pattern instead of just `build/`. Do we need to streamline other .gitignore files as well?

willclarktech:
> There we have some pattern instead of just build/. Do we need to streamline other .gitignore files as well?

I believe this is just a hangover from when I was explicitly including files that were in the `build` directory.

webmaster128 (approved):

Issue #642, opened 2021-01-27
Split Tendermint client into Tendermint33Client and Tendermint34Client


webmaster128:
Creating a Tendermint client that makes the full Tendermint RPC API available to the user should be easy. Users need to do that to e.g. create custom quey clients.

Auto-detecting the Tendermint version of the backend does not work anymore starting with Tendermint 0.34. We should give it up. Also `Client` exposes and interface that tries to generalize between Tendermint 0.33 and Tendermint 0.34. If we split it into two clients, the interfaces can be more specific.

Right now, using the Tendermint client [looks like](
https://github.com/CosmWasm/ibc-visualizer/blob/55fe307/src/contexts/ClientContext.tsx)

```ts
import { adaptor34, Client as TendermintClient } from "@cosmjs/tendermint-rpc";

// ...

const tmClient = await TendermintClient.connect(config.rpcUrl, adaptor34);
```

which has a few flaws:

- `Client` is never a great name to import and always needs to be renamed. So it would be nice if we give it a useful name by default, even if slightly redundant with the package name
- `adaptor34` was designed to be private but was exposed later on. Not pretty to work with.
- The async in the constructor is only needed for auto-detection. It can probably be sync now.

Instead, this could look like

```ts
import { Tendermint34Client } from "@cosmjs/tendermint-rpc";

// ...

const tmClient = new Tendermint34Client(config.rpcUrl);
```

ethanfrey:
Great idea. We can have different response values for them as well, rather than auto-converting.

0.33 can keep the present return values.

for 0.34, I would suggest importing the generated tendermint types: https://github.com/cosmos/cosmjs/tree/main/packages/stargate/src/codec/tendermint/types

They use protobuf internally and marshal to json for the rpc. There will need to be a bit of checks/conversion I think, especially things like oneof are different between go and js I think, but most of it is there. And returning these protobuf types makes us more future-compatible with 0.35 which plans a grpc interface.

PR #661, opened 2021-02-10
Fix Tendermint RPC Commit signature types

### LINKED ISSUE #658: Incorrect type for Commit in tendermint-rpc
Running simapp (from scripts) and querying a commit gives me:

`curl http://localhost:26658/commit | jq .result.signed_header.commit`

```json
{
  "height": "21",
  "round": 0,
  "block_id": {
    "hash": "7E259DA907B95BB56E28B2041D7A19FD1E58ADB0494965C15A8D4719C67EC089",
    "parts": {
      "total": 1,
      "hash": "2F24E85EC71C73B04A5723A743C3171C181E59FE03C523D609A9D43978411ED2"
    }
  },
  "signatures": [
    {
      "block_id_flag": 2,
      "validator_address": "E359FA31E749FFDC571AA4E5F629F70CD53D328A",
      "timestamp": "2021-02-10T12:25:34.578877636Z",
      "signature": "ezhYxn2ktbFf2k2MKQovSyaXrxfx24cxpr8res8/LPZEKcGOHckwnjRuUlM9HKYjNduY3VWU+IiDlHXyzEL2Aw=="
    }
  ]
}
```

The current [Commit response from tendermint-rpc](https://github.com/CosmWasm/cosmjs/blob/52dc82bbe4799c2b4fcd1e69a81c9980ffcdc7c8/packages/tendermint-rpc/src/responses.ts#L223-L227) is quite different: 

```ts
export interface Commit {
  readonly blockId: BlockId;
  readonly signatures: readonly ValidatorSignature[];
}

export type ValidatorSignature = ValidatorEd25519Signature;

export interface ValidatorEd25519Signature {
  readonly algorithm: "ed25519";
  readonly data: Uint8Array;
}
```

Basically, `round` must be added and the `signatures` array looks quite different

You can look at the types from `tendermint/types/types.ts` in the protobuf definitions as a reference to what the rpc endpoint returns: https://github.com/confio/ts-relayer/blob/b0c348c417803fcdc4cac2bfe3b83413bf41a94d/src/codec/tendermint/types/types.ts#L176-L191 (and maybe just copy in those types)

willclarktech:
Closes #658

### DIFF
--- a/packages/tendermint-rpc/src/adaptors/v0-33/responses.ts
+++ b/packages/tendermint-rpc/src/adaptors/v0-33/responses.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import { fromBase64, fromHex } from "@cosmjs/encoding";
 import { JsonRpcSuccessResponse } from "@cosmjs/json-rpc";
+import { assert } from "@cosmjs/utils";
 
 import {
   assertArray,
@@ -18,7 +19,7 @@ import {
 } from "../../encodings";
 import * as responses from "../../responses";
 import { SubscriptionEvent } from "../../rpcclients";
-import { ValidatorPubkey, ValidatorSignature } from "../../types";
+import { BlockIdFlag, CommitSignature, ValidatorPubkey } from "../../types";
 import { hashTx } from "./hasher";
 
 interface AbciInfoResult {
@@ -396,6 +397,11 @@ function decodeBroadcastTxCommit(data: RpcBroadcastTxCommitResponse): responses.
   };
 }
 
+function decodeBlockIdFlag(blockIdFlag: number): BlockIdFlag {
+  assert(blockIdFlag in BlockIdFlag);
+  return blockIdFlag;
+}
+
 type RpcSignature = {
   readonly block_id_flag: number;
   /** hex encoded */
@@ -405,22 +411,28 @@ type RpcSignature = {
   readonly signature: string;
 };
 
-function decodeSignature(data: RpcSignature): ValidatorSignature {
+function decodeCommitSignature(data: RpcSignature): CommitSignature {
   return {
-    algorithm: "ed25519",
-    data: fromBase64(assertNotEmpty(data.signature)),
+    blockIdFlag: decodeBlockIdFlag(data.block_id_flag),
+    validatorAddress: fromHex(data.validator_address),
+    timestamp: new Date(assertNotEmpty(data.timestamp)),
+    signature: fromBase64(assertNotEmpty(data.signature)),
   };
 }
 
 interface RpcCommit {
   readonly block_id: RpcBlockId;
+  readonly height: string;
+  readonly round: string;
   readonly signatures: readonly RpcSignature[];
 }
 
 function decodeCommit(data: RpcCommit): responses.Commit {
   return {
     blockId: decodeBlockId(assertObject(data.block_id)),
-    signatures: assertArray(data.signatures).map(decodeSignature),
+    height: Integer.parse(assertNotEmpty(data.height)),
+    round: Integer.parse(data.round),
+    signatures: assertArray(data.signatures).map(decodeCommitSignature),
   };
 }
 
--- a/packages/tendermint-rpc/src/client.spec.ts
+++ b/packages/tendermint-rpc/src/client.spec.ts
@@ -115,12 +115,26 @@ function defaultTestSuite(rpcFactory: () => RpcClient, adaptor: Adaptor, expecte
     client.disconnect();
   });
 
+  it("can get a commit", async () => {
+    pendingWithoutTendermint();
+    const client = await Client.create(rpcFactory(), adaptor);
+    const response = await client.commit(4);
+
+    expect(response).toBeTruthy();
+    expect(response.commit.signatures.length).toBeGreaterThanOrEqual(1);
+    expect(response.commit.signatures[0].blockIdFlag).toEqual(2);
+    expect(response.commit.signatures[0].validatorAddress.length).toEqual(20);
+    expect(response.commit.signatures[0].timestamp).toBeInstanceOf(Date);
+    expect(response.commit.signatures[0].signature.length).toEqual(64);
+
+    client.disconnect();
+  });
+
   it("can call a bunch of methods", async () => {
     pendingWithoutTendermint();
     const client = await Client.create(rpcFactory(), adaptor);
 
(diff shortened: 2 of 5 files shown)


packages/tendermint-rpc/src/types.ts
```diff
@@ -22,3 +22,18 @@ export interface ValidatorEd25519Signature {
  * Currently only Ed25519 supported.
  */
 export type ValidatorSignature = ValidatorEd25519Signature;
+
+export enum BlockIdFlag {
+  BlockIdFlagUnknown = 0,
+  BlockIdFlagAbsent = 1,
+  BlockIdFlagCommit = 2,
+  BlockIdFlagNil = 3,
```
webmaster128, line 30: Do we really want to releat the enum name in the cases? As far as I can tell, we must write them anyways (`BlockIdFlag.BlockIdFlagCommit`)

packages/tendermint-rpc/src/types.ts
```diff
@@ -22,3 +22,18 @@ export interface ValidatorEd25519Signature {
...
+  Unrecognized = -1,
+}
+
+export interface CommitSignature {
+  blockIdFlag: BlockIdFlag;
+  validatorAddress: Uint8Array;
+  timestamp?: Date;
+  signature: ValidatorSignature;
+}
```
webmaster128, line 39: Do we need the new types in index.ts?

webmaster128 (commented):
Is this change compatible to both Tendermint 0.33 and 0.34?

ethanfrey:
> Is this change compatible to both Tendermint 0.33 and 0.34?

I adjusted the proofOp recently, even though it was for 0.34 not 0.33. However, as now one is using proofs with tendermint 0.33, it is breaking an unused functionality.

Either we keep doing that for such features that are different or bite the bullet and make a new 0.34 version.

I must say that there was no failing test nor any user complaints on this functionality before, so it is safe to say it will not break much. However, we could be clear and rename the adaptor to v0.34 and use a v0.33 alias, saying "it is compatible for normal usage, but not for commits or proofs"

packages/tendermint-rpc/src/adaptors/v0-33/responses.ts
```diff
@@ -396,6 +397,18 @@ function decodeBroadcastTxCommit(data: RpcBroadcastTxCommitResponse): responses.
   };
 }
 
+function decodeValidatorSignature(signature: string): ValidatorSignature {
```
ethanfrey, line 400: Since you are just guessing at the algorithm anyway, let's just decode it and return it as Uint8Array.

https://github.com/confio/ts-relayer/blob/b0c348c417803fcdc4cac2bfe3b83413bf41a94d/src/codec/tendermint/types/types.ts#L189 is just a Uint8Array and I will need to convert into that to submit it anyway. The algorithm info is available in the validator pubkey (which is a different endpoint)

packages/tendermint-rpc/src/adaptors/v0-33/responses.ts
```diff
@@ -405,22 +418,28 @@ type RpcSignature = {
...
 };
 
-function decodeSignature(data: RpcSignature): ValidatorSignature {
+function decodeCommitSignature(data: RpcSignature): CommitSignature {
   return {
-    algorithm: "ed25519",
-    data: fromBase64(assertNotEmpty(data.signature)),
+    blockIdFlag: decodeBlockIdFlag(data.block_id_flag),
+    validatorAddress: fromHex(data.validator_address),
+    timestamp: new Date(assertNotEmpty(data.timestamp)),
```
ethanfrey, line 425: Thanks for the Date and not ReadonlyDate here :)

webmaster128:
> Either we keep doing that for such features that are different or bite the bullet and make a new 0.34 version.

I'm fine either way. #642 is planned anyways. It is just a question of timing.

ethanfrey (commented):
Looks good.
Can you add a test case? Just a basic sanity check. Query `/commit` endpoint and get the value. Ensure there is one signature (whatever local app we test against) and all fields are set.

packages/tendermint-rpc/src/adaptors/v0-33/responses.ts
webmaster128: Using `Date` discards information. Do we need fractionl digits 4-9? Then we should create a `DateWithNanoseconds`, which works like `ReadonlyDateWithNanoseconds` but is a `Date` with an extra field that can be assigned to a `Date`.

ethanfrey: We don't need nanoseconds for now. We could create a new type as well.

ethanfrey:
> I'm fine either way. #642 is planned anyways. It is just a question of timing.

Good point here. Let's do the quick and dirty way now (it is blocking) and then do a nice refactor in #642

packages/tendermint-rpc/src/adaptors/v0-33/responses.ts
willclarktech: Let's make a new type when we need it.

packages/tendermint-rpc/src/adaptors/v0-33/responses.ts
willclarktech: Does this apply to the other responses too, eg voting, that were already using this?

ethanfrey: Voting? Which endpoint is that? consensus?

I would just run a local simapp and check the rpc response. Otherwise, I suggest looking at tendermint/types dir

willclarktech: `ValidatorSignature` is also used in the `Vote` type, but looks like this is just exported. I'll update to remove the algorithm guess.

packages/tendermint-rpc/src/client.spec.ts
```diff
@@ -12,7 +12,7 @@ import { ExpectedValues, tendermintInstances } from "./config.spec";
 import { buildQuery } from "./requests";
 import * as responses from "./responses";
 import { HttpClient, RpcClient, WebsocketClient } from "./rpcclients";
-import { chainIdMatcher } from "./testutil.spec";
+import { chainIdMatcher, hexMatcher } from "./testutil.spec";
```
webmaster128, line 15: `hexMatcher` is unused, isn't it?

webmaster128 (approved):

PR #664, opened 2021-02-10
Fix Tendermint RPC Validator types

### LINKED ISSUE #659: Fix validator type in tendermint rpc
The output from rpc looks like:

`curl http://localhost:26658/validators | jq .result`

```json
{
  "block_height": "845",
  "validators": [
    {
      "address": "E359FA31E749FFDC571AA4E5F629F70CD53D328A",
      "pub_key": {
        "type": "tendermint/PubKeyEd25519",
        "value": "L6f41DzvG/B+jUIc+avA3d0+BOhqgKB0pfiYSaug+2I="
      },
      "voting_power": "3",
      "proposer_priority": "0"
    }
  ],
  "count": "1",
  "total": "1"
}
```

The [types from tendermint-rpc](https://github.com/CosmWasm/cosmjs/blob/52dc82bbe4799c2b4fcd1e69a81c9980ffcdc7c8/packages/tendermint-rpc/src/responses.ts#L141-L145) look like:

```ts
export interface ValidatorsResponse {
  readonly blockHeight: number;
  readonly results: readonly Validator[];
}

export interface Validator {
  readonly address?: Uint8Array;
  readonly pubkey: ValidatorPubkey;
  readonly votingPower: number;
  readonly accum?: number;
  readonly name?: string;
}
```

Please:
* (failing) test that calling validators gets a list of one result with pubkey and non-zero voting power.
* rename results -> validators
* remove accum/name and add proposerPower

willclarktech:
Closes #659

### DIFF
--- a/packages/tendermint-rpc/src/client.spec.ts
+++ b/packages/tendermint-rpc/src/client.spec.ts
@@ -130,6 +130,24 @@ function defaultTestSuite(rpcFactory: () => RpcClient, adaptor: Adaptor, expecte
     client.disconnect();
   });
 
+  it("can get validators", async () => {
+    pendingWithoutTendermint();
+    const client = await Client.create(rpcFactory(), adaptor);
+    const response = await client.validators();
+
+    expect(response).toBeTruthy();
+    expect(response.blockHeight).toBeGreaterThanOrEqual(1);
+    expect(response.count).toBeGreaterThanOrEqual(1);
+    expect(response.total).toBeGreaterThanOrEqual(1);
+    expect(response.validators.length).toBeGreaterThanOrEqual(1);
+    expect(response.validators[0].address.length).toEqual(20);
+    expect(response.validators[0].pubkey).toBeDefined();
+    expect(response.validators[0].votingPower).toBeGreaterThanOrEqual(0);
+    expect(response.validators[0].proposerPriority).toBeGreaterThanOrEqual(0);
+
+    client.disconnect();
+  });
+
   it("can call a bunch of methods", async () => {
     pendingWithoutTendermint();
     const client = await Client.create(rpcFactory(), adaptor);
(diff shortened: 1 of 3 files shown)


packages/tendermint-rpc/src/client.spec.ts
```diff
@@ -130,14 +130,31 @@ function defaultTestSuite(rpcFactory: () => RpcClient, adaptor: Adaptor, expecte
...
+    pendingWithoutTendermint();
+    const client = await Client.create(rpcFactory(), adaptor);
+    const response = await client.validators();
+
+    expect(response).toBeTruthy();
+    expect(response.blockHeight).toBeGreaterThanOrEqual(1);
+    expect(response.count).toBeGreaterThanOrEqual(1);
+    expect(response.total).toBeGreaterThanOrEqual(1);
+    expect(response.validators.length).toBeGreaterThanOrEqual(1);
+    expect(response.validators[0].address!.length).toEqual(20);
```
webmaster128, line 143: Any clue if/why this address can be unset? Would be good to have a note in `export interface Validator {` if you do

webmaster128 (approved):

packages/tendermint-rpc/src/client.spec.ts
willclarktech: Checking the type from the codec it doesn't seem to be optional.

willclarktech: But pubkey does!

packages/tendermint-rpc/src/client.spec.ts
```diff
@@ -130,14 +130,31 @@ function defaultTestSuite(rpcFactory: () => RpcClient, adaptor: Adaptor, expecte
     client.disconnect();
   });
 
+  it("can get validators", async () => {
```
ethanfrey, line 133: nice

ethanfrey (approved):
Looking good

PR #665, opened 2021-02-10
Allow any string eventType in logs.findAttribute

### LINKED ISSUE #662: Allow any event type in logs.findAttribute
This is in @cosmjs/launchpad as part of the generic stuff that may go into it's own package (see #657)

The current type is `findAttribute(logs: readonly Log[], eventType: "message" | "transfer", attrKey: string): Attribute`

I need to get a lot more data with this for the relayer on ibc. To get the result from creating a channel, I had to do this:

```ts
    const clientId = logs.findAttribute(
      parsedLogs,
      // TODO: they enforce 'message' | 'transfer'
      /* eslint @typescript-eslint/no-explicit-any: "off" */
      'create_client' as any,
      'client_id'
    ).value;
```

This typing is unnecessarily string, it should accept any string

willclarktech:
Closes #662

### DIFF
--- a/packages/launchpad/src/logs.ts
+++ b/packages/launchpad/src/logs.ts
@@ -68,11 +68,7 @@ export function parseLogs(input: unknown): readonly Log[] {
  *
  * Throws if the attribute was not found.
  */
-export function findAttribute(
-  logs: readonly Log[],
-  eventType: "message" | "transfer",
-  attrKey: string,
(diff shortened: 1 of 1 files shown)


ethanfrey (approved):
Thanks. Looks good.

webmaster128 (approved):

PR #666, opened 2021-02-10
Add validatorsAll query method to tendermint-rpc

### LINKED ISSUE #660: Add auto-pagination to validators query in tendermint-rpc
It used to return all validators, now it does it by pages, default 30 per page.
This will cause a problem to the relayer when connecting to networks with many validators (eg. 100-150).

To get an example (as soon as a few more validators join)
https://rpc.musselnet.cosmwasm.com/validators?per_page=2&page=1

This rpc client call should make multiple requests and get the whole list, even if it is 3 or 4 pages. We can also specify 50 per_page, so it takes fewer calls. (Test on a testnet with many validators and a low page size)

willclarktech:
Merge after #664 
Closes #660

I added a separate method on the model of `txSearch`/`txSearchAll`.

### DIFF
--- a/packages/tendermint-rpc/src/adaptors/v0-33/requests.ts
+++ b/packages/tendermint-rpc/src/adaptors/v0-33/requests.ts
@@ -84,6 +84,19 @@ function encodeTxSearchParams(params: requests.TxSearchParams): RpcTxSearchParam
   };
 }
 
+interface RpcValidatorsParams {
+  readonly height?: string;
+  readonly page?: string;
+  readonly per_page?: string;
+}
+function encodeValidatorsParams(params: requests.ValidatorsParams): RpcValidatorsParams {
+  return {
+    height: may(Integer.encode, params.height),
+    page: may(Integer.encode, params.page),
+    per_page: may(Integer.encode, params.per_page),
+  };
+}
+
 export class Params {
   public static encodeAbciInfo(req: requests.AbciInfoRequest): JsonRpcRequest {
     return createJsonRpcRequest(req.method);
@@ -141,6 +154,6 @@ export class Params {
   }
 
   public static encodeValidators(req: requests.ValidatorsRequest): JsonRpcRequest {
-    return createJsonRpcRequest(req.method, encodeHeightParam(req.params));
+    return createJsonRpcRequest(req.method, encodeValidatorsParams(req.params));
   }
 }
--- a/packages/tendermint-rpc/src/client.spec.ts
+++ b/packages/tendermint-rpc/src/client.spec.ts
@@ -148,6 +148,24 @@ function defaultTestSuite(rpcFactory: () => RpcClient, adaptor: Adaptor, expecte
     client.disconnect();
   });
 
+  it("can get all validators", async () => {
+    pendingWithoutTendermint();
+    const client = await Client.create(rpcFactory(), adaptor);
+    const response = await client.validatorsAll({});
+
+    expect(response).toBeTruthy();
+    expect(response.blockHeight).toBeGreaterThanOrEqual(1);
+    expect(response.count).toBeGreaterThanOrEqual(1);
+    expect(response.total).toBeGreaterThanOrEqual(1);
+    expect(response.validators.length).toBeGreaterThanOrEqual(1);
+    expect(response.validators[0].address.length).toEqual(20);
+    expect(response.validators[0].pubkey).toBeDefined();
+    expect(response.validators[0].votingPower).toBeGreaterThanOrEqual(0);
+    expect(response.validators[0].proposerPriority).toBeGreaterThanOrEqual(0);
+
+    client.disconnect();
+  });
+
   it("can call a bunch of methods", async () => {
     pendingWithoutTendermint();
     const client = await Client.create(rpcFactory(), adaptor);
--- a/packages/tendermint-rpc/src/client.spec.ts
+++ b/packages/tendermint-rpc/src/client.spec.ts
@@ -133,7 +133,7 @@ function defaultTestSuite(rpcFactory: () => RpcClient, adaptor: Adaptor, expecte
   it("can get validators", async () => {
     pendingWithoutTendermint();
     const client = await Client.create(rpcFactory(), adaptor);
-    const response = await client.validators();
+    const response = await client.validators({});
 
     expect(response).toBeTruthy();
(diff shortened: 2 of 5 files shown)


packages/tendermint-rpc/src/adaptors/v0-33/requests.ts
```diff
@@ -84,6 +84,19 @@ function encodeTxSearchParams(params: requests.TxSearchParams): RpcTxSearchParam
   };
 }
 
+interface RpcValidatorsParams {
+  readonly height?: string;
+  readonly page?: string;
+  readonly per_page?: string;
+}
+function encodeValidatorsParams(params: requests.ValidatorsParams): RpcValidatorsParams {
```
ethanfrey, line 92: I'm a bit confused. This means we locally set a string but it is sent as a number over rpc?

Why not just use number everywhere. (The reverse, local numbers auto-encoding/decoding to rpc string, makes much more sense)

ethanfrey: Nevermind, it does take local numbers into rpc strings. Saw the usage below.

packages/tendermint-rpc/src/client.ts
```diff
@@ -248,14 +249,43 @@ export class Client {
...
+    let blockHeight = 0;
+
+    while (!done) {
+      const resp = await this.validators({
+        per_page: 50,
+        ...params,
+        page: page,
+      });
+      validators.push(...resp.validators);
+      blockHeight = resp.blockHeight;
```
ethanfrey, line 273: Okay, this is good to record, but what happens if a block passes while querying (and the validators changed even?

I would explicitly set the height before querying. (See above)

packages/tendermint-rpc/src/client.ts
```diff
@@ -248,14 +249,43 @@ export class Client {
...
 
+  public async validatorsAll(params: requests.ValidatorsParams): Promise<responses.ValidatorsResponse> {
+    let page = params.page || 1;
+    const validators: responses.Validator[] = [];
+    let done = false;
+    let blockHeight = 0;
+
+    while (!done) {
+      const resp = await this.validators({
+        per_page: 50,
```
ethanfrey, line 268: Let's just process params into args before the loop, using the input as defaults.

```ts
let page = params.page || 1;
let per_page = params.per_page || 50;
let height = params.height || (await this.status()).height;
let args = {per_page, height}

//....
const resp = await this.validators({...args, page});
```

packages/tendermint-rpc/src/client.spec.ts
```diff
@@ -130,14 +130,49 @@ function defaultTestSuite(rpcFactory: () => RpcClient, adaptor: Adaptor, expecte
...
+    expect(response.validators.length).toBeGreaterThanOrEqual(1);
+    expect(response.validators[0].address.length).toEqual(20);
+    expect(response.validators[0].pubkey).toBeDefined();
+    expect(response.validators[0].votingPower).toBeGreaterThanOrEqual(0);
+    expect(response.validators[0].proposerPriority).toBeGreaterThanOrEqual(0);
+
+    client.disconnect();
+  });
+
+  it("can get all validators", async () => {
```
ethanfrey, line 151: shame we don't have a multi-node testnet to test against.

ethanfrey (commented):
Generally looks good.
Added some comments.

packages/tendermint-rpc/src/client.ts
willclarktech: `this.status()` doesn't return a height. I'll leave it out for this PR. Is it something we should be parsing?

ethanfrey: Sure, that or from the latest header.

packages/tendermint-rpc/src/client.ts
```diff
@@ -248,14 +249,47 @@ export class Client {
...
+  public async validatorsAll(params: requests.ValidatorsParams): Promise<responses.ValidatorsResponse> {
+    let page = params.page || 1;
+    const validators: responses.Validator[] = [];
+    const baseParams = {
+      per_page: 50,
+      height: params.height ?? (await this.status()).syncInfo.latestBlockHeight,
+      ...params,
+    };
+    let done = false;
+    let blockHeight = 0;
```
ethanfrey, line 269: Hmmm... if this was undefined or number, we could use this instead of querying.
Whatever the first one requested, keep using it.

```
let page = params.page || 1;
let blockHeight = params.height;
let perPage = params.perPage || 50;

while (!done) {
    const resp = await this.validators({
      page: page,
      height: blockHeight,
      perPage: perPage,
    });
    // ??? only set if blockHeight was undefined before?
    blockHeight = resp.blockHeight;
}
```

ethanfrey: This also removes the need for the status call (the first response will give us the most recent validator set if height was undefined)

ethanfrey (approved):
Another look at the loop. Maybe I am being too nit-picky. I just know this can be tough to change after a release. Anyway, take it or leave it. And then feel free to merge

packages/tendermint-rpc/src/client.ts
```diff
@@ -248,14 +249,44 @@ export class Client {
...
+    let blockHeight = height;
+
+    while (!done) {
+      const response = await this.validators({
+        per_page: 50,
+        height: blockHeight,
+        page: page,
+      });
+      validators.push(...response.validators);
+      blockHeight = blockHeight || response.blockHeight;
```
ethanfrey, line 273: perfect

packages/tendermint-rpc/src/client.ts
```diff
@@ -248,14 +249,44 @@ export class Client {
...
+  public async validators(params: requests.ValidatorsParams): Promise<responses.ValidatorsResponse> {
     const query: requests.ValidatorsRequest = {
       method: requests.Method.Validators,
-      params: { height: height },
+      params: params,
     };
     return this.doCall(query, this.p.encodeValidators, this.r.decodeValidators);
   }
 
+  public async validatorsAll(height?: number): Promise<responses.ValidatorsResponse> {
```
ethanfrey, line 260: This is very nice choice and makes it all a lot simpler

ethanfrey (approved):
Great simplification!

PR #667, opened 2021-02-10
Avoid auto-importing types in CLI and install stargate deps

### LINKED ISSUE #621: Add stargate support to @cosmjs/cli
I was using a custom helper but failing to connect to musselnet.
Then I saw https://github.com/cosmos/cosmjs/blob/master/packages/cli/src/cli.ts#L40-L64

This is still focused on launchpad. I would prefer to use cosmjs rather than the cli tooling to interact with blockchains, and happy for stargate support. I am not sure the long-term plans for this package:

* split into cli-launchpad, cli-stargate 
* only stargate and let launchpad users use v0.23
* abandon it completely :cry: 

This is my swiss-army knife for working with testnets. Not sure what other options there are here

webmaster128:
Closes #621


packages/cli/package.json
```diff
@@ -40,11 +40,15 @@
   ],
   "dependencies": {
     "@cosmjs/cosmwasm-launchpad": "^0.24.0-alpha.23",
+    "@cosmjs/cosmwasm-stargate": "^0.24.0-alpha.23",
```
ethanfrey, line 43: nice to add those in for sure.

packages/cli/src/cli.ts
```diff
@@ -35,117 +49,47 @@ export async function main(originalArgs: readonly string[]): Promise<void> {
...
+        encodeAminoPubkey,
+        encodeBech32Pubkey,
+        decodeBech32Pubkey,
+        decodeAminoPubkey,
+        makeCosmoshubPath,
+        makeSignDoc,
+        Secp256k1HdWallet,
+        Secp256k1Wallet,
+        StdFee,
+      } from "@cosmjs/launchpad";
```
ethanfrey, line 89: Maybe some stargate imports here (or in the example scripts)

packages/cli/src/cli.ts
```diff
@@ -35,117 +49,47 @@ export async function main(originalArgs: readonly string[]): Promise<void> {
...
-  }
-
-  let init = `
-    import axios from "axios";
-    import * as fs from "fs";
-  `;
-  for (const [moduleName, symbols] of imports.entries()) {
-    init += `import { ${symbols.join(", ")} } from "${moduleName}";\n`;
-  }
+  const visiblePackages = (await installedPackages()).filter(
```
ethanfrey, line 53: very nice to auto-generate them

ethanfrey (approved):
Looks nice, maybe some stargate tests

packages/cli/package.json
```diff
@@ -40,11 +40,15 @@
...
     "@cosmjs/crypto": "^0.24.0-alpha.23",
     "@cosmjs/encoding": "^0.24.0-alpha.23",
     "@cosmjs/faucet-client": "^0.24.0-alpha.23",
     "@cosmjs/launchpad": "^0.24.0-alpha.23",
     "@cosmjs/math": "^0.24.0-alpha.23",
+    "@cosmjs/proto-signing": "^0.24.0-alpha.23",
+    "@cosmjs/stargate": "^0.24.0-alpha.23",
+    "@cosmjs/tendermint-rpc": "^0.24.0-alpha.23",
```
willclarktech, line 51: `@cosmjs/tendermint-rpc` could be an optional dependency? Maybe it's unnecessarily complicated though.

willclarktech (approved):
Nice

packages/cli/package.json
webmaster128: Thy optional? I added it here to allow users to access @cosmjs/tendermint-rpc functionality directly. And it is pollued in by `@cosmjs/stargate` anyways.

willclarktech: I was thinking that since it's not directly depended on by any of the code here it's up to the user whether they want to be able to import it. But since it's installed anyway I guess there's no real benefit, it just makes things more complicated.

packages/cli/src/cli.ts
webmaster128: I add an example script which talks to the chain. This can then be executed conditionally.

packages/cli/package.json
webmaster128: The problem I see with packages that are not installed automatically is that they are very hard to install on demand, especially when using `npx` to start the CLI.

packages/cli/src/cli.ts
webmaster128: Stargate example added in the last commit. Thanks

PR #668, opened 2021-02-10
Simple query to get sequence to sign

ethanfrey:
There is no real reason we need merkle proofs to check out sequence and accountNumber (if we really care, we track them locally). By using the verifying query here, we take 1 block for the query (it must wait for next header), then a second block to process the transaction. That means submitting transactions are twice as slow as they need to be.

They are the slowest thing in the system, let's try to speed them up a bit.

(Pulled this out of another PR so we can debate this point separate from proof bugfixes)


webmaster128 (approved):
Seems like verified queries will never become the default. Too many limitations.

PR #669, opened 2021-02-10
Add examples and docs to readme

okwme:


webmaster128:
Thank you Billy. Don't worry about the CI fail (formatting). I'll pick it up from here.

PR #671, opened 2021-02-10
Add optional height argument to QueryClient methods

### LINKED ISSUE #670: Ability to set height in QueryClient methods
At least when getting proofs for ibc packets, I will need to make 2-4 queries all at the same height so the proofs have the same root.

Neither QueryVerified not QueryRawProof allow one to pass that in. 

I propose to add an optional `height?: number` argument, which can be used to set the height without breaking existing code.

ethanfrey:
Closes #670

Allows a QueryExtension to get a proof by height. How to pass in the height is an issue for the calling library, but we should provide the ability to configure it.

### DIFF
--- a/packages/stargate/src/queries/queryclient.ts
+++ b/packages/stargate/src/queries/queryclient.ts
@@ -166,8 +166,8 @@ export class QueryClient {
     this.tmClient = tmClient;
   }
 
-  public async queryVerified(store: string, key: Uint8Array): Promise<Uint8Array> {
-    const { height, proof, value } = await this.queryRawProof(store, key);
+  public async queryVerified(store: string, key: Uint8Array, desiredHeight?: number): Promise<Uint8Array> {
+    const { height, proof, value } = await this.queryRawProof(store, key, desiredHeight);
 
     const subProof = checkAndParseOp(proof.ops[0], "ics23:iavl", key);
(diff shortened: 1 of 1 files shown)


packages/stargate/src/queries/queryclient.ts
```diff
@@ -197,13 +197,18 @@ export class QueryClient {
...
+    queryKey: Uint8Array,
+    desiredHeight?: number,
+  ): Promise<ProvenQuery> {
     const { key, value, height, proof, code, log } = await this.tmClient.abciQuery({
       // we need the StoreKey for the module, not the module name
       // https://github.com/cosmos/cosmos-sdk/blob/8cab43c8120fec5200c3459cbf4a92017bb6f287/x/auth/types/keys.go#L12
       path: `/store/${store}/key`,
       data: queryKey,
       prove: true,
+      height: desiredHeight,
```
webmaster128, line 211: How is it possible that state is available for past heights? Is every state key available for every height?

ethanfrey: It depends. You can keep every state (archive node) or only the current one (prunning=all).

Default is to keep historical state for the last 100 block I believe. They have a very complex database algorithm to maintain these historical state with proofs.

ethanfrey: Dig into iavl tree and look into orphan nodes and pruning

webmaster128 (approved):

PR #675, opened 2021-02-11
Return gas used

ethanfrey:
This is very useful info in the apps when constructing transactions, so we can adjust the gas and fees for submitting transactions. We get it from tendermint rpc, but didn't parse it nor return it in the StargateClient.

This adds the extra info.

TODO: also add codespace, which is important along with the code for identifying errors

### DIFF
--- a/packages/stargate/src/stargateclient.ts
+++ b/packages/stargate/src/stargateclient.ts
@@ -62,6 +62,7 @@ export interface BroadcastTxSuccess {
   readonly transactionHash: string;
   readonly rawLog?: string;
   readonly data?: readonly MsgData[];
+  readonly gasUsed: number;
 }
 
 export type BroadcastTxResponse = BroadcastTxSuccess | BroadcastTxFailure;
@@ -262,6 +263,7 @@ export class StargateClient {
         transactionHash: toHex(response.hash).toUpperCase(),
         rawLog: response.deliverTx?.log,
         data: response.deliverTx?.data ? TxMsgData.decode(response.deliverTx?.data).data : undefined,
+        gasUsed: response.deliverTx?.gasUsed || 0,
       };
     }
     return response.checkTx.code !== 0
@@ -274,7 +276,7 @@ export class StargateClient {
         }
       : {
           height: response.height,
-          code: response.deliverTx?.code,
+          code: response.deliverTx?.code || 0,
           transactionHash: toHex(response.hash).toUpperCase(),
           rawLog: response.deliverTx?.log,
           data: response.deliverTx?.data ? TxMsgData.decode(response.deliverTx?.data).data : undefined,
--- a/packages/tendermint-rpc/src/responses.ts
+++ b/packages/tendermint-rpc/src/responses.ts
@@ -178,7 +178,7 @@ export interface TxData {
   readonly log?: string;
   readonly data?: Uint8Array;
   readonly events: readonly Event[];
-  readonly gasWanted: number;
+  readonly gasWanted?: number;
   readonly gasUsed?: number;
 }
 
--- a/packages/tendermint-rpc/src/responses.ts
+++ b/packages/tendermint-rpc/src/responses.ts
@@ -171,12 +171,15 @@ export interface Event {
   readonly attributes: readonly Attribute[];
 }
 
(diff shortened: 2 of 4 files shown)


packages/stargate/src/stargateclient.ts
```diff
@@ -274,7 +276,7 @@ export class StargateClient {
         }
       : {
           height: response.height,
-          code: response.deliverTx?.code,
+          code: response.deliverTx?.code || 0,
```
webmaster128, line 279: If we set the default here, could you make the `code` in the interface non-optional?

packages/stargate/src/stargateclient.ts
```diff
@@ -62,6 +62,7 @@ export interface BroadcastTxSuccess {
   readonly transactionHash: string;
   readonly rawLog?: string;
   readonly data?: readonly MsgData[];
+  readonly gasUsed: number;
```
webmaster128, line 65: Hmm, BroadcastTxSuccess does not mean it was executed. If we assume this type is only used for `broadcastTxCommit`, we should make it specific to that case.

ethanfrey:
Okay, let's punt this one for today's release and re-visit it with time. I think the issue is annotating this with all the possible approaches. There are actually 3. `async` is fire and forget, not even waiting for checktx and returns no info. `sync` waits for check to pass or fail, and then returns data. `block` (default) returns early if check fails. If check passes, it returns the deliver result (pass or fail).

I think the [current logic](https://github.com/CosmWasm/cosmjs/blob/1fc22afb3d5cde737c685bf3d73e0fe2dd09dd13/packages/stargate/src/stargateclient.ts#L257-L283) only works properly for block. I added some annotations:

```ts
  public async broadcastTx(tx: Uint8Array): Promise<BroadcastTxResponse> {
    const response = await this.tmClient.broadcastTxCommit({ tx });
    // this matches if check and deliver are both set and both code == 0 (block success)
    if (broadcastTxCommitSuccess(response)) {
      return {
        height: response.height,
        transactionHash: toHex(response.hash).toUpperCase(),
        rawLog: response.deliverTx?.log,
        data: response.deliverTx?.data ? TxMsgData.decode(response.deliverTx?.data).data : undefined,
      };
    }
    return response.checkTx.code !== 0
      ? {  
          // this matches for sync or block if it fails early in checktx
          height: response.height,
          code: response.checkTx.code,
          transactionHash: toHex(response.hash).toUpperCase(),
          rawLog: response.checkTx.log,
          data: response.checkTx.data ? TxMsgData.decode(response.checkTx.data).data : undefined,
        }
      : {
         // this one gets confused. In block, this is the case that check passed and deliver failed (this is correct return value)
         // if sync, then check passed, but there is no deliver info. What do we return??? (this will work but just have height and hash, and the height is not guaranteed to be the block height it is included in)
         // if async, we don't even have check and returning this as success mode is quite confusing
          height: response.height,
          code: response.deliverTx?.code,
          transactionHash: toHex(response.hash).toUpperCase(),
          rawLog: response.deliverTx?.log,
          data: response.deliverTx?.data ? TxMsgData.decode(response.deliverTx?.data).data : undefined,
        };
  }
```

Let's revisit this again later when all the major issues blocking the relayer have been resolved.

packages/tendermint-rpc/src/responses.ts
```diff
@@ -171,12 +171,15 @@ export interface Event {
   readonly attributes: readonly Attribute[];
 }
 
+// All fields are defined here: https://github.com/tendermint/tendermint/blob/v0.34.2/abci/types/types.pb.go#L2061-L2070
```
webmaster128, line 174: The fields are for `ResponseDeliverTx`, but `TxData` is used in checkTx as well:

```ts
export interface BroadcastTxSyncResponse extends TxData {
  readonly hash: Uint8Array;
}
```

and

```
export interface BroadcastTxCommitResponse {
  readonly height: number;
  readonly hash: Uint8Array;
  readonly checkTx: TxData;
  readonly deliverTx?: TxData;
}
```

webmaster128: I think it is better to split the Tendermint clients now and then closely follow the new types

webmaster128:
Closing iin favour of https://github.com/cosmos/cosmjs/issues/761 where we do the same thing, but for 0.34 only.

PR #677, opened 2021-02-11
Cleanup date

webmaster128:
Follow up on #676 

- Create dedicates dates module
- Create mutable `DateWithNanoseconds`
- Prefer free functions over `DateTime` namespace class

### DIFF
--- a/packages/tendermint-rpc/src/dates.ts
+++ b/packages/tendermint-rpc/src/dates.ts
@@ -26,6 +26,7 @@ export function toRfc3339WithNanoseconds(dateTime: ReadonlyDateWithNanoseconds):
   return `${millisecondIso.slice(0, -1)}${nanoseconds.padStart(6, "0")}Z`;
 }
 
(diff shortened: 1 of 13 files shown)


packages/tendermint-rpc/src/dates.ts
```diff
@@ -0,0 +1,39 @@
...
+  /* Nanoseconds after the time stored in a vanilla ReadonlyDate (millisecond granularity) */
+  readonly nanoseconds?: number;
+}
+
+export interface DateWithNanoseconds extends Date {
+  /** Nanoseconds after the time stored in a vanilla Date (millisecond granularity) */
+  nanoseconds?: number;
+}
+
+export function fromRfc3339WithNanoseconds(dateTimeString: string): DateWithNanoseconds {
```
ethanfrey, line 14: Nice to return the more permissive variant, which can "down cast" to the Readonly variant

ethanfrey (approved):
Looks good

PR #682, opened 2021-02-15
Add toSeconds/fromSeconds converters

### LINKED ISSUE #678: Add seconds/nanos converter to {ReadonlyDate,Date}WithNanoseconds
The tendermint-rpc types `{ReadonlyDate,Date}WithNanoseconds` use an unvonventional way to store the data. In one component, the whole digits and the fractioal digits 1-3 are stored. In another component the fractional digits 4-9 are stored. This is because DateWithNanoseconds is a Date and ReadonlyDateWithNanoseconds is a ReadonlyDate, and the inherit all the functionality from the standard library.

When using the types in protobuf (or any other external system), it is helpful to be able to split into second/nanoseconds. Callers can do that using the code like this

```ts
function timestampFromDateNanos(date: ReadonlyDateWithNanoseconds): Timestamp {
  const nanos = (date.getTime() % 1000) * 1000000 + (date.nanoseconds ?? 0);
  return Timestamp.fromPartial({
    seconds: new Long(date.getTime() / 1000),
    nanos,
  });
}
```

It would be convenient if ComsJS implemented and tested the logic for the user, without making assumptions on the types used for `seconds`/`nanos`.

If we create a helper function:

```ts
toComponents: () => { seconds: number, nanos: number } {
  return {
    seconds: Math.floor(date.getTime() / 1000),
    nanos: (date.getTime() % 1000) * 1000000 + (date.nanoseconds ?? 0),
  }
}
```

users can do

```ts
function timestampFromDateNanos(date: ReadonlyDateWithNanoseconds): Timestamp {
  const {seconds, nanos } = date.toComponents();
  return Timestamp.fromPartial({ seconds: Long.fromNumber(seconds), nanos });
}
```

webmaster128:
Closes #678

### DIFF
--- a/packages/tendermint-rpc/src/dates.ts
+++ b/packages/tendermint-rpc/src/dates.ts
@@ -1,4 +1,5 @@
 import { fromRfc3339 } from "@cosmjs/encoding";
+import { Uint32 } from "@cosmjs/math";
 import { ReadonlyDate } from "readonly-date";
 
 export interface ReadonlyDateWithNanoseconds extends ReadonlyDate {
@@ -23,6 +24,16 @@ export function toRfc3339WithNanoseconds(dateTime: ReadonlyDateWithNanoseconds):
   const millisecondIso = dateTime.toISOString();
   const nanoseconds = dateTime.nanoseconds?.toString() ?? "";
   return `${millisecondIso.slice(0, -1)}${nanoseconds.padStart(6, "0")}Z`;
+}
+
+export function fromSeconds(seconds: number, nanos = 0): DateWithNanoseconds {
+  const checkedNanos = new Uint32(nanos).toNumber();
+  if (checkedNanos > 999_999_999) {
+    throw new Error("Nano seconds must not exceed 999999999");
+  }
+  const out: DateWithNanoseconds = new Date(seconds * 1000 + Math.floor(checkedNanos / 1000000));
+  out.nanoseconds = checkedNanos % 1000000;
+  return out;
 }
 
 /**
--- a/packages/tendermint-rpc/src/dates.ts
+++ b/packages/tendermint-rpc/src/dates.ts
@@ -37,7 +37,7 @@ export function fromSeconds(seconds: number, nanos = 0): DateWithNanoseconds {
 }
 
 /**
- * Caclulates the UNIX timestamp in seconds as well as the nanoseconds after the given second.
+ * Calculates the UNIX timestamp in seconds as well as the nanoseconds after the given second.
  *
  * This is useful when dealing with external systems like the protobuf type
  * [.google.protobuf.Timestamp](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp)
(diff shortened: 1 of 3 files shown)


packages/tendermint-rpc/src/dates.ts
```diff
@@ -25,6 +26,29 @@ export function toRfc3339WithNanoseconds(dateTime: ReadonlyDateWithNanoseconds):
   return `${millisecondIso.slice(0, -1)}${nanoseconds.padStart(6, "0")}Z`;
 }
 
+export function fromSeconds(seconds: number, nanos = 0): DateWithNanoseconds {
```
ethanfrey, line 29: I see how this can be useful in repos such as ts-relayer. I did make a few errors before I got the timestamps calculated properly. :+1:

ethanfrey (approved):
lgtm

packages/tendermint-rpc/src/dates.ts
```diff
@@ -25,6 +26,29 @@ export function toRfc3339WithNanoseconds(dateTime: ReadonlyDateWithNanoseconds):
...
+  if (checkedNanos > 999_999_999) {
+    throw new Error("Nano seconds must not exceed 999999999");
+  }
+  const out: DateWithNanoseconds = new Date(seconds * 1000 + Math.floor(checkedNanos / 1000000));
+  out.nanoseconds = checkedNanos % 1000000;
+  return out;
+}
+
+/**
+ * Caclulates the UNIX timestamp in seconds as well as the nanoseconds after the given second.
```
willclarktech, line 40: Typo: "Calculates"

packages/tendermint-rpc/src/dates.ts
```diff
@@ -25,6 +26,29 @@ export function toRfc3339WithNanoseconds(dateTime: ReadonlyDateWithNanoseconds):
...
+ *
+ * This is useful when dealing with external systems like the protobuf type
+ * [.google.protobuf.Timestamp](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp)
+ * or any other system that does not use millisecond precision.
+ */
+export function toSeconds(date: ReadonlyDateWithNanoseconds): [number, number] {
```
willclarktech, line 46: Why not return a `Timestamp`?

webmaster128: I wanted to have this function independent of protobuf generated code. There are too many implementation details, such as number/Long/string for the `seconds`.

We could return an object instead, but then it is a bit harder for the caller to rename the components. Now

```ts
function timestampFromDateNanos(date: ReadonlyDateWithNanoseconds): Timestamp {
  const [seconds, nanos] = toSeconds(date);
  return Timestamp.fromPartial({ seconds: Long.fromNumber(seconds), nanos });
}
```

with object

```ts
function timestampFromDateNanos(date: ReadonlyDateWithNanoseconds): Timestamp {
  const { seconds, nanoseconds: nanos } = toSeconds(date);
  return Timestamp.fromPartial({ seconds: Long.fromNumber(seconds), nanos });
}
```

But I'm open to this alternative.

ethanfrey: Then why not return `{seconds, nanos}` ? And standarize with a "Timestamp-like" format

webmaster128: If the two of you prefer that over an unnamed tuples, this is fine for me. I don't have a real preference.

willclarktech: I also don't have any real preference.

webmaster128: I'll change to the object, which is easier to debug for our non-TS users

PR #683, opened 2021-02-15
Update ts-proto

webmaster128:
At https://github.com/cosmos/cosmjs/wiki/Maintenance#update-ts-proto we now have instructions how to regenerate our codecs


package.json
```diff
@@ -59,7 +59,7 @@
     "ses": "^0.11.0",
     "shx": "^0.3.2",
     "source-map-support": "^0.5.19",
-    "ts-proto": "^1.53.0",
+    "ts-proto": "=1.61.0",
```
willclarktech, line 62: `=` is not actually a defined prefix for npm. See https://docs.npmjs.com/cli/v6/configuring-npm/package-json#dependencies

> `version` Must match `version` exactly

webmaster128: Good to know, thanks. This should go back to `^` anyways when https://github.com/stephenh/ts-proto/issues/218 is resolved.

webmaster128: > = Equal. If no operator is specified, then equality is assumed, so this operator is optional, but MAY be included.

https://docs.npmjs.com/cli/v6/using-npm/semver#ranges

webmaster128:
@willclarktech this is ready for another review & merge. Commits starting with "Upgrade ts-proto to 1.67.0 and regenerate" are new. I had to re-generate the test vectors are restructure tests a little bit because body bytes and auth info bytes are now different from what Cosmos SDK generated. This is perfectly legal and intended. The reason those fields are `bytes` in the SignDoc is to allow different serialization in all messages except the root `SignDoc`.

willclarktech (approved):

PR #689, opened 2021-02-23
Create dedicated Tendermint34Client

### LINKED ISSUE #642: Split Tendermint client into Tendermint33Client and Tendermint34Client

webmaster128:
`Tendermint34Client` is a copy of `Client` without version detection. I don't think we need to create a `Tendermint33Client`. Instead the legacy `Client` can be used.

Closes #642


packages/tendermint-rpc/src/index.ts
```diff
@@ -78,6 +47,38 @@ export {
...
+  QueryTag,
+  StatusRequest,
+  SubscriptionEventType,
+  TxParams,
+  TxRequest,
+  TxSearchParams,
+  TxSearchRequest,
+  ValidatorsRequest,
+  ValidatorsParams,
+} from "./legacy";
```
willclarktech, line 72: Why not just `export * from "./legacy"`?

packages/tendermint-rpc/src/rpcclients/websocketclient.spec.ts
```diff
@@ -57,8 +56,9 @@ describe("WebsocketClient", () => {
 
         if (events.length === 2) {
           // make sure they are consequtive heights
```
willclarktech, line 58: Can you fix the "consecutive" typo while you're here?

packages/tendermint-rpc/src/tendermint34/index.ts
```diff
@@ -0,0 +1,74 @@
+// Note: all exports in this module are publicly available via
+// `import { tendermint34 } from "@cosnjs/tendermint-rpc"`
```
willclarktech, line 2: typo: "@cosmjs"

packages/tendermint-rpc/src/index.ts
webmaster128: I'd rather see that is exported in the top-level file, especially since wee need to take care of collisions now. I was thinking about `export * as legacy from "./legacy"` but this would be heavily breaking.

webmaster128:
okay, 3 more commits starting at _Deduplicate test utils_

willclarktech (approved):

ethanfrey:
Nice one

PR #691, opened 2021-02-23
Remove detectVersion call from Tendermint34Client

webmaster128:


webmaster128:
Still no idea why the extra call helps but 🤷‍♂️

PR #693, opened 2021-02-24
Bring back support for pbjs generated types

### LINKED ISSUE #692: pbjs generated types compatibility.
Hello and excuse me if the answer is obvious as I haven't looked heavily into the inner workings of pbsjs/ts-proto but what is the reason for the switch from `.fromJSON` to `.fromPartial` ?

The reason I'm asking is because generated types from pbjs only output the former function while ones from ts-proto output both functions. Thus keeping `.fromJSON` seems like would aid backwards compatibility.

As it stands we need to update our build pipeline to use ts-proto instead to generate types (which isn't as easy as it sounds) or cosmjs complains with a:
`TypeError: type.fromPartial is not a function`

webmaster128:
Closes #692


packages/proto-signing/src/registry.ts
```diff
@@ -56,6 +81,22 @@ export class Registry {
     this.types.set(typeUrl, type);
   }
 
+  /**
+   * Looks up a type that was proviously added to the registry.
```
willclarktech, line 85: Typo: "previously"

willclarktech (approved):
Very nice

packages/proto-signing/src/registry.ts
webmaster128: Thanks!

fadeev:
Thanks, @webmaster128 for working on this! 🙏We'll be using protobujs initially, and most likely eventually transition to ts-proto.

webmaster128:
@fadeev I'm happy to support both for the sake of modularaliry and avoiding vendor lock-in.

PR #700, opened 2021-03-11
Update testing blockchains and prepare 0.24.0 release

### LINKED ISSUE #638: Update scripts to cosmos-sdk 0.41.0 and wasmd 0.15.0
This includes some minor ibc fixes, and notably "amino signing" support for `MsgTransfer`.
This will be the version to go live on the cosmos hub, so we should test it. There should be no breaking api changes, just internal (accepts messages that were previously rejected)

webmaster128:
Closes #638


ethanfrey (approved):
Perfect

PR #701, opened 2021-03-11
Upgrade client to support Comso SDK 0.42 and wasmd 0.16

### LINKED ISSUE #695: Ensure CosmWasm 0.14 / Wasmd 0.16 compatibility
There should not be any breaking changes here, mainly some new entrypoints.
At the minimum, we should update the versions in the CI and ensure they work

webmaster128:
Closes #695


ethanfrey (approved):
Changes look good.
Failure seems to be the issue of not-encoding/decoding empty fields, while fromPartial does set them. That's a bit annoying to fix in general, but I'm sure there is a simple patch for the one or 2 tests (eg. don't use fromPartial if you want to allow undefined fields)

webmaster128:
> Failure seems to be the issue of not-encoding/decoding empty fields, while fromPartial does set them.

This is a ts-proto bug: https://github.com/stephenh/ts-proto/issues/237. I'm looking into it now.

PR #706, opened 2021-03-12
Make CommitSignature fields optional

webmaster128:
A little nicer API for the fix from 0.24.1

### DIFF
--- a/packages/tendermint-rpc/src/tendermint34/adaptors/v0-34/responses.ts
+++ b/packages/tendermint-rpc/src/tendermint34/adaptors/v0-34/responses.ts
@@ -3,7 +3,7 @@ import { fromBase64, fromHex } from "@cosmjs/encoding";
 import { JsonRpcSuccessResponse } from "@cosmjs/json-rpc";
 import { assert } from "@cosmjs/utils";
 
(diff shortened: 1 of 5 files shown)


packages/tendermint-rpc/src/tendermint34/adaptors/v0-34/responses.ts
```diff
@@ -416,17 +416,23 @@ type RpcSignature = {
   readonly signature: string | null;
 };
 
+/**
+ * In some cases a timestamp is optional and set to the value 0 in Go.
+ * This can lead to strings like "0001-01-01T00:00:00Z" (see https://github.com/cosmos/cosmjs/issues/704#issuecomment-797122415).
+ * This decoder tries to clean up such encoding from the API and turn them
+ * into undefined values.
+ */
+function decodeOptionalTime(timestamp: string): DateWithNanoseconds | undefined {
```
ethanfrey, line 425: Nice this looks clean.

You only get these absent commit signatures in a multi-node environment, when one node is down (or most recent commit when one has slow network)

ethanfrey (approved):
lgtm

PR #707, opened 2021-03-13
Add support for PeriodicVestingAccount (and many more account types)

### LINKED ISSUE #354: Add support for different account types in Stargate account query
See https://github.com/CosmWasm/cosmjs/pull/352#discussion_r466339476

User request:

- [ ] PeriodicVestingAccount

webmaster128:
Closes #354

~This is just a draft and I don't think we want a type system like that where the auth query returns a union type of dozens of options. Instead, we'll probably want to return just `Any` and let `function accountFromProto(` do the decoding and switching over the types.~

### DIFF
--- a/packages/stargate/src/accounts.spec.ts
+++ b/packages/stargate/src/accounts.spec.ts
@@ -1,0 +1,27 @@
+import { accountFromAny } from "./accounts";
+import { Any } from "./codec/google/protobuf/any";
+
+describe("accounts", () => {
+  describe("accountFromAny", () => {
+    it("works for PeriodicVestingAccount", () => {
+      // Arbitrary entry from https://raw.githubusercontent.com/ovrclk/net/24ddbb427/mainnet/genesis.json
+      // queried from chain via `packages/cli/examples/get_vesting_account.ts`.
+      const any = Any.fromJSON({
+        typeUrl: "/cosmos.vesting.v1beta1.PeriodicVestingAccount",
+        value:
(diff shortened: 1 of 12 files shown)


packages/stargate/src/accounts.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { accountFromAny } from "./accounts";
+import { Any } from "./codec/google/protobuf/any";
+
+describe("accounts", () => {
+  describe("accountFromAny", () => {
+    it("works for PeriodicVestingAccount", () => {
+      // Arbitrary entry from https://raw.githubusercontent.com/ovrclk/net/24ddbb427/mainnet/genesis.json
+      // queried from chain via `packages/cli/examples/get_vesting_account.ts`.
```
willclarktech, line 8:
```suggestion
      // queried from chain via `packages/cli/examples/get_akash_vesting_account.ts`.
```

packages/stargate/src/accounts.ts
```diff
@@ -0,0 +1,85 @@
...
+    pubkey: pubkey,
+    accountNumber: uint64FromProto(accountNumber).toNumber(),
+    sequence: uint64FromProto(sequence).toNumber(),
+  };
+}
+
+/**
+ * Takes an `Any` encoded account from the chain and extracts some common
+ * `Account` information from it. This is supposed to support the most relevant
+ * common Cosmos SDK account types. If you need support for exotix account types,
```
willclarktech, line 42:
```suggestion
 * common Cosmos SDK account types. If you need support for exotic account types,
```

packages/stargate/src/queries/auth.ts
```diff
@@ -1,17 +1,27 @@
...
 import { createRpc, toAccAddress } from "./utils";
 
 export interface AuthExtension {
   readonly auth: {
-    readonly account: (address: string) => Promise<BaseAccount | null>;
+    /**
+     * Returns an account if it exists and `null` otherwise.
+     *
+     * The account is a protobuf Any in order to be able to support many different
+     * account types in one API. The caller needs to switch over the expeced and supported
```
willclarktech, line 12:
```suggestion
     * account types in one API. The caller needs to switch over the expected and supported
```

packages/stargate/src/queries/auth.ts
```diff
@@ -1,17 +1,27 @@
...
+     * `typeUrl` and decode the `value` using its own type decoder.
+     */
+    readonly account: (address: string) => Promise<Any | null>;
     readonly unverified: {
-      readonly account: (address: string) => Promise<BaseAccount | null>;
+      /**
+       * Returns an account if it exists and `null` otherwise.
+       *
+       * The account is a protobuf Any in order to be able to support many different
+       * account types in one API. The caller needs to switch over the expeced and supported
```
willclarktech, line 21:
```suggestion
       * account types in one API. The caller needs to switch over the expected and supported
```

willclarktech (approved):

PR #710, opened 2021-03-16
Expose registry and fee table from signing clients

### LINKED ISSUE #628: Feature Request: Ability to register new message types on a previously instantiated signing client
Currently, a developer would have to load and register all Message Types used by his app prior to instantiating the client.

It would be preferable that we can instantiate the client and then register more message types as desired at runtime.

e.g. something like 
```
const registry=new Registry();
const wallet = await DirectSecp256k1HdWallet.fromMnemonic(mnemonic);
const client = await SigningStargateClient.connectWithWallet(api, wallet, { registry });
client.getRegistry().registerType(typeUrl,SomeMessageType);
```

willclarktech:
Closes #628

I also exposed the fee table because they’re readonly anyway and it might be useful to see the fees.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to
 
 ### Added
 
+- @cosmjs/cosmwasm-launchpad: Expose `SigningCosmWasmClient.fees`.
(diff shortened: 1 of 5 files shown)


webmaster128 (approved):

webmaster128:
Ah, could you add a CHANGELOG entry?

Issue #640, opened 2021-01-27
Document how to generate protobuf codec file and use it


ethanfrey:
This allows us to add support in an app for key modules that are not already in @cosmjs/stargate, or any custom chain-specific modules (which will never be in @cosmjs/stargate). This seems quite automated, but no one outside the cosmjs team knows how to do it.

Documenting this is an alternative solution to #636

PR #711, opened 2021-03-16
Add ibc support

### LINKED ISSUE #636: Add ibc messages to codec
I was trying to use stargate.codec to create IBC client messages, but I couldn't find the Msg types.

Looking at [the generation scripts](https://github.com/cosmos/cosmjs/blob/6182791cb82eb48466a6410ac970e5d37ba0f808/packages/stargate/scripts/predefine-proto.sh#L42-L47), I saw that `tx.proto` and in 2 cases `query.proto` were not compiles, although they are present in those directories: https://github.com/cosmos/cosmos-sdk/tree/master/proto/ibc/core/client/v1

Please add those generated codecs, so I can try to create messages.

Also, if it is not too much work, a query extension for client and connections would also be useful (although I guess I could do that outside of the package if the protobuf definitions are there?)

willclarktech:
Part 1 of #636

### DIFF
--- a/packages/stargate/src/queries/ibc.spec.ts
+++ b/packages/stargate/src/queries/ibc.spec.ts
@@ -14,220 +14,224 @@ async function makeClientWithIbc(rpcUrl: string): Promise<[QueryClient & IbcExte
 describe("IbcExtension", () => {
   describe("unverified", () => {
     describe("channel", () => {
-      it("works", async () => {
-        pendingWithoutSimapp();
-        const [client, tmClient] = await makeClientWithIbc(simapp.tendermintUrl);
+      describe("channel", () => {
+        it("works", async () => {
+          pendingWithoutSimapp();
+          const [client, tmClient] = await makeClientWithIbc(simapp.tendermintUrl);
 
-        const response = await client.ibc.unverified.channel(ibcTest.portId, ibcTest.channelId);
-        expect(response.channel).toEqual(ibcTest.channel);
-        expect(response.proofHeight).toBeDefined();
-        expect(response.proofHeight).not.toBeNull();
+          const response = await client.ibc.unverified.channel.channel(ibcTest.portId, ibcTest.channelId);
+          expect(response.channel).toEqual(ibcTest.channel);
+          expect(response.proofHeight).toBeDefined();
+          expect(response.proofHeight).not.toBeNull();
 
-        tmClient.disconnect();
+          tmClient.disconnect();
+        });
+      });
+
+      describe("channels", () => {
+        it("works", async () => {
+          pendingWithoutSimapp();
+          const [client, tmClient] = await makeClientWithIbc(simapp.tendermintUrl);
+
+          const response = await client.ibc.unverified.channel.channels();
+          expect(response.channels).toEqual([ibcTest.identifiedChannel]);
+          expect(response.pagination).toBeDefined();
+          expect(response.pagination).not.toBeNull();
+          expect(response.height).toBeDefined();
+          expect(response.height).not.toBeNull();
+
+          tmClient.disconnect();
+        });
+      });
+
+      describe("connectionChannels", () => {
+        it("works", async () => {
+          pendingWithoutSimapp();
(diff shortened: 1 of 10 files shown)


webmaster128 (commented):
Hmm, is this the way to go? Should @cosmjs/stargate expose codecs it does not use internally?

This might be friendly for starters, but sooner or later users will reach the point where they need https://github.com/cosmos/cosmjs/issues/640.

webmaster128:
What's your experience building the IBC relayer?

willclarktech:
In `ts-relayer` we just have our own codec. This was straightforward for us to do because we already knew how to do it from CosmJS, but it's probably a real pain for newcomers. I have no opinions about what should/shouldn't be exposed here because I'm not very clear on who we're targeting with this exposed codec and what their use cases are (besides us).

ethanfrey:
I think we definitely need #640 as people will have app-specific proto. I am now happy to have the ibc in the relayer as we can iterate on it quickly when I need to adjust it (which happened often).

That said, queries that are expected to be commonly used should be in here and exposed in the StargateClient. or SigningStargateClient. I would be happy if someone took over https://github.com/cosmos/cosmjs/pull/690 (which I paused as added a new message was quite hard with the current setup).

Adding support for the ibc-transfer queries and messages seems very useful, so wallets can easily send a token over ics20 and query the denom behind a hash (you get something like "ibc/E957AD87675BEA6576576" as denom, which needs the DenomTrace query to decypher.

packages/stargate/src/queries/ibc.ts
```diff
@@ -14,76 +17,155 @@ import {
...
-        portId: string,
-        channelId: string,
-        sequence: number,
-      ) => Promise<QueryPacketCommitmentResponse>;
-      readonly packetCommitments: (
-        portId: string,
-        channelId: string,
-        paginationKey?: Uint8Array,
-      ) => Promise<QueryPacketCommitmentsResponse>;
+    readonly channel: {
```
willclarktech, line 66: Do we need any more verified queries at this time?

packages/stargate/src/queries/ibc.spec.ts
```diff
@@ -14,220 +14,562 @@ async function makeClientWithIbc(rpcUrl: string): Promise<[QueryClient & IbcExte
...
+        xit("works", async () => {
+          pendingWithoutSimapp();
+          const [client, tmClient] = await makeClientWithIbc(simapp.tendermintUrl);
+
+          const response = await client.ibc.unverified.channel.consensusState(
+            ibcTest.portId,
+            ibcTest.channelId,
+            // TODO: Find valid values
+            0,
+            0,
```
willclarktech, line 111: I'm not sure which values I need here. Do I need to define them dynamically? @ethanfrey

packages/stargate/src/queries/ibc.spec.ts
```diff
@@ -14,220 +14,562 @@ async function makeClientWithIbc(rpcUrl: string): Promise<[QueryClient & IbcExte
...
+      });
+
+      describe("stateTm", () => {
+        it("works", async () => {
+          pendingWithoutSimapp();
+          const [client, tmClient] = await makeClientWithIbc(simapp.tendermintUrl);
+
+          const response = await client.ibc.unverified.client.stateTm(ibcTest.clientId);
+          expect(response.chainId).toEqual("ibc-1");
+          // TODO: Fill these expectations out
```
willclarktech, line 429: This result has a lot of properties. Do we need to assert against all of them?

webmaster128: I think it is fine like this. Good to have the test in place. When someone is working on this particular call or we run into issues, we can fill out the test body more and more.

webmaster128 (approved):

packages/stargate/src/queries/ibc.ts
willclarktech: See discussion here: https://github.com/cosmos/cosmjs/issues/713

PR #716, opened 2021-03-18
Deprecate legacy tendermint-rpc Client class

### LINKED ISSUE #708: Deprecate Client from @cosmjs/tendermint-rpc
`Client` is hardly needed as user either use @cosmjs/launchpad (which does not connect to Tendermint) or can use `Tendermint34Client`.

willclarktech:
Closes #708


webmaster128 (commented):
Could you also add a CHANGELOG entry?

webmaster128 (approved):

PR #718, opened 2021-03-22
Fix normalizePubkey test

webmaster128:
Before, the unit under test (normalizePubkey) was not called.


willclarktech (approved):
Nice catch

Issue #673, opened 2021-02-11
Support for offchain multisig account


samepant:
### Context
I'm working on a browser based off-chain aggregation multisig tool and hoping to use cosmjs to make it happen. I built a prototype using the gaia-cli (rip) multisig tool last fall and am now working on a stargate compatible web version. 

The tool creates an account from multiple pubkeys, creates transactions from this account and then broadcasts them once it's signed by the members. Unfortunately, it's a client requirement to support this offline multisig construction.

After investigating cosmjs, it appears the library does not support multisigs in this way (mentioned here #416 ). 

### Questions
- How possible would it be to support this type of multisig construction?
- How best can I contribute to this feature?
	- With a little bit of guidance, I'd love to help out and grow my cosmos knowledge 🌞

🍃 Thanks for reading

ethanfrey:
Hi @somepant, I love the idea of off-chain signature aggregation. If you have any demos (even working with launchpad), I would be happy to take a look. But this is a personal interest more than a CosmJS blessing.

There was a large change in how one can aggregate multiple signatures in Stargate. In particular, read [ADR 20](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-020-protobuf-transaction-encoding.md), especially look at [SIGN_MODE_DIRECT and the use of AuthInfo](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-020-protobuf-transaction-encoding.md#signing). 

Basically, you must now know upfront who is signing and in which order and what mode (ledger uses `SIGN_MODE_LEGACY_AMINO` currently). This makes such aggregation much more difficult and why we did not include it as a basic use case in stargate (while it is tested in launchpad).

There is discussion of ["future work" to make this easier](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-020-protobuf-transaction-encoding.md#future-improvements). Both `SIGN_MODE_DIRECT_AUX` and `SIGN_MODE_DIRECT_RELAXED` would make your job considerably easier, and I would be happy with both of them as well.

Please raise an [issue on Cosmos SDK](https://github.com/cosmos/cosmos-sdk/issues) and maybe jump on the discord to raise some discussion on this point.

webmaster128:
[...]

Issue #657, opened 2021-02-10
Copy shared functionality from @cosmjs/launchpad


webmaster128:
We now see CosmJS based Stargate clients using imports like this:

```ts
import { coins, GasPrice } from "@cosmjs/launchpad";
import { SigningStargateClient } from "@cosmjs/stargate";
import { DirectSecp256k1HdWallet } from "@cosmjs/proto-signing";
```

But ideally, `@cosmjs/stargate` was entirely independent of `@cosmjs/launchpad`. Especially users of stargate should not be required to import launchpad packages to keep caller code clean. In order to avoid duplicating functionality, I suggest pulling out a shared package `@cosmjs/cosmos`. What I think should be in there is:

- [x] `coin`/`coins`/`Coin`/`parseCoins` (#721)
- [x] `GasPrice`/`CosmosFeeTable`/`GasLimits`/`buildFeeTable`
- [x] fee things?
- [x] `makeCosmoshubPath`
- [x] pubkey and address things (#719)
- [x] `AccountData` for `OfflineDirectSigner`
- [x] logs module (or parts of it) (#721)

willclarktech:
How confident are we that these types/functions will be applicable unchanged to whatever follows Stargate?

What about simply re-exporting the `@cosmjs/launchpad` code from `@cosmjs/stargate` so that consumers of the latter don't need to know anything about the former but we also don't duplicate functionality? Then if/when we drop launchpad support we can just copy the code across, but this is an internal change that doesn't affect the `@cosmjs/stargate` API.

webmaster128:
Right, very good idea. Let's copy them right away and work towards `@cosmjs/stargate` being independent of `@cosmjs/launchpad`.

webmaster128:
Most likely we cannot complete this as long as there is amino signing in stargate. But we can do important steps in the right direction.

willclarktech:
@webmaster128 Good to close?

PR #719, opened 2021-03-22
Add address derivation support for MultisigThreshold public keys

webmaster128:
This is the address derivation for multisig accounts for #673

Also resolves _pubkey and address things?_ from #657


packages/amino/src/encoding.ts
```diff
@@ -0,0 +1,127 @@
...
+ * Uvarint encoder for Amino. This is the same encoding as `binary.PutUvarint` from the Go
+ * standard library.
+ *
+ * @see https://github.com/tendermint/go-amino/blob/8e779b71f40d175/encoder.go#L77-L85
+ */
+function encodeUvarint(value: number | string): number[] {
+  const checked = Uint53.fromString(value.toString()).toNumber();
+  if (checked > 127) {
+    throw new Error(
+      "Encoding numbers > 127 is not supported here. Please tell those lazy CosmJS maintainers to port the binary.PutUvarint implementation from the Go standard library and write some tests.",
```
willclarktech, line 90: 😄

packages/amino/src/encoding.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Encodes a public key to binary Amino.
+ */
+export function encodeAminoPubkey(pubkey: Pubkey): Uint8Array {
+  if (isMultisigThresholdPubkey(pubkey)) {
+    const out = Array.from(pubkeyAminoPrefixMultisigThreshold);
+    out.push(8); // TODO: What is this?
+    out.push(...encodeUvarint(pubkey.value.threshold));
+    for (const pubkeyData of pubkey.value.pubkeys.map((p) => encodeAminoPubkey(p))) {
+      out.push(18); // TODO: What is this?
```
willclarktech, line 105: Should we use `0x12` style here?

packages/amino/src/encoding.spec.ts
```diff
@@ -0,0 +1,163 @@
...
+  testgroup2Address,
+  testgroup3,
+  testgroup3Address,
+  testgroup4,
+  testgroup4Address,
+} from "./testutils.spec";
+
+describe("encoding", () => {
+  describe("encodeSecp256k1Pubkey", () => {
+    it("encodes a compresed pubkey", () => {
```
willclarktech, line 24: Typo: "compressed"

packages/amino/src/multisig.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+    it("return < 0 for left < right", () => {
+      expect(compareArrays(new Uint8Array([5, 5, 4]), new Uint8Array([5, 5, 5]))).toBeLessThan(0);
+      expect(compareArrays(new Uint8Array([5, 4, 5]), new Uint8Array([5, 5, 5]))).toBeLessThan(0);
+      expect(compareArrays(new Uint8Array([4, 5, 5]), new Uint8Array([5, 5, 5]))).toBeLessThan(0);
+      expect(compareArrays(new Uint8Array([5, 5]), new Uint8Array([5, 5, 5]))).toBeLessThan(0);
+      expect(compareArrays(new Uint8Array([5]), new Uint8Array([5, 5, 5]))).toBeLessThan(0);
+      expect(compareArrays(new Uint8Array([]), new Uint8Array([5, 5, 5]))).toBeLessThan(0);
+    });
```
willclarktech, line 28: Would be good to have cases like `[4,5,5]` vs `[5,5,4]`, ie is it left or right precedence? Also `[6]` vs `[5,5]`, ie is magnitude or length more important?

packages/amino/src/multisig.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+      expect(createMultisigThresholdPubkey([test1, test2, test3], 2, false)).toEqual(testgroup1);
+      expect(createMultisigThresholdPubkey([test1, test2, test3], 1, false)).toEqual(testgroup2);
+      expect(createMultisigThresholdPubkey([test3, test1], 2, false)).toEqual(testgroup3);
+    });
+
+    it("works with nosort", () => {
+      expect(createMultisigThresholdPubkey([test3, test1], 2, true)).toEqual(testgroup4);
+    });
+
+    it("throws for threshold lager than number of keys", () => {
```
willclarktech, line 63: Typo: "larger"

packages/amino/src/multisig.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+      expect(() => createMultisigThresholdPubkey([test1, test2, test3], 5)).toThrowError(
```
willclarktech, line 64: Recommend using 4 to catch off-by-one errors.

packages/amino/src/multisig.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Compare arrays lexicographically.
+ *
+ * Returns value < 0 if `a < b`.
+ * Returns value > 0 if `a > b`.
+ * Returns 0 if `a === b`.
+ */
+export function compareArrays(a: Uint8Array, b: Uint8Array): number {
+  return toHex(a).localeCompare(toHex(b));
```
willclarktech, line 15: Interesting. Can we make the locale explicit?

packages/amino/src/encoding.ts
webmaster128: Absolutely. Especially since I did not realize that this is the same value is in the signature's hex dump before your comment. Thanks!

packages/amino/src/multisig.ts
webmaster128: yeah, hex encoding preserves ordering because the ASCII value of 0-9 is smaller than the one of a-f.

Locale is a good point. I will check.

packages/launchpad/src/index.ts
```diff
@@ -1,7 +1,18 @@
...
+  encodeSecp256k1Pubkey,
+  pubkeyToAddress,
+  pubkeyType,
+  SinglePubkey as PubKey,
+} from "@cosmjs/amino";
+
 import * as logs from "./logs";
 export { logs };
 
-export { pubkeyToAddress, rawSecp256k1PubkeyToAddress } from "./address";
```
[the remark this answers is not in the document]
webmaster128: I removed it because it was a shortcut that does not fit in the more general flow. We only have rawSecp256k1PubkeyTo**Raw**Address now. Have a look at the CHANGELOG entry.

webmaster128:
Good, added a few commits starting with _Use locale independent sorting_

packages/amino/src/multisig.spec.ts
```diff
@@ -16,6 +16,12 @@ describe("multisig", () => {
       expect(compareArrays(new Uint8Array([5, 5, 5]), new Uint8Array([5, 5]))).toBeGreaterThan(0);
       expect(compareArrays(new Uint8Array([5, 5, 5]), new Uint8Array([5]))).toBeGreaterThan(0);
       expect(compareArrays(new Uint8Array([5, 5, 5]), new Uint8Array([]))).toBeGreaterThan(0);
+
+      // left or right precedence
+      expect(compareArrays(new Uint8Array([5, 5, 4]), new Uint8Array([4, 5, 5]))).toBeGreaterThan(0);
+
+      // magnitude is more imporant than length
```
willclarktech, line 23: Typo: "important" (also below)

packages/launchpad/src/index.ts
```diff
@@ -1,7 +1,18 @@
...
+  decodeAminoPubkey,
+  decodeBech32Pubkey,
+  encodeAminoPubkey,
+  encodeBech32Pubkey,
+  encodeSecp256k1Pubkey,
+  pubkeyToAddress,
+  pubkeyType,
+  SinglePubkey as PubKey,
+} from "@cosmjs/amino";
+
```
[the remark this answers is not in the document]
willclarktech: Maybe not directly in the export statement, but it's at least possible like this:
```ts
import { SinglePubkey } from "@cosmjs/amino";
/** @deprecated some notice */
export type Pubkey = SinglePubkey;
```

willclarktech (approved):
Just a couple of docstring typos and one comment on the deprecation notice.

PR #721, opened 2021-03-23
Reorganise logs/coins for Launchpad/Stargate

willclarktech:
Part of #657


webmaster128 (approved):
Nice!

I think fee can follow the exact same way

PR #727, opened 2021-03-24
Reorganise Amino SignDoc and OfflineSigner

willclarktech:
Part of #657


packages/proto-signing/package.json
```diff
@@ -44,7 +44,6 @@
   },
   "dependencies": {
     "@cosmjs/amino": "^0.25.0-alpha.0",
-    "@cosmjs/launchpad": "^0.25.0-alpha.0",
```
webmaster128, line 47: 🎉

packages/amino/src/signer.ts
```diff
@@ -1,6 +1,6 @@
-import { StdSignature } from "@cosmjs/amino";
-
-import { StdSignDoc } from "./encoding";
+/* eslint-disable @typescript-eslint/naming-convention */
```
webmaster128, line 1: Why is this newly added here?

packages/launchpad/src/signingcosmosclient.ts
```diff
@@ -1,13 +1,12 @@
 /* eslint-disable @typescript-eslint/naming-convention */
+import { AminoMsg, Coin, makeSignDoc, OfflineAminoSigner, StdFee } from "@cosmjs/amino";
+import {} from "@cosmjs/amino/build/signdoc";
```
webmaster128, line 3: What happened here?

packages/amino/src/signer.ts
willclarktech: Thanks, `StdSignDoc` was here for a part of the transition.

packages/launchpad/src/signingcosmosclient.ts
willclarktech: Thanks, this is a remnant of autocomplete when I'd forgotten to export something.

webmaster128 (approved):
Could you rebase this to ensure it works in well together with my PR?

PR #733, opened 2021-03-25
Add common staking methods to SigningStargateClient

### LINKED ISSUE #687: Add support for staking operations on SigningComsosClient
And ideally on SigningCosmWasmClient.

Beyond sendTokens, there are 3 very common operations exposed on almost all wallets (and some block explorers):

* stake tokens
* unstake tokens
* withdraw rewards

We should support those 3 (and maybe re-delegate) inside CosmJS to make it as useful for basic use-cases without forcing them all to compile ts-proto from scratch (needed for any custom modules).

We have some code on the pre-`ts-proto` format that imports objects from the `codec` inside cosmjs. That approach seems to be deprecated and before adding all the ts-proto codegen, it would be nice to just see that inside CosmJS. I can share the equivalent code with protobufjs.

@abefernan maybe can add more

willclarktech:
Closes #690
Closes #687

### DIFF
--- a/packages/stargate/src/signingstargateclient.ts
+++ b/packages/stargate/src/signingstargateclient.ts
@@ -207,41 +207,41 @@ export class SigningStargateClient extends StargateClient {
   }
 
   public async delegateTokens(
-    senderAddress: string,
+    delegatorAddress: string,
     validatorAddress: string,
     amount: Coin,
     memo = "",
   ): Promise<BroadcastTxResponse> {
     const delegateMsg = {
       typeUrl: "/cosmos.staking.v1beta1.MsgDelegate",
-      value: MsgDelegate.fromPartial({ delegatorAddress: senderAddress, validatorAddress, amount }),
+      value: MsgDelegate.fromPartial({ delegatorAddress: delegatorAddress, validatorAddress, amount }),
     };
-    return this.signAndBroadcast(senderAddress, [delegateMsg], this.fees.delegate, memo);
+    return this.signAndBroadcast(delegatorAddress, [delegateMsg], this.fees.delegate, memo);
   }
 
   public async undelegateTokens(
-    senderAddress: string,
+    delegatorAddress: string,
     validatorAddress: string,
     amount: Coin,
     memo = "",
   ): Promise<BroadcastTxResponse> {
     const undelegateMsg = {
       typeUrl: "/cosmos.staking.v1beta1.MsgUndelegate",
-      value: MsgUndelegate.fromPartial({ delegatorAddress: senderAddress, validatorAddress, amount }),
+      value: MsgUndelegate.fromPartial({ delegatorAddress: delegatorAddress, validatorAddress, amount }),
     };
-    return this.signAndBroadcast(senderAddress, [undelegateMsg], this.fees.undelegate, memo);
+    return this.signAndBroadcast(delegatorAddress, [undelegateMsg], this.fees.undelegate, memo);
   }
 
   public async withdrawRewards(
-    senderAddress: string,
+    delegatorAddress: string,
     validatorAddress: string,
     memo = "",
   ): Promise<BroadcastTxResponse> {
     const withdrawMsg = {
       typeUrl: "/cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward",
-      value: MsgWithdrawDelegatorReward.fromPartial({ delegatorAddress: senderAddress, validatorAddress }),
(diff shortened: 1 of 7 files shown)


packages/stargate/src/signingstargateclient.ts
```diff
@@ -198,6 +206,44 @@ export class SigningStargateClient extends StargateClient {
...
+  public async delegateTokens(
+    senderAddress: string,
+    validatorAddress: string,
+    amount: Coin,
+    memo = "",
+  ): Promise<BroadcastTxResponse> {
+    const delegateMsg = {
+      typeUrl: "/cosmos.staking.v1beta1.MsgDelegate",
+      value: MsgDelegate.fromPartial({ delegatorAddress: senderAddress, validatorAddress, amount }),
+    };
```
webmaster128, line 218: Is there a type-safe way to create messages like this? I mean binding the URL to the value type? Bit I guess we'd have to create interfaces for each message type for this. Any thoughts?

packages/stargate/src/signingstargateclient.ts
```diff
@@ -198,6 +206,44 @@ export class SigningStargateClient extends StargateClient {
...
+  ): Promise<BroadcastTxResponse> {
+    const undelegateMsg = {
+      typeUrl: "/cosmos.staking.v1beta1.MsgUndelegate",
+      value: MsgUndelegate.fromPartial({ delegatorAddress: senderAddress, validatorAddress, amount }),
+    };
+    return this.signAndBroadcast(senderAddress, [undelegateMsg], this.fees.undelegate, memo);
+  }
+
+  public async withdrawRewards(
+    senderAddress: string,
```
webmaster128, line 236: I'd call this `delegatorAddress`, which is signer but not sender as in token transfer source.

packages/stargate/src/signingstargateclient.ts
```diff
@@ -198,6 +206,44 @@ export class SigningStargateClient extends StargateClient {
...
+  ): Promise<BroadcastTxResponse> {
+    const delegateMsg = {
+      typeUrl: "/cosmos.staking.v1beta1.MsgDelegate",
+      value: MsgDelegate.fromPartial({ delegatorAddress: senderAddress, validatorAddress, amount }),
+    };
+    return this.signAndBroadcast(senderAddress, [delegateMsg], this.fees.delegate, memo);
+  }
+
+  public async undelegateTokens(
+    senderAddress: string,
```
webmaster128, line 223: delegatorAddress as below

packages/stargate/src/signingstargateclient.ts
```diff
@@ -198,6 +206,44 @@ export class SigningStargateClient extends StargateClient {
     return this.signAndBroadcast(senderAddress, [sendMsg], this.fees.send, memo);
   }
 
+  public async delegateTokens(
+    senderAddress: string,
```
webmaster128, line 210: delegatorAddress as below

packages/stargate/src/signingstargateclient.ts
willclarktech: We could do something like this?
```ts
type RegistryMap = {
  "/cosmos.bank.v1beta1.MsgMultiSend": MsgMultiSend;
  "/cosmos.distribution.v1beta1.MsgFundCommunityPool": MsgFundCommunityPool;
  // ...
};

interface NotJustAny<K extends keyof RegistryMap> {
  typeUrl: K;
  value: RegistryMap[K];
}
```

webmaster128: Is this better than a decentral type system like this?

```ts
interface EncodeMsgDelegate extends EncodeObject {
  readonly typeUrl: "/cosmos.staking.v1beta1.MsgDelegate",
  readonly value: MsgDelegate,
}
```

willclarktech: Not really. Do we want this overhead?

webmaster128: We have code like this a bunch of times already:

```ts
        const msg = MsgDelegate.fromPartial({
          delegatorAddress: faucet.address0,
          validatorAddress: validator.validatorAddress,
          amount: coin(1234, "ustake"),
        });
        const msgAny = {
          typeUrl: "/cosmos.staking.v1beta1.MsgDelegate",
          value: msg,
        };
```

I'm a bit concerned about typos in those type URLs or URL/type mismatches.

But we can leave it like it is right now. Don't have a strong opinion on this.

willclarktech: Type safety is always nice: https://github.com/cosmos/cosmjs/issues/734

webmaster128 (approved):

PR #735, opened 2021-03-25
Make clients easier to extend

### LINKED ISSUE #696: Make it easier to extend in external apps
The use of `private` everywhere makes it easier for you to refactor and keep an API boundary, but does cause some annoyance while trying to customize usage in other packages. In particular, I wanted to make a custom Client in an external repo that supports some more message types, as https://github.com/cosmos/cosmjs/pull/690 is hanging, at it would be good to add such functionality without depending on the CosmJS core team (this is even more critical for all other wallet/app creators). I also wanted to adjust things like `getBalance` to use the unverified variant (so it didn't take so long to list tokens).

So, I tried to start doing it as follows: 

```ts
import { SigningCosmWasmClient } from "@cosmjs/cosmwasm-stargate";

// This wraps the singing client to support extra messages
// and allow us to customize things like query delay
export class SigningClient extends SigningCosmWasmClient {

}
```

But got the error:

```
Cannot extend a class '"/home/ethan/js/wallet/node_modules/@cosmjs/cosmwasm-stargate/build/signingcosmwasmclient".SigningCosmWasmClient'. Class constructor is marked as private.ts(2675)
```

I'd love your opinions here, or maybe a CosmJS dev could take a look at https://github.com/confio/wallet/pull/70 and see if they have an idea on how to approach this *without touching the cosmjs repo*. We should let app devs customize quite a bit, they will want to.

willclarktech:
Closes #696

### DIFF
--- a/packages/cosmwasm-stargate/src/cosmwasmclient.ts
+++ b/packages/cosmwasm-stargate/src/cosmwasmclient.ts
@@ -42,13 +42,13 @@ import { setupWasmExtension, WasmExtension } from "./queries";
 
 /** Use for testing only */
 export interface PrivateCosmWasmClient {
-  readonly tmClient: Tendermint34Client;
-  readonly queryClient: QueryClient & AuthExtension & BankExtension & WasmExtension;
+  readonly tmClient: Tendermint34Client | undefined;
+  readonly queryClient: (QueryClient & AuthExtension & BankExtension & WasmExtension) | undefined;
 }
 
 export class CosmWasmClient {
-  private readonly tmClient: Tendermint34Client;
-  private readonly queryClient: QueryClient & AuthExtension & BankExtension & WasmExtension;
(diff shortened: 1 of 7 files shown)


packages/cosmwasm-stargate/src/cosmwasmclient.ts
```diff
@@ -166,11 +192,11 @@ export class CosmWasmClient {
   }
 
   public disconnect(): void {
-    this.tmClient.disconnect();
+    this.forceGetTmClient().disconnect();
```
webmaster128, line 195: Can we unify this with `packages/stargate/src/stargateclient.ts`? There we have 

```ts
  public disconnect(): void {
    if (this.tmClient) this.tmClient.disconnect();
  }
```

for no reason. I think both options are okay.

webmaster128 (approved):

PR #741, opened 2021-03-30
Harmonise getAccount methods

### LINKED ISSUE #713: Harmonize getAccount inconsistency
We started Starget development assuming you can get all data in a light-client-verified way. It turned out that this is not possible for all queries and even the simple ones where you can get a proof, you often do not want to wait for one more block in order to get the result.

So in some places, unverified `getAccount` queries were added. Right now, we have a wild mix.

1. `StargateClient` in `packages/stargate/src/stargateclient.ts` has `getAccount` and `getAccountUnverified`; `StargateClient.getSequence` still uses the verified version.
2. `SigningStargateClient` in `packages/stargate/src/signingstargateclient.ts` uses unverified `getAccountUnverified`.
3. `CosmWasmClient` in `packages/cosmwasm-stargate/src/cosmwasmclient.ts` only has the verified `getAccount`.
4. `SigningCosmWasmClient` in `packages/cosmwasm-stargate/src/signingcosmwasmclient.ts` uses verified `getAccount`.

Given @ethanfrey's changes, I think it makes most sense to make `getAccount` unverified in both cases and add an explicit `getAccountVerified` to both.

willclarktech:
Closes #713

The other query extensions will need to be rearranged too. I started doing that here, but I think
it’ll be best in separate PRs.

### DIFF
--- a/packages/stargate/src/stargateclient.ts
+++ b/packages/stargate/src/stargateclient.ts
@@ -177,14 +177,17 @@ export class StargateClient {
     return account ? accountFromAny(account) : null;
   }
 
-  public async getSequence(address: string): Promise<SequenceResponse | null> {
+  public async getSequence(address: string): Promise<SequenceResponse> {
     const account = await this.getAccount(address);
-    return account
-      ? {
-          accountNumber: account.accountNumber,
-          sequence: account.sequence,
-        }
-      : null;
+    if (!account) {
+      throw new Error(
+        "Account does not exist on chain. Send some tokens there before trying to query sequence.",
+      );
+    }
+    return {
+      accountNumber: account.accountNumber,
+      sequence: account.sequence,
+    };
   }
 
   public async getBlock(height?: number): Promise<Block> {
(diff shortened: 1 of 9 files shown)


packages/stargate/src/stargateclient.ts
```diff
@@ -160,29 +160,31 @@ export class StargateClient {
...
-      };
-    } else {
-      return null;
-    }
+    return account
+      ? {
+          accountNumber: account.accountNumber,
+          sequence: account.sequence,
+        }
+      : null;
```
webmaster128, line 187: I'd make this math the Launchpad behaviour. Otherwise this method does not add anything to `getAccount`.

```ts
  /**
   * Returns account number and sequence.
   *
   * Throws if the account does not exist on chain.
   *
   * @param address returns data for this address. When unset, the client's sender adddress is used.
   */
  public async getSequence(address: string): Promise<GetSequenceResult> {
    const account = await this.getAccount(address);
    if (!account) {
      throw new Error(
        "Account does not exist on chain. Send some tokens there before trying to query sequence.",
      );
    }
    return {
      accountNumber: account.accountNumber,
      sequence: account.sequence,
    };
  }
```

webmaster128 (approved):

PR #749, opened 2021-03-31
Add IBC MsgTransfer

### LINKED ISSUE #639: Add MsgTransfer (coin transfers via IBC) to codecs and implement Amino JSON conversion
Requires #638 

This is a [key message used by ibctransfer](https://github.com/cosmos/cosmos-sdk/blob/master/proto/ibc/applications/transfer/v1/tx.proto) to initiate an ics20 coin transfer. This will be a key usage for many apps with ibc.

Also, they just added amino signing for `MsgTransfer`. We should add support for that as well, so cosmjs -> ledger works for this key ibc operation

willclarktech:
Closes #639

I'm not particularly confident about the `uint64` fields since we haven't added any of those yet.

### DIFF
--- a/packages/stargate/src/aminomsgs.ts
+++ b/packages/stargate/src/aminomsgs.ts
@@ -18,7 +18,7 @@ export interface AminoMsgSend extends AminoMsg {
 }
 
 export function isAminoMsgSend(msg: AminoMsg): msg is AminoMsgSend {
-  return (msg as AminoMsgSend).type === "cosmos-sdk/MsgSend";
+  return msg.type === "cosmos-sdk/MsgSend";
 }
 
 interface Input {
@@ -43,7 +43,7 @@ export interface AminoMsgMultiSend extends AminoMsg {
 }
 
 export function isAminoMsgMultiSend(msg: AminoMsg): msg is AminoMsgMultiSend {
-  return (msg as AminoMsgMultiSend).type === "cosmos-sdk/MsgMultiSend";
+  return msg.type === "cosmos-sdk/MsgMultiSend";
 }
 
 // crisis - see https://github.com/cosmos/cosmos-sdk/blob/efa73c7/proto/cosmos/crisis/crisis.proto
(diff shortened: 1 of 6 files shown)


packages/stargate/src/aminomsgs.ts
```diff
@@ -331,7 +329,38 @@ export interface AminoMsgUndelegate extends AminoMsg {
...
 }
 
 // upgrade (no messages) - see https://github.com/cosmos/cosmos-sdk/blob/efa73c7/proto/cosmos/upgrade/upgrade.proto
+
+// ibc
+
+// https://github.com/cosmos/ibc-go/blob/07b6a97b67d17fd214a83764cbdb2c2c3daef445/modules/core/02-client/types/client.pb.go#L297-L312
+interface AminoHeight {
+  readonly revision_number: number;
+  readonly revision_height: number;
```
webmaster128, line 342: For this type, the JSON serialization becomes a string: https://github.com/cosmos/cosmos-sdk/blob/v0.42.3/x/auth/legacy/legacytx/stdsign.go#L21-L22. Since there is no annotation for that, I assume uin64 -> string is the standard

willclarktech: Hm, I was basing my choice on this: https://github.com/cosmos/cosmjs/blob/main/packages/stargate/src/aminomsgs.ts#L181

willclarktech: But maybe that's a bug!

webmaster128: It is very well possible that this is a bug.

We have in the LCD API:

```
export interface GovProposerResponse {
  readonly height: string;
  readonly result: {
    readonly proposal_id: string;
    readonly proposer: string;
  };
}
```

and in genesis

```
      "starting_proposal_id": "1",
```

willclarktech: Any idea where the definitive resource for deciding this question might be?

webmaster128: This is implemented here: https://github.com/tendermint/go-amino/blob/master/json-encode.go#L100-L113

webmaster128 (approved):

PR #750, opened 2021-04-01
Add SigningStargateClient.sendIbcTokens

### LINKED ISSUE #725: Add SigningCosmosClient.sendIbcTokens
Create a method similar to `SigningCosmosClient.sendTokens` but for ICS 20 token transters (via MsgTransfer). This requires adding [MsgTransfer](https://github.com/cosmos/cosmos-sdk/blob/v0.42.2/proto/ibc/applications/transfer/v1/tx.proto#L16-L40) to the generated code.

(If you don't want to wait for CosmJS, this can be done in external applications as well. Give it a try and share your experience!)

Requires #639

willclarktech:
Closes #725

Opening as a draft because I'm not sure what the best testing approach is here. Do we want a full integration test with two chains? Just one chain (and if so how do we set up a channel)?

### DIFF
--- a/packages/stargate/src/signingstargateclient.ts
+++ b/packages/stargate/src/signingstargateclient.ts
@@ -13,6 +13,7 @@ import {
 } from "@cosmjs/proto-signing";
 import { Tendermint34Client } from "@cosmjs/tendermint-rpc";
 import { assert } from "@cosmjs/utils";
+import Long from "long";
 
 import { AminoTypes } from "./aminotypes";
 import { MsgMultiSend } from "./codec/cosmos/bank/v1beta1/tx";
@@ -45,6 +46,7 @@ import {
   MsgTimeout,
   MsgTimeoutOnClose,
 } from "./codec/ibc/core/channel/v1/tx";
+import { Height } from "./codec/ibc/core/client/v1/client";
 import {
   MsgCreateClient,
   MsgSubmitMisbehaviour,
@@ -76,6 +78,7 @@ export const defaultGasPrice = GasPrice.fromString("0.025ucosm");
 export const defaultGasLimits: GasLimits<CosmosFeeTable> = {
   send: 80_000,
   delegate: 160_000,
+  transfer: 160_000,
   undelegate: 160_000,
   withdraw: 160_000,
 };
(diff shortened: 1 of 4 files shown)


packages/stargate/src/signingstargateclient.ts
```diff
@@ -241,11 +252,43 @@ export class SigningStargateClient extends StargateClient {
...
+    transferAmount: Coin,
+    sourcePort: string,
+    sourceChannel: string,
+    timeoutHeight?: Height,
+    /** timeout in seconds */
+    timeoutTimestamp?: number,
+    memo = "",
+  ): Promise<BroadcastTxResponse> {
+    const timeoutTimestampNanoseconds = timeoutTimestamp
+      ? Long.fromNumber(timeoutTimestamp * 1_000_000_000)
```
webmaster128, line 275: This easily exceeds `MAX_SAFE_INTEGER`:

```
> const timestamp = (Date.now()/1000)
undefined
> (timestamp*1_000_000_000) / Number.MAX_SAFE_INTEGER
179.5539240475597
```

Let's convert `timeoutTimestamp` to string, append "000000000" and use Long.fromString

packages/stargate/src/signingstargateclient.ts
```diff
@@ -241,11 +252,43 @@ export class SigningStargateClient extends StargateClient {
...
 
+  public async sendIbcTokens(
+    senderAddress: string,
+    recipientAddress: string,
+    transferAmount: Coin,
+    sourcePort: string,
+    sourceChannel: string,
+    timeoutHeight?: Height,
+    /** timeout in seconds */
+    timeoutTimestamp?: number,
```
webmaster128, line 271: Should we make those required as `Height | undefined` and `number | undefined`? It seems to be dangerous to leave them out.

packages/stargate/src/signingstargateclient.ts
ethanfrey: What about `Long.fromNumber(timeoutTimestamp).multiply(1_000_000_000)`?

willclarktech: @bartmacbartek This also applies to ts-relayer:

https://github.com/confio/ts-relayer/blob/main/src/lib/ibcclient.ts#L1348

packages/stargate/src/signingstargateclient.ts
ethanfrey: The server will reject any packet that has both undefined. At least one must be set (it is valid to set both as well)

packages/stargate/src/signingstargateclient.ts
webmaster128: > What about `Long.fromNumber(timeoutTimestamp).multiply(1_000_000_000)`?

This is what you do when you know the Long API :)

packages/stargate/src/signingstargateclient.ts
webmaster128: Good. Let's force the user to explicitly set them to undefined instead of skipping the arguments.

webmaster128 (approved):

PR #755, opened 2021-04-06
Rearrange HD wallet options

### LINKED ISSUE #736: Request for Bip 39 passphrase functionality
DirectSecp256k1HdWallet.fromMnemonic should take a passphrase parameter as well, it is used in creating bip39.mnemonicToSeed.

willclarktech:
Closes #736

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,11 @@ and this project adheres to
 - @cosmjs/cosmwasm-stargate: Add `SigningCosmWasmClient.offline` static method
   for constructing offline clients without a Tendermint client.
 - @cosmjs/stargate: Add `SigningStargateClient.sendIbcTokens` method.
+- @cosmjs/amino: Export `Secp256k1HdWalletOptions` interface.
+- @cosmjs/amino: Add `bip39Password` option to `Secp256k1HdWallet` options.
+- @cosmjs/proto-signing: Export `DirectSecp256k1HdWalletOptions` interface.
+- @cosmjs/proto-signing: Add `bip39Password` option to `DirectSecp256k1HdWallet`
+  options.
 
 ### Changed
 
@@ -100,6 +105,10 @@ and this project adheres to
   account is not found, instead of returning null.
 - @cosmjs/cosmwasm-stargate: `CosmWasmClient.getBalance` now returns a 0 balance
   instead of null.
+- @cosmjs/amino: Options for `Secp256k1HdWallet.fromMnemonic` are now passed via
+  a `Secp256k1HdWalletOptions` object.
+- @cosmjs/proto-signing: Options for `DirectSecp256k1HdWallet.fromMnemonic` are
+  now passed via a `DirectSecp256k1HdWalletOptions` object.
 
 ### Deprecated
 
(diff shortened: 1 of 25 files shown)


CHANGELOG.md
```diff
@@ -100,6 +105,10 @@ and this project adheres to
   account is not found, instead of returning null.
 - @cosmjs/cosmwasm-stargate: `CosmWasmClient.getBalance` now returns a 0 balance
   instead of null.
+- @cosmjs/amino: Options for `Secp256k1HdWallet` are now passed via a
```
webmaster128, line 108:
```suggestion
- @cosmjs/amino: Options for `Secp256k1HdWallet.fromMnemonic` are now passed via a
```

CHANGELOG.md
```diff
@@ -100,6 +105,10 @@ and this project adheres to
...
+  `Secp256k1HdWalletOptions` object.
+- @cosmjs/proto-signing: Options for `DirectSecp256k1HdWallet` are now passed
```
webmaster128, line 110:
```suggestion
- @cosmjs/proto-signing: Options for `DirectSecp256k1HdWallet.fromMnemonic` are now passed
```

packages/amino/src/secp256k1hdwallet.ts
```diff
@@ -107,6 +107,12 @@ interface DerivationInfo {
   readonly prefix: string;
 }
 
+export interface Secp256k1HdWalletOptions {
+  readonly bip39Password?: string;
+  readonly hdPath?: HdPath;
+  readonly prefix?: string;
+}
```
webmaster128, line 114: What about making the fields required and using `Partial<Secp256k1HdWalletOptions>`? Then we can use `Secp256k1HdWalletOptions` for places where all options are set (e.g. after replacing unset with defaults).

packages/amino/src/secp256k1hdwallet.ts
```diff
@@ -107,6 +107,12 @@ interface DerivationInfo {
   readonly prefix: string;
 }
 
+export interface Secp256k1HdWalletOptions {
+  readonly bip39Password?: string;
+  readonly hdPath?: HdPath;
```
webmaster128, line 112: Now that users have to specify the argument name, we could simplify this to just `path`. But I don't have a strong opinion on this one.

willclarktech: We use `hdPath` in a bunch of other places. I think it could get confusing.

packages/amino/src/secp256k1hdwallet.ts
willclarktech: Updated. I'm not sure how you wanted to handle the `bip39Password` option though: I left it optional, but maybe you prefer the type to be `string | undefined`, or we could even give it an empty string as default since it's just appended to the salt.

webmaster128: I would default to an empty string because this is how the algorithm is defined:

> To create a binary seed from the mnemonic, we use the PBKDF2 function with a mnemonic sentence (in UTF-8 NFKD) used as the password and the string "mnemonic" + passphrase (again in UTF-8 NFKD) used as the salt.

https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#From_mnemonic_to_seed

For the implementation `undefined` and `""` are synonyms.

webmaster128 (approved):

PR #757, opened 2021-04-06
Add address helpers to tendermint-rpc

### LINKED ISSUE #753: Add helpers to generate Tendermint addresses from pubkeys
Right now, users have to calculate Tendermint addresses manually. This is shown in https://github.com/cosmos/cosmjs/wiki/What-can-CosmJS-do-for-me%3F#deriving-a-tendermint-address-from-a-consensus-pubkey.

Here is how Tendermint addresses are calculated: https://github.com/informalsystems/tendermint-rs/blob/v0.18.1/tendermint/src/account.rs#L93-L111. This seems to be the same derivation as Cosmos SDK but those will likely diverge in the future ([ADR-028](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-028-public-key-addresses.md) and use of signature aggregation in Tendermint).

It would be good to have a `pubkeyToAddress` and `pubkeyToRawAddress` in @cosmjs/tendermint-rpc that supports both Ed25519 and Secp256k1 inputs.

willclarktech:
Closes #753

Opening as a draft because I wasn’t sure about a few things:

- What should the output format be? Just uppercase hex?
- What exactly is the input expected to be?
- Should multisig pubkeys be included?
- Should there be a dependency between @cosmjs/amino and @cosmjs/tendermint-rpc? If so in which
  direction?
- Are there any test vectors I can use?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ and this project adheres to
 - @cosmjs/proto-signing: Export `DirectSecp256k1HdWalletOptions` interface.
 - @cosmjs/proto-signing: Add `bip39Password` option to `DirectSecp256k1HdWallet`
   options.
+- @cosmjs/amino: Add `rawEd25519PubkeyToRawAddress` helper function.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,9 @@ and this project adheres to
 - @cosmjs/proto-signing: Add `bip39Password` option to `DirectSecp256k1HdWallet`
   options.
 - @cosmjs/amino: Add `rawEd25519PubkeyToRawAddress` helper function.
+- @cosmjs/tendermint-rpc: Add `pubkeyToAddress`, `pubkeyToRawAddress`,
+  `rawEd25519PubkeyToRawAddress`, and `rawSecp256k1PubkeyToRawAddress` helper
+  functions.
 
 ### Changed
 
--- a/packages/amino/src/addresses.ts
+++ b/packages/amino/src/addresses.ts
@@ -6,6 +6,13 @@ import { Bech32, fromBase64 } from "@cosmjs/encoding";
 import { encodeAminoPubkey } from "./encoding";
 import { isEd25519Pubkey, isMultisigThresholdPubkey, isSecp256k1Pubkey, Pubkey } from "./pubkeys";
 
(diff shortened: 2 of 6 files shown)


packages/amino/src/pubkeys.ts
```diff
@@ -67,5 +67,5 @@ export interface MultisigThresholdPubkey extends Pubkey {
 }
 
 export function isMultisigThresholdPubkey(pubkey: Pubkey): pubkey is MultisigThresholdPubkey {
-  return (pubkey as MultisigThresholdPubkey).type === "tendermint/PubKeyMultisigThreshold";
+  return pubkey.type === "tendermint/PubKeyMultisigThreshold";
```
webmaster128, line 70: Why did you remove those casts? They ensure we don't have a typo in the check. I know this is strange, but casting to the check type is the documented way of docing such checks.

packages/tendermint-rpc/src/pubkeys.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface Pubkey {
+  // type is one of the strings defined in pubkeyType
+  // I don't use a string literal union here as that makes trouble with json test data:
+  // https://github.com/cosmos/cosmjs/pull/44#pullrequestreview-353280504
+  readonly type: string;
+  readonly value: any;
+}
+
+export interface Ed25519Pubkey extends Pubkey {
+  readonly type: "tendermint/PubKeyEd25519";
```
webmaster128, line 12: Those type identifiers are Amino specific and not needed anymore 🎉 The pubkeys are now just defined as Ed25519 or Secp256k1: https://github.com/tendermint/tendermint/blob/v0.34.8/proto/tendermint/crypto/keys.proto#L8-L17

This means we can use a simpler API here, like:

```
export function pubkeyToRawAddress(type: "ed25519" | "secp256k1", data: Uint8Array): Uint8Array
```

Then users can come with all their different JSON dumps or ptotobuf files and bring it in this simple form.

Sorry for pretending that reusing the Amino `Pubkey` types is a good idea.

packages/amino/src/pubkeys.ts
willclarktech: Oh! I just thought they were unnecessary. Makes sense though.

webmaster128: If I was a compiler, I would optimize that out at compile time as always true. But TypeScript does not do that and does not seem to have plans to change that.

webmaster128 (approved):
Great, thanks

PR #758, opened 2021-04-07
Improve broadcastTx methods

### LINKED ISSUE #739: Adapt CosmWasm client to use sync mode not block
From discord: We need to adapt CosmWasm to use sync (not block) mode, and then subscribe/poll for the results. This will handle the case when it takes a few blocks to get in.

When the sentry accepts 0 gas, but no validator does, it will pass sync but never make it into a block, ever. An annoying UX issue with these subjective min gas prices.

willclarktech:
Part 1 of #739

~I'm not sure how to test this. Opening as a draft to get feedback before I do the same for
`CosmWasmClient`.~


packages/stargate/src/stargateclient.ts
```diff
@@ -268,7 +270,34 @@ export class StargateClient {
...
+    timeoutMs = 60_000,
+    pollIntervalMs = 3_000,
+  ): Promise<BroadcastTxResponse> {
+    return new Promise((resolve, reject) => {
+      const rejectTimeout = setTimeout(reject, timeoutMs, `timeout of ${timeoutMs}ms exceeded`);
+      this.broadcastTxCommit(tx)
+        .catch((error) => {
+          try {
+            const errorJson = JSON.parse(error.message);
+            if (errorJson.code !== -32603) {
```
willclarktech, line 284: Compare with the check here: https://github.com/tendermint/vue/blob/develop/packages/vuex/src/modules/common/relayers/lib/starportSigningClient.js#L21-L24
I'm not sure if that code could be from other causes, but matching the string seems undesirable if there's a nice numerical code to use.

packages/stargate/src/stargateclient.ts
```diff
@@ -1,11 +1,13 @@
 /* eslint-disable @typescript-eslint/naming-convention */
+import { sha256 } from "@cosmjs/crypto";
 import { toHex } from "@cosmjs/encoding";
 import { Uint53 } from "@cosmjs/math";
 import {
   broadcastTxCommitSuccess,
   Tendermint34Client,
   toRfc3339WithNanoseconds,
 } from "@cosmjs/tendermint-rpc";
+import { sleep } from "../../utils/build";
```
webmaster128, line 10: This should be imported via the package

packages/stargate/src/stargateclient.ts
```diff
@@ -268,7 +270,34 @@ export class StargateClient {
...
+        .catch((error) => {
+          try {
+            const errorJson = JSON.parse(error.message);
+            if (errorJson.code !== -32603) {
+              throw error;
+            }
+            // timed out waiting for tx to be included in a block
+            // now we poll to artificially extend the timeout
+            const txHash = toHex(sha256(tx));
+            return this.pollTxById(txHash, pollIntervalMs);
```
webmaster128, line 290: Will this recursive poll stack ever end after the timeout? If I understand correctly this continues in the background.

packages/stargate/src/stargateclient.ts
```diff
@@ -268,7 +270,34 @@ export class StargateClient {
...
   }
 
-  public async broadcastTx(tx: Uint8Array): Promise<BroadcastTxResponse> {
+  public async broadcastTx(
+    tx: Uint8Array,
+    timeoutMs = 60_000,
+    pollIntervalMs = 3_000,
+  ): Promise<BroadcastTxResponse> {
+    return new Promise((resolve, reject) => {
+      const rejectTimeout = setTimeout(reject, timeoutMs, `timeout of ${timeoutMs}ms exceeded`);
```
webmaster128, line 279: I would create a nice error type (like TimeoutError) that apps can handle. It can create the transaction ID, which we can calculate locally from `tx` (hopefully). Then an app can say: _hey, we submitted your transaction XYZ but did not find it on chain yet. You might want to check manually or come back later._ and store the transaction ID in some local storage.

packages/stargate/src/stargateclient.ts
willclarktech: Good point!

ethanfrey:
> I'm not sure how to test this.

The issue happens when the tendermint rpc timeout expired before the tx makes it in a block.

The only way I can think to test it in CI is to either make a new chain (better than mangling an existing one) as `slow-network`. Set block commit timeout to say 10s and the rpc timeout to 1s. 90% of the time the rpc should timeout before it is included in a block. If you have a few tests that rely on this, you can check. 

To be more deterministic, you can poll every 0.5s until a new block is out, then submit a tx, knowing this will have 9.5-10s before it is included in a block. Then try the following:

* Submit tx with larger timeout (20s) and ensure we get the answer as if there was no timeout (like other code)
* Submit tx with small timeout (3s) and ensure we get an error. Wait 10s and query with the tx id in the error and ensure we find it.

All these timeout values can be set in `config.toml`

packages/stargate/src/stargateclient.ts
clockworkgr: error code: -32603 is "Internal Error" and CAN be from causes other than the timeout AFAIK.

Hence the check on both in my original implementation.

packages/stargate/src/stargateclient.ts
clockworkgr: The reason I had the original timeout error re-thrown was because it is essentially the "same" error (regardless of whether rpc timed out or polling did) and maybe people already had their own flow of handling that error/exception and didn't want to break that flow with a new exception. YMMV

packages/stargate/src/stargateclient.ts
```diff
@@ -268,7 +279,61 @@ export class StargateClient {
     if (this.tmClient) this.tmClient.disconnect();
   }
 
-  public async broadcastTx(tx: Uint8Array): Promise<BroadcastTxResponse> {
+  public async broadcastTx(
+    tx: Uint8Array,
+    timeoutMs = 60_000,
+    pollIntervalMs = 3_000,
+  ): Promise<BroadcastTxResponse> {
+    const txId = toHex(sha256(tx));
```
webmaster128, line 287: Can make this upper case? This is the typical representation in Cosmos (see https://cosmos.bigdipper.live/transactions)

packages/stargate/src/stargateclient.ts
```diff
@@ -268,7 +279,61 @@ export class StargateClient {
...
+    return new Promise((resolve, reject) =>
+      this.broadcastTxCommit(tx)
+        .catch((error) => {
+          try {
+            const errorJson = JSON.parse(error.message);
+            if (
+              // -32603 is "Internal Error"
+              // https://github.com/tendermint/tendermint/blob/v0.34.0/rpc/jsonrpc/types/types.go#L236
+              errorJson.code === -32603 &&
+              /timed out waiting for tx to be included in a block/i.test(errorJson.data)
```
webmaster128, line 321: If this check ever becomes a problem we can switch to non-committing boradcasts (sync) and start the polling right away.

packages/stargate/src/testutils.spec.ts
```diff
@@ -54,6 +64,17 @@ export const simapp = {
...
 };
 
+export const slowSimapp = {
+  tendermintUrl: "localhost:26660",
+  tendermintUrlWs: "ws://localhost:26660",
+  tendermintUrlHttp: "http://localhost:26660",
+  chainId: "simd-testing",
+  denomStaking: "ustake",
+  denomFee: "ucosm",
+  blockTime: 1_000, // ms
```
webmaster128, line 74: This looks wrong

.circleci/config.yml
```diff
@@ -106,6 +106,10 @@ jobs:
           name: Start simapp
           command: ./scripts/simapp/start.sh
           background: true
+      - run:
+          name: Start slow simapp
+          command: ./scripts/simapp/start_slow.sh
```
webmaster128, line 111: I would name-prefix the new scripts: `slow_start.sh`, `slow_stop.sh`

webmaster128 (commented):
💪

packages/stargate/src/stargateclient.ts
willclarktech: Updated in line with https://github.com/cosmos/cosmjs/issues/739#issuecomment-818587639

webmaster128 (approved):

PR #763, opened 2021-04-14
Add gasUsed to Stargate broadcast responses

### LINKED ISSUE #761: Tendermint: Return gas used
Follow-up of https://github.com/cosmos/cosmjs/pull/675 but let's only implement it for `Tendermint34Client`

willclarktech:
Closes #761
~Merge after #762~

### DIFF
--- a/packages/stargate/src/stargateclient.ts
+++ b/packages/stargate/src/stargateclient.ts
@@ -1,11 +1,7 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import { toHex } from "@cosmjs/encoding";
 import { Uint53 } from "@cosmjs/math";
-import {
-  BroadcastTxSyncResponse,
-  Tendermint34Client,
-  toRfc3339WithNanoseconds,
-} from "@cosmjs/tendermint-rpc";
+import { Tendermint34Client, toRfc3339WithNanoseconds } from "@cosmjs/tendermint-rpc";
 import { sleep } from "@cosmjs/utils";
 
 import { Account, accountFromAny } from "./accounts";
@@ -63,6 +59,8 @@ export interface IndexedTx {
   readonly code: number;
   readonly rawLog: string;
   readonly tx: Uint8Array;
+  readonly gasUsed: number;
+  readonly gasWanted: number;
 }
 
 export interface SequenceResponse {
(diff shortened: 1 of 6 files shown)


packages/stargate/src/stargateclient.ts
```diff
@@ -303,14 +300,17 @@ export class StargateClient {
...
-        : pollForTx(txId);
+        : pollForTx(txId, gasUsed);
     };
 
     return new Promise((resolve, reject) =>
       this.forceGetTmClient()
         .broadcastTxSync({ tx })
-        .then(({ hash }: BroadcastTxSyncResponse) => pollForTx(toHex(hash).toUpperCase()))
+        .then(({ gasUsed, hash }: tendermint34.BroadcastTxSyncResponse) =>
+          pollForTx(toHex(hash).toUpperCase(), gasUsed),
```
webmaster128, line 312: Why do we need to pass the value here from the `sync` response? How can `sync` broadcast even know the value?

Can't we get it from `const result = await this.getTx(txId);`?

packages/tendermint-rpc/src/tendermint34/adaptors/v0-34/responses.ts
```diff
@@ -127,18 +127,24 @@ function decodeEvents(events: readonly RpcEvent[]): readonly responses.Event[] {
 
 interface RpcTxData {
   readonly code?: number;
+  readonly codespace?: string;
```
webmaster128, line 130: Could you move this above code? This is a namespace schema like: `<codespace>/<code>`

packages/stargate/src/stargateclient.ts
```diff
@@ -83,6 +80,7 @@ export interface BroadcastTxSuccess {
   readonly transactionHash: string;
   readonly rawLog?: string;
   readonly data?: readonly MsgData[];
+  readonly gasUsed: number;
```
webmaster128, line 83: Can we add gasWanted here as well? This should be trivial to get now and probably makes some user happy

packages/stargate/src/stargateclient.ts
```diff
@@ -63,6 +59,7 @@ export interface IndexedTx {
   readonly code: number;
   readonly rawLog: string;
   readonly tx: Uint8Array;
+  readonly gasUsed: number;
```
webmaster128, line 62: Can we add gasWanted here as well? This should be trivial to get now and probably makes some user happy

webmaster128 (approved):

PR #769, opened 2021-04-20
Add support for multiple accounts to HD wallets

### LINKED ISSUE #760: Add support for multiple accounts in HD wallets
The signer interface supports multiple accounts per wallet. However, both `Secp256k1HdWallet` and `DirectSecp256k1HdWallet` only support one account. It would be nice if multiple paths could be used (similar to `LedgerSigner`).

If that gets too hard to implement for `Secp256k1HdWallet` without braking the serialization/deserialization, I think it is fine to just implement it for the newer `DirectSecp256k1HdWallet`.

willclarktech:
Closes #760

### DIFF
--- a/packages/amino/src/secp256k1hdwallet.ts
+++ b/packages/amino/src/secp256k1hdwallet.ts
@@ -311,12 +311,10 @@ export class Secp256k1HdWallet implements OfflineAminoSigner {
   ): Promise<string> {
     const dataToEncrypt: Secp256k1HdWalletData = {
       mnemonic: this.mnemonic,
-      accounts: this.accounts.map(
-        (account): DerivationInfoJson => ({
-          hdPath: pathToString(account.hdPath),
-          prefix: account.prefix,
-        }),
-      ),
+      accounts: this.accounts.map(({ hdPath, prefix }) => ({
+        hdPath: pathToString(hdPath),
+        prefix: prefix,
+      })),
     };
     const dataToEncryptRaw = toUtf8(JSON.stringify(dataToEncrypt));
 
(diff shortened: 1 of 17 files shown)


packages/amino/src/secp256k1hdwallet.ts
```diff
@@ -213,11 +211,17 @@ export class Secp256k1HdWallet implements OfflineAminoSigner {
         assert(typeof mnemonic === "string");
         if (!Array.isArray(accounts)) throw new Error("Property 'accounts' is not an array");
         if (accounts.length !== 1) throw new Error("Property 'accounts' only supports one entry");
-        const account = accounts[0];
-        if (!isDerivationJson(account)) throw new Error("Account is not in the correct format.");
+        if (!accounts.every((account) => isDerivationJson(account))) {
+          throw new Error("Account is not in the correct format.");
+        }
+        const firstPrefix = accounts[0].prefix;
+        if (!accounts.every(({ prefix }) => prefix === firstPrefix)) {
```
webmaster128, line 217: Here you could also use .some(). However, I do't think it matters much.

packages/amino/src/secp256k1hdwallet.ts
```diff
@@ -237,58 +240,47 @@ export class Secp256k1HdWallet implements OfflineAminoSigner {
 
   /** Base secret */
   private readonly secret: EnglishMnemonic;
+  /** BIP39 seed */
+  private readonly seed: Uint8Array;
```
webmaster128, line 244: Nice to store this. Deriving this is the slow part.

webmaster128 (commented):
Looks good.

I just have oone big concern: this changes the serialization format on `packages/amino/src/secp256k1hdwallet.ts`. It makes previously serialized wallets unusable. I don't know of any user using this serialization, but I will try to find out if someone does. The minimum we should do is bump the version of the serialization format and have a helpful error messge that explains why deserialization `"secp256k1wallet-v1";` serializations are not supported.

webmaster128 (approved):

PR #773, opened 2021-04-21
Add support for secp256k1 validator pubkeys to tendermint-rpc

### LINKED ISSUE #681: Support for secp256k1 validator pubkeys
There is a comment here to only support ed25519: https://github.com/cosmos/cosmjs/blob/main/packages/tendermint-rpc/src/types.ts#L11-L15

This has been updated in tendermint and they also support secp256k1. Please update the type definitions and the validation in the parsing code to allow both types. (not urgent, just noted while working on the relayer)

willclarktech:
Closes #681

Do we need tests for this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,7 @@ and this project adheres to
   `isMsgClearAdminEncodeObject` etc helpers.
 - @cosmjs/stargate: Add transfer queries codec, as well as transfer query
   methods to IBC query extension.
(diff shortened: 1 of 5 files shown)


webmaster128 (commented):
Nice. I don't think we need tests here. But a CHANGELOG entry would be nice. This is a breaking change in the `ValidatorPubkey` type.

webmaster128 (approved):
Thanks

PR #775, opened 2021-04-22
Add decodeTxRaw

webmaster128:
This change is motivated by this problem which came up a few times before:

Bildschirmfoto 2021-04-22 um 17 57 46

I always wondered why `Tx.decode` worked for the code in the diff because in Tendermint TxRaw is stored. Turns out the decoding works by accident because the tho types have the same binary representation (including their nested fields).

### DIFF
--- a/packages/stargate/src/stargateclient.ts
+++ b/packages/stargate/src/stargateclient.ts
@@ -58,6 +58,20 @@ export interface IndexedTx {
   /** Transaction execution error code. 0 on success. */
   readonly code: number;
   readonly rawLog: string;
+  /**
+   * Raw transaction bytes stored in Tendermint.
+   *
+   * If you hash this, you get the transaction hash (= transaction ID):
+   *
+   * ```js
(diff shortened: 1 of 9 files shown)


packages/stargate/src/stargateclient.ts
```diff
@@ -58,6 +58,20 @@ export interface IndexedTx {
...
+   * If you hash this, you get the transaction hash (= transaction ID):
+   *
+   * ```js
+   * import { sha256 } from "@cosmjs/crypto";
+   * import { toHex } from "@cosmjs/encoding";
+   *
+   * const transactionId = toHex(sha256(indexTx.tx)).toUpperCase();
+   * ```
+   *
+   * Use `decodeTxRaw` fromn @cosmjs/proto-signing to decode this.
```
willclarktech, line 73: Typo: "from"

willclarktech (approved):
Oh wow, nice catch

PR #777, opened 2021-04-25
Replace relative import

vilsbole:
Relative import fails in production.
Prefer import from node_modules


webmaster128 (approved):
Right, relative imports must not be used between different packages. Thank you!

PR #780, opened 2021-04-28
Move prettier config for markdown to config

webmaster128:
With this change, different prettier tools get the chance to produce the same result. This helps us with IDE plugins.

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,4 +1,13 @@
 {
   "printWidth": 110,
-  "trailingComma": "all"
+  "trailingComma": "all",
+  "overrides": [
+    {
+      "files": "*.md",
+      "options": {
+        "printWidth": 80,
+        "proseWrap": "always"
+      }
+    }
+  ]
 }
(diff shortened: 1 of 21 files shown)


.prettierrc.json
```diff
@@ -1,4 +1,13 @@
...
   "printWidth": 110,
-  "trailingComma": "all"
-}
\ No newline at end of file
+  "trailingComma": "all",
+  "overrides": [
+    {
+      "files": "*.md",
+      "options": {
+        "printWidth": 80,
```
willclarktech, line 8: Do we even want this to be 80?

willclarktech (approved):

.prettierrc.json
webmaster128: I don't mind. The setting exists because this is prettier default and we wanted our code lines to be longer than that. In cosmwasm we do not set the value and get 80 by default.

willclarktech: My mild preference would be to use the default value everywhere, but I don't really mind either way.

webmaster128: I think 80 is very short for code. What I used in other projects is 100, which can then work for code and text.

PR #781, opened 2021-04-28
Wasmd 0.16 support

### LINKED ISSUE #656: Find automated solution to path bug from ts-proto/wasm
Currently the import paths in `@cosmjs/cosmwasm-stargate/src/codec/x/wasm/internal/types/{query,tx}.ts` have to be manually adjusted to add an additional `../` to other codec file import paths after being generated by ts-proto. I've not yet investigated where this is coming from.

### LINKED ISSUE #779: Test wasmd 0.16 / cosmwasm 0.14 compatibility
This is the goal for CosmJS 0.25 cosmwasm package.

We have `cosmwasm v0.14.0-beta5` out now and soon a `wasmd v0.16.0-beta`. Those should be close enough to final to be worth upgrading to and debugging. I will post here when relevant releases are out.

ethanfrey:
Closes #779
Closes #656

Don't merge yet. This uses CosmWasm v0.14.0-beta5 contracts and wasmd v0.16.0-alpha2 runtime. We expect final versions of all in the next day or two, but this should be close enough to check for any compatibility issues.

Most minor issues fixed. There remained a proto encoding issue.
I regenerated the proto files, but notably the package changed:

`x/wasm/internal/types` -> `cosmwasm/wasm/v1beta1`

I also updated the `contracts` query that only returns a list of string now. And has pagination (see TODO in code).

Check with @alpe for details on other updates


scripts/wasmd/generate_template.sh
```diff
@@ -47,7 +47,7 @@ function inline_jq() {
   inline_jq "template/.wasmd/config/genesis.json" -S
 
   # Custom settings in config.toml
-  sed -i "" \
+  sed -i"" \
```
webmaster128, line 50: This unfortunately will not work in the mac. Can you use

```
gnused="$(command -v gsed || echo sed)"
```

and

```
"$gnused" -i"" \
```

ethanfrey: You can change it back as well. I remembered this was a trick that once worked on both.

Since I rarely use this repo, you can make it OSX easier.

ethanfrey:
Fixed compile errors, a number of tests still fail.

scripts/wasmd/generate_template.sh
ethanfrey: Reverted.

ethanfrey:
Thanks for fixing this.
I merged your branch in. Feel free to merge this into master if you feel this is finished.

ethanfrey:
I guess we can cut final releases of everything, then another CosmJS PR to update everything to the final releases?

webmaster128 (approved):

packages/cosmwasm-stargate/src/codec/cosmwasm/wasm/v1beta1/query.ts
```diff
@@ -1,97 +1,128 @@
...
   address: string;
   /** QueryData contains the query data passed to the contract */
   queryData: Uint8Array;
 }
 
-/** QuerySmartContractStateResponse is the response type for the Query/SmartContractState RPC method */
+/**
+ * QuerySmartContractStateResponse is the response type for the
+ * Query/SmartContractState RPC method
+ */
```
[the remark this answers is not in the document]
webmaster128: This might be a formatting change of the commoents in the wasmd source files

PR #782, opened 2021-04-29
Increase query client extensions to 18

litvintech:
The current max amount of extensions for queries client is set to 8.
Enabling cosmos-sdk's default/advanced modules and cosmwasm's will fulfill all current slots.
Mostly for cosmos-sdk based computers, we will have 2-10 custom modules.


webmaster128 (approved):
Very nice, thank you

PR #783, opened 2021-05-03
Finalize Wasmd 0.16 support

webmaster128:


ethanfrey (approved):
This was all that was left?

I thought I was getting issues with the soruce and builder flags, but maybe my setup scripts.

PR #788, opened 2021-05-04
Update yarn to v2

### LINKED ISSUE #301: Upgrade to yarn2
I never tried it but it seems like yarn1 is more or less unmaintained and all development effort goes into yarn2.

willclarktech:
Closes #301


docs/YARN.md
```diff
@@ -0,0 +1,294 @@
...
+## Consider using constraints
+
+E.g. to ensure that each package specifies the right fields in its
+`package.json`. Bonus: you get to learn Prolog!
+
+https://yarnpkg.com/features/constraints
+
+## TypeScript plugin
+
+Automatically installs `@types` definitions if the project doesn’t have its own.
```
webmaster128, line 217: Those packages? 👇

```suggestion
Automatically installs `@types/*` definitions if the project doesn’t have its own.
```

docs/YARN.md
```diff
@@ -0,0 +1,294 @@
...
+https://yarnpkg.com/features/pnp#the-node_modules-problem
+
+### Check whether your project is ready
+
+```sh
+yarn dlx @yarnpkg/doctor@2
+```
+
+> NOTE: You need to specify the `@2` because they released the incompatible v3
+> RC as `latest` on npm.
```
webmaster128, line 248: Why format this as quote?

docs/YARN.md
```diff
@@ -0,0 +1,294 @@
...
+
+Replace calls to node outside of a `package.json` script with `yarn node`. For
+example in shebangs in development scripts. (You probably don’t want to update
+executables you will ship to your users because they may not be using Yarn.)
+
+## Switch versioning from Lerna to Yarn
+
+The Lerna team seems to be pretty ambivalent about its future. After a period of
+hibernation, they released v4 on 10th February 2021. But the main `README.md`
+for the project still refers to v3, and
```
webmaster128, line 287: Could we formulate this section more a bit more neutral? I was reading stories about one or two people working on it with one person long gone and the other completely burned out. Let's try to avoid any unnecessary pressure on those people. I think it is safe to simply call it unmaintained after security advisories have not been fixed for months and there is no commit since that one spike in February.

webmaster128 (commented):
Great work!

docs/YARN.md
willclarktech: I also don't like it, but there isn't a good information box in GFMD as far as I can tell. People seem to use either this or a horrible little table.

webmaster128: What about a simple paragraph starting with bold, like this:

**NOTE:** You need to specify the `@2` because they released the incompatible v3 RC as `latest` on npm.

packages/amino/jasmine-testrunner.js
```diff
@@ -36,4 +35,4 @@ const reporter = new SpecReporter({
 // initialize and execute
 jasmine.env.clearReporters();
 jasmine.addReporter(reporter);
-jasmine.execute();
+void jasmine.execute();
```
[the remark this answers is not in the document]
willclarktech: It tells the linter that I didn't handle rejections on purpose.

webmaster128 (approved):
🎉

PR #792, opened 2021-05-05
Fix hdPaths runtime error with proper defaultOptions and options merging

bartmacbartek:


webmaster128 (commented):
Which runtime error does this fix? I don't see the difference created by this PR. Can we create a test that failed before and passes after this change?

packages/proto-signing/src/directsecp256k1hdwallet.ts
```diff
@@ -91,7 +91,11 @@ export class DirectSecp256k1HdWallet implements OfflineDirectSigner {
...
   protected constructor(mnemonic: EnglishMnemonic, options: DirectSecp256k1HdWalletConstructorOptions) {
-    const { seed, hdPaths, prefix } = { ...defaultOptions, ...options };
+    const { seed, hdPaths, prefix } = {
+      seed: options.seed,
+      prefix: options.prefix ?? defaultOptions.prefix,
+      hdPaths: options.hdPaths ?? defaultOptions.hdPaths,
+    };
     this.secret = mnemonic;
     this.seed = seed;
     this.accounts = hdPaths.map((hdPath) => ({
```
bartmacbartek, line 101: options parameter (line 93) type definition equals to:
```ts
export interface DirectSecp256k1HdWalletConstructorOptions {
  readonly bip39Password: string | undefined
  readonly hdPaths: readonly HdPath[] | undefined;
  readonly prefix: string | undefined;
  readonly seed: Uint8Array;
}
```

so it's possible to create a new instance of the `DirectSecp256k1HdWallet` class like this:
```ts
new DirectSecp256k1HdWallet(..., {
    ...
    hdPaths: undefined,
})
```

which results in a runtime TypeError on line 101
> 'Cannot read property \'map\' of undefined',

and that's because of `options` and `defaultOptions` merging with a spread operator (line 95-96):
```ts
{
    ...{
        hdPaths: [makeCosmoshubPath(0)],
    },
    ...{
        hdPaths: undefined,
    }
}
```

The typescript inferred types are misleading. One possible explicit fix is in this PR, however, there are other ways.

**functional like** (incomplete, it needs ugly casting because of filter inference):
```ts
const notUndefinedOptions = Object.fromEntries(
    Object.entries(options).filter(([_key, value]):  => value !== undefined),
)

const { seed, hdPaths, prefix } = {
  ...defaultOptions,
  ...notUndefinedOptions,
}
```

**mutate** (requires removal of `readonly`):
```ts
if (!options.hdPaths) {
  delete options.hdPaths;
}

if (!options.prefix) {
  delete options.prefix;
}

const { seed, hdPaths, prefix } = {
  ...defaultOptions,
  ...options,
}
```

Not sure which one you prefer, I like the explicit one. This is definitely an edge case but still, this is a bug, and it prevented me from a quick cosmjs upgrade to `0.25.0` in `ts-relayer`. I can make a workaround now but it's better to have it fixed anyway.

webmaster128: Great explanation, thank you. Let's go for the explicit way you proposed. No need to rely on even more magic that is hard to review.

packages/proto-signing/src/directsecp256k1hdwallet.ts
```diff
@@ -91,7 +91,11 @@ export class DirectSecp256k1HdWallet implements OfflineDirectSigner {
   private readonly accounts: readonly Secp256k1Derivation[];
 
   protected constructor(mnemonic: EnglishMnemonic, options: DirectSecp256k1HdWalletConstructorOptions) {
-    const { seed, hdPaths, prefix } = { ...defaultOptions, ...options };
+    const { seed, hdPaths, prefix } = {
```
webmaster128, line 94: Could you turn those into 3 simple `const`s? No need for the surrounding object anymore.

webmaster128 (commented):
I'd appreciate a test, otherwise 👍 🎉

packages/proto-signing/src/directsecp256k1hdwallet.ts
webmaster128: So to wrap up what you are saying: In JavaScript we need to watch out for 3 types of null: `null`, `undefined` and not set 🤯

packages/proto-signing/src/directsecp256k1hdwallet.ts
bartmacbartek: 8bed00ed1

test on its way

bartmacbartek:
> I'd appreciate a test, otherwise 👍 🎉

ed020e6

this test fails if you try it without the changes in this PR:
image

webmaster128:
Great, thanks. Let's see what @willclarktech says. We probably have the same problem in different places in the codebase.

packages/proto-signing/src/directsecp256k1hdwallet.ts
bartmacbartek: Yes, exactly! This is odd... 🤯

willclarktech: Ooph. Yes, this probably affects other code. :(

willclarktech: @bartmacbartek Do you know why TypeScript doesn't catch this?

bartmacbartek: I spent some time on this but no idea. I think it's worth a typescript ticket. I reproduced it here: https://www.typescriptlang.org/play?ts=4.3.0-beta#code/MYewdgzgLgBAJgUwGYEMCuAbKB5ADlAS3AgC4YBvAWACgY6YALBDDEM6AJwLAHMBtALo0AvjAC8FGvUbNWZPgHIA7iA4Y4CgQBoRNGqEiwQ+IpDJVa9JixAB+dlC69BI8ZMt1rcmGjCIk3AhwutT6xLBQCNBuFtIAdAn+6Fh4hMQ6HjAJccZpkCE0kdBxXiBxALYouAAUKOIAfDAoAJRAA

willclarktech (approved):

packages/ledger-amino/src/launchpadledger.ts
```diff
@@ -44,13 +44,10 @@ export class LaunchpadLedger {
...
-      ...defaultOptions,
-      ...options,
-    };
-    this.testModeAllowed = testModeAllowed;
-    this.hdPaths = hdPaths;
-    this.prefix = prefix;
+
+    this.testModeAllowed = options.testModeAllowed ?? defaultOptions.testModeAllowed;
+    this.hdPaths = options.hdPaths ?? defaultOptions.hdPaths;
+    this.prefix = options.prefix ?? cosmosBech32Prefix;
```
webmaster128, line 50: With this change, `defaultOptions.prefix` is unused. Let's either put all defaults here or collect all defaults in `defaultOptions`. Both ways are fine for me but I have a slight tendency towards just removing `defaultOptions`

webmaster128 (commented):
Nice. Could you add a CHANGELOG entry the the "Unreleased"/Fixed section and 💅 that code section below? Then this is good to go.

packages/ledger-amino/src/launchpadledger.ts
dontmakemethink: just my mistake: 8042c252e

bartmacbartek: just my mistake: 8042c25

Issue #800, opened 2021-05-12
bug: New broadcast logic hides errors


clockworkgr:
The latest implementation of broadcastTx that polls the chain for the tx result hides all pre-commit errors.

https://github.com/cosmos/cosmjs/blob/main/packages/stargate/src/stargateclient.ts#L329

This will return a timeout error for any kind of invalid transaction including insufficient fees. etc.

webmaster128:
This would explain the bahaviour described in https://github.com/confio/ts-relayer/issues/180

clockworkgr:
ah..never noticed that issue....same thing happened to us and I just worked back through the code and found it :)

fwiw, ibc-setup costs about 2.2m gas on each chain

webmaster128:
Fix is in #805. Happy for any 👀 on it.

webmaster128:
Released as 0.25.3

PR #802, opened 2021-05-13
Remove named capture groups from regexes

AlexBHarley:
Documented a little bit of my investigation in [this](https://github.com/cosmos/cosmjs/issues/801) ticket. Basically we need Hermes compatibility as we're building a React Native application and are using this library.

All the tests pass on my machine 👍 and as a side note I learned that the plural of regex is not in fact regices 🤷‍♂️

### DIFF
--- a/packages/launchpad/src/fee.ts
+++ b/packages/launchpad/src/fee.ts
@@ -13,11 +13,11 @@ export class GasPrice {
   }
 
   public static fromString(gasPrice: string): GasPrice {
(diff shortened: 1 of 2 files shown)


webmaster128:
Thank you for the report and the fix. I opened a slightly different patch in #803 based on your work. It increases the test coverage before doing the change and also handles the general pattern matching and the delegation to the Decimal parser slightly differently. We can release this as a patch release together with #800 next week.

PR #804, opened 2021-05-14
Use built in bip39

AlexBHarley:
`bip39.mnemonicToSeed` is already async so no need for the funkiness of wrapping `pbkdf2 `

### DIFF
--- a/packages/crypto/src/bip39.ts
+++ b/packages/crypto/src/bip39.ts
@@ -1,8 +1,5 @@
-import { fromHex, toHex, toUtf8 } from "@cosmjs/encoding";
+import { fromHex, toHex } from "@cosmjs/encoding";
 import * as bip39 from "bip39";
-import { pbkdf2 } from "pbkdf2";
-import * as unorm from "unorm";
-
 import { EnglishMnemonic } from "./englishmnemonic";
 
 export class Bip39 {
@@ -36,30 +33,6 @@ export class Bip39 {
   }
 
   public static async mnemonicToSeed(mnemonic: EnglishMnemonic, password?: string): Promise<Uint8Array> {
-    // reimplementation of bip39.mnemonicToSeed using the asynchronous
-    // interface of https://www.npmjs.com/package/pbkdf2
-    const mnemonicBytes = toUtf8(unorm.nfkd(mnemonic.toString()));
-    const salt = "mnemonic" + (password ? unorm.nfkd(password) : "");
-    const saltBytes = toUtf8(salt);
-    return this.pbkdf2(mnemonicBytes, saltBytes, 2048, 64, "sha512");
-  }
-
-  // convert pbkdf2's callback interface to Promise interface
-  private static async pbkdf2(
-    secret: Uint8Array,
-    salt: Uint8Array,
-    iterations: number,
-    keylen: number,
-    digest: string,
-  ): Promise<Uint8Array> {
-    return new Promise<any>((resolve, reject) => {
-      pbkdf2(secret, salt, iterations, keylen, digest, (err, derivedKey) => {
-        if (err) {
-          reject(err);
-        } else {
-          resolve(new Uint8Array(derivedKey));
-        }
-      });
-    });
+    return new Uint8Array(await bip39.mnemonicToSeed(mnemonic.toString(), password));
   }
 }
--- a/packages/crypto/src/bip39.ts
+++ b/packages/crypto/src/bip39.ts
@@ -1,5 +1,6 @@
 import { fromHex, toHex } from "@cosmjs/encoding";
 import * as bip39 from "bip39";
+
 import { EnglishMnemonic } from "./englishmnemonic";
 
 export class Bip39 {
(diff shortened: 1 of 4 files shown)


webmaster128:
Nice find, thanks.

The re-plementation there was built for version 2.5.0 of the bip39 package 3 years ago where the use of async pbkdf2 led to a massive speedup (https://github.com/iov-one/iov-core/pull/80).

AlexBHarley:
Makes sense. I'm trying to hack around bip39 being dog slow on React Native for Android. I have my own NativeModule written to wrap it, but cosmjs is doing it's own thing.

webmaster128:
i.e. you override the `bip39` module in your build? Could you somehow inject the subtle object? Then pdkdf2 [would pick it up](https://github.com/crypto-browserify/pbkdf2/blob/v3.1.2/lib/async.js#L9).

packages/crypto/src/bip39.ts
```diff
@@ -1,8 +1,5 @@
-import { fromHex, toHex, toUtf8 } from "@cosmjs/encoding";
+import { fromHex, toHex } from "@cosmjs/encoding";
 import * as bip39 from "bip39";
-import { pbkdf2 } from "pbkdf2";
-import * as unorm from "unorm";
-
```
webmaster128, line 5: Lint is complaining because imports should be in two blocks separated by an empty line: external and local.

packages/proto-signing/src/directsecp256k1hdwallet.ts
```diff
@@ -243,7 +243,7 @@ export class DirectSecp256k1HdWallet implements OfflineDirectSigner {
   /** Derivation instructions */
   private readonly accounts: readonly Secp256k1Derivation[];
 
-  protected constructor(mnemonic: EnglishMnemonic, options: DirectSecp256k1HdWalletConstructorOptions) {
+  constructor(mnemonic: EnglishMnemonic, options: DirectSecp256k1HdWalletConstructorOptions) {
```
webmaster128, line 246: Why this (unrelated?) change?

packages/crypto/src/bip39.ts
```diff
@@ -1,8 +1,5 @@
-import { fromHex, toHex, toUtf8 } from "@cosmjs/encoding";
+import { fromHex, toHex } from "@cosmjs/encoding";
 import * as bip39 from "bip39";
-import { pbkdf2 } from "pbkdf2";
-import * as unorm from "unorm";
```
webmaster128, line 4: Could you remove the now unused dependencies `unorm` and `pbkdf2` from `packages/crypto/package.json`?

webmaster128: Also `@types/pbkdf2` and `@types/unorm` 🔥

packages/proto-signing/src/directsecp256k1hdwallet.ts
AlexBHarley: I want to manually construct an instance, not be forced to go via `fromMnemonic`. Going via `fromMnemonic` calls your custom BIP39 implementation.

Tbh this is not totally necessary with these changes, but I have a ts-ignore allowing it on my end for now.

AlexBHarley:
Yes that's definitely another way to do it. Will likely revisit at some point

packages/proto-signing/src/directsecp256k1hdwallet.ts
```diff
@@ -243,7 +243,7 @@ export class DirectSecp256k1HdWallet implements OfflineDirectSigner {
   /** Derivation instructions */
   private readonly accounts: readonly Secp256k1Derivation[];
 
-  protected constructor(mnemonic: EnglishMnemonic, options: DirectSecp256k1HdWalletConstructorOptions) {
+  public constructor(mnemonic: EnglishMnemonic, options: DirectSecp256k1HdWalletConstructorOptions) {
```
webmaster128, line 246: Looking at this constructur again, this is easy to use in a wrong way by providing a seed that does not match the mnemonic. Let's keep this protected as it was before.

If you want to make this public in a type-safe way, it is easier to derive your own wallet type:

```ts
class MyWallet extends DirectSecp256k1HdWallet {
  public constructor(yourOptions) {
    super(mnemonic, options);    
  }
}
```

This should be more powerful, type safe and allow merging this PR without further CosmJS API refinement.

yarn.lock
```diff
@@ -7984,11 +7972,6 @@ universalify@^2.0.0:
   resolved "https://registry.yarnpkg.com/universalify/-/universalify-2.0.0.tgz#75a4984efedc4b08975c5aeb73f530d02df25717"
   integrity sha512-hAZsKq7Yy11Zu1DE0OzWjw7nnLZmJZYTDZZyEFHZdUhV8FkH5MCfoU1XMaxXovpyW5nq5scPqq0ZDP9Zyl04oQ==
 
-unorm@^1.5.0:
```
webmaster128, line 7987: Turns out this functionality is available natively now :) https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize

webmaster128 (approved):

PR #805, opened 2021-05-17
Fix error propagation for checkTx errors

### LINKED ISSUE #800: bug: New broadcast logic hides errors

webmaster128:
Closes #800

### DIFF
--- a/packages/cosmwasm-stargate/src/cosmwasmclient.ts
+++ b/packages/cosmwasm-stargate/src/cosmwasmclient.ts
@@ -207,6 +207,17 @@ export class CosmWasmClient {
     if (this.tmClient) this.tmClient.disconnect();
   }
 
+  /**
+   * Broadcasts a signed transaction to the network and monitors its inclusion in a block.
+   *
+   * If broadcasting is rejected by the node for some reason (e.g. because of a CheckTx failure),
+   * an error is thrown.
+   *
+   * If the transaction is not included in a block before the provided timeout, this errors with a `TimeoutError`.
+   *
+   * If the transaction is included in a block, a `BroadcastTxResponse` is returned. The caller then
+   * usually needs check for execution success or failure.
+   */
   // NOTE: This method is tested against slow chains and timeouts in the @cosmjs/stargate package.
   // Make sure it is kept in sync!
   public async broadcastTx(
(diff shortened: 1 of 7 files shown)


packages/cosmwasm-stargate/src/cosmwasmclient.ts
```diff
@@ -240,10 +251,15 @@ export class CosmWasmClient {
...
+      throw new Error(
+        `Broadcasting transaction failed with code ${broadcasted.code} (codespace: ${broadcasted.codeSpace}). Log: ${broadcasted.log}`,
+      );
+    }
+    const transactionId = toHex(broadcasted.hash).toUpperCase();
     return new Promise((resolve, reject) =>
-      this.forceGetTmClient()
-        .broadcastTxSync({ tx })
-        .then(({ hash }) => pollForTx(toHex(hash).toUpperCase()))
+      pollForTx(transactionId)
```
webmaster128, line 262: I'm still unsure if the use of

```
        .then(resolve, reject)
        .finally(() => clearTimeout(txPollTimeout)),
```

is correct. Shoudn't the call of `resolve` or `resolve` be the last thing that happens here?

willclarktech: Happy to move the `clearTimeout` into the `then` handlers to make it clearer.

packages/stargate/src/stargateclient.ts
```diff
@@ -99,6 +99,15 @@ export interface BroadcastTxSuccess {
   readonly gasWanted: number;
 }
 
+/**
+ * The response after sucessfully broadcasting a transaction.
```
willclarktech, line 103:
```suggestion
 * The response after successfully broadcasting a transaction.
```

packages/stargate/src/stargateclient.ts
```diff
@@ -292,6 +301,17 @@ export class StargateClient {
...
+  /**
+   * Broadcasts a signed transaction to the network and monitors its inclusion in a block.
+   *
+   * If broadcasting is rejected by the node for some reason (e.g. because of a CheckTx failure),
+   * an error is thrown.
+   *
+   * If the transaction is not included in a block before the provided timeout, this errors with a `TimeoutError`.
+   *
+   * If the transaction is included in a block, a `BroadcastTxResponse` is returned. The caller then
+   * usually needs check for execution success or failure.
```
willclarktech, line 313:
```suggestion
   * usually needs to check for execution success or failure.
```

willclarktech (approved):

packages/cosmwasm-stargate/src/cosmwasmclient.ts
webmaster128: Done in 33ff475. This changes the order of execution such that nothing happens after the resolve/reject call.

PR #810, opened 2021-05-25
Support for `/block_search` RPC endpoint 

RiccardoM:
This PR adds the support to the `/block_search` RPC endpoint that was introduced with [PR #6626](https://github.com/tendermint/tendermint/pull/6226), made available with [`v0.34.9`](https://github.com/tendermint/tendermint/blob/master/CHANGELOG.md#features) and is already available inside the Cosmos SDK with [`v0.42.4`](https://github.com/cosmos/cosmos-sdk/blob/master/CHANGELOG.md#improvements-1).

This endpoint is needed to allow to search for block events that are not emitted by any transaction and are found only inside the `begin_block_events` or `end_block_events` arrays. Without this endpoint it is not possible to search for such events.

### DIFF
--- a/packages/tendermint-rpc/src/tendermint34/tendermint34client.spec.ts
+++ b/packages/tendermint-rpc/src/tendermint34/tendermint34client.spec.ts
@@ -204,6 +204,72 @@ function defaultTestSuite(rpcFactory: () => RpcClient, expected: ExpectedValues)
     });
   });
 
+  describe("blockSearch", () => {
+    const key = randomString();
+
+    beforeAll(async () => {
+      if (tendermintEnabled()) {
+        const client = await Tendermint34Client.create(rpcFactory());
+
+        // eslint-disable-next-line no-inner-declarations
+        async function sendTx(): Promise<void> {
+          const me = randomString();
+          const tx = buildKvTx(key, me);
+
+          const txRes = await client.broadcastTxCommit({ tx: tx });
+          expect(responses.broadcastTxCommitSuccess(txRes)).toEqual(true);
+          expect(txRes.height).toBeTruthy();
+          expect(txRes.hash.length).not.toEqual(0);
+        }
+
+        // send 3 txs
+        await sendTx();
+        await sendTx();
+        await sendTx();
+
+        client.disconnect();
+
+        await tendermintSearchIndexUpdated();
+      }
+    });
+
+    it("can paginate over txSearch results", async () => {
+      pendingWithoutTendermint();
+      const client = await Tendermint34Client.create(rpcFactory());
+
+      const query = buildQuery({ raw: "block.height >= 1 AND block.height <= 5" });
+
+      // expect one page of results
+      const s1 = await client.blockSearch({ query: query, page: 1, per_page: 2 });
+      expect(s1.totalCount).toEqual(5);
+      expect(s1.blocks.length).toEqual(2);
+
+      // second page
+      const s2 = await client.blockSearch({ query: query, page: 2, per_page: 2 });
+      expect(s2.totalCount).toEqual(5);
+      expect(s2.blocks.length).toEqual(2);
+
+      client.disconnect();
+    });
+
+    it("can get all search results in one call", async () => {
+      pendingWithoutTendermint();
+      const client = await Tendermint34Client.create(rpcFactory());
+
+      const query = buildQuery({ raw: "block.height >= 1 AND block.height <= 3" });
+
+      const sall = await client.blockSearchAll({ query: query, per_page: 2 });
+      expect(sall.totalCount).toEqual(5);
+      expect(sall.blocks.length).toEqual(5);
+      // make sure there are in order from lowest to highest height
+      const [b1, b2, b3] = sall.blocks;
+      expect(b2.block.header.height).toEqual(b1.block.header.height + 1);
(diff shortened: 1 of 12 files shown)


packages/tendermint-rpc/src/tendermint34/tendermint34client.ts
```diff
@@ -98,6 +98,47 @@ export class Tendermint34Client {
     return this.doCall(query, this.p.encodeBlockResults, this.r.decodeBlockResults);
   }
 
+  /**
+   * Search for events that are in a block
+   *
+   * @see https://docs.tendermint.com/master/rpc/#/Info/block_search
+   */
+  public async blockSearch(params: requests.BlockSearchParams): Promise<responses.BlockSearchResponse> {
```
ethanfrey, line 106: This looks good to me (in terms of functionality).

My one question is that this is only available in tendermint v0.34.9+, correct? Minimally that should be documented in the comments along with a link to the PR. However, what happens when someone tries to run this on tendermint v0.34.8 node? There should be a clearly defined error case that distinguishes "network error" from "tendermint version too old" or it will be very hard for people to debug this. 

Maybe @webmaster128 has some idea how to do this properly - it seems this error would be thrown in `this.doCall`

ethanfrey (commented):
I agree this is an important addition and the code looks good to my eyes. (Simon is the final reviewer there however).

My big question is documentation and displaying errors well, so a client could decide to handle the case of an older chain.

packages/tendermint-rpc/src/tendermint34/tendermint34client.ts
webmaster128: I think a note regarding compatibility here would be nice.

I would not get into the version checking business again, which led to a lot of trouble sttarting with Tendermint 0.34. If the backend does not support an RPC method, it will return some kind of error that explains the problem somehow.

webmaster128 (commented):
Thank you four your contribution.

Could you write a test for this? I assume it is tricky to get values via plain Tendermint but at least it should be possible to send a request and parse and empty result. This would most likely require an update in `scripts/tendermint/all_start.sh` and `scripts/tendermint/all_stop.sh` from 0.34.0 to a sufficiently new version.

RiccardoM:
I've tried updating the Tendermint version to `v0.34.9` (the latest one) as @webmaster128 suggested. However, I found out that all the Tendermint versions starting from `v0.34.4` do not return a valid version, which results in the `adaptorForVersion` method to throw an exception (which makes fail all the tests). 

I've opened a Tendermint issue here: tendermint/tendermint#6488

webmaster128:
Thank you for exploring this, @RiccardoM. Much appreciated. In https://github.com/cosmos/cosmjs/pull/813 I pulled out the upgrade of the local Tendermint version for testing. When we have this merged, you can rebase your PR and have 0.34.10 available.

webmaster128:
#813 is merged. Could you rebase this PR on top of current main?

RiccardoM:
@webmaster128 I've rebased the PR into the `main` branch. In order for all the tests to pass I had to slightly edit the response parsers as they were assuming things that are not always true. Particularly, if you query block with `height = 1` (the genesis block), the following happens: 

- `block.header.last_block_id.hash` is empty
- `block.header.last_block_id.parts.hash` is empty
- `block.header.app_hash` is empty

So I had to remove the associated `assertNotEmpty` from the code

packages/tendermint-rpc/src/tendermint34/tendermint34client.spec.ts
```diff
@@ -204,6 +204,72 @@ function defaultTestSuite(rpcFactory: () => RpcClient, expected: ExpectedValues)
...
+
+        await tendermintSearchIndexUpdated();
+      }
+    });
+
+    it("can paginate over txSearch results", async () => {
+      pendingWithoutTendermint();
+      const client = await Tendermint34Client.create(rpcFactory());
+
+      const query = buildQuery({ raw: "block.height >= 1 AND block.height <= 5" });
```
webmaster128, line 240: Let's get all query results as in the txSearch and blockSearchAll case

```suggestion
      const query = buildQuery({ raw: "block.height >= 1 AND block.height <= 3" });
```

packages/tendermint-rpc/src/tendermint34/tendermint34client.spec.ts
```diff
@@ -204,6 +204,72 @@ function defaultTestSuite(rpcFactory: () => RpcClient, expected: ExpectedValues)
...
+        await sendTx();
+        await sendTx();
+
+        client.disconnect();
+
+        await tendermintSearchIndexUpdated();
+      }
+    });
+
+    it("can paginate over txSearch results", async () => {
```
webmaster128, line 236:
```suggestion
    it("can paginate over blockSearch results", async () => {
```

webmaster128 (commented):
> Particularly, if you query block with `height = 1` (the genesis block), the following happens:

makes sense, thank you.

RiccardoM:
@webmaster128 Applied all the fixes

webmaster128:
Some of the new tests are failing now. Could you fix those?

Also could you give me edit permission of the branch? Then I would create a CHANGELOG entry and fix up the commit history in order to get the changes from my other PR out of this diff.

PR #820, opened 2021-06-03
Use cosmjs-types

### LINKED ISSUE #799: Move generated types into dedicated repository
Users of CosmJS currently need to generate their own types from proto files if they want to build something non-trivial. This is a burden we can reduce at least as long as common cosmos types are used.

Some of those projects are:
- ts-relayer: https://github.com/confio/ts-relayer/tree/main/src/codec
- interchain-accounts-client: https://github.com/Ryan-Gordon/interchain-accounts-client/tree/master/src/codec
- CosmJS itself

Those types are generated the same way we show and I think it is safe to assume that whatever code generation options we suggest good for the majority of our users. We should make clear that the code generation is opinionated and CosmJS specific.

The types should be compiled from .proto to .ts and from .ts to .d.ts/.js. This makes them much easier to access from JavaScript-only users because they do not need a TypeScript compiler on their end (see also https://github.com/cosmos/cosmjs/issues/778).

I think we should create a monorepo that has different output folders and different npm packages for the namespaces. Typical folders are

```
$ ls src/codec 
confio     cosmos     google     ibc        intertx    tendermint
```

or

```
$ ls src/codec/
cosmos   cosmwasm google
```

which we can publish as `@typesnamespace/confio` etc. I would imagine an import to look like this:

```js
import { MsgSend } from "@typesnamespace/cosmos/bank/v1beta1/tx";

let msg = MsgSend.fromPartial({ /* ... */ });
```

This has a bunch of advantages for CosmJS development:
- No ts-proto installation needed here
- Faster builds by using pre-compiled types
- We can recommend to never import from sub-folder instead of seeing thinks like `import { MsgSend } from "@cosmjs/stargate/build/codec/cosmos/bank/v1beta1/tx";`

The same has been done in [cosmos_sdk_proto](https://docs.rs/cosmos-sdk-proto/0.5.0/cosmos_sdk_proto/index.html) for Rust. This is not generic Rust but code generation specific to [prost](https://crates.io/crates/prost). I like the use of git submodules as a .proto source in https://github.com/cosmos/cosmos-rust.

willclarktech:
Closes #799

This PR demonstrates the compatibility of existing @cosmjs packages with the new cosmjs-types
package.


webmaster128 (approved):
Very nice 🔥

PR #821, opened 2021-06-08
Fix company name and update NOTICE

webmaster128:


ethanfrey (approved):
Looks good.
Do you not want to add Confio GmbH to the package.json contributors as well?

PR #824, opened 2021-06-09
Add Yarn PnP eslint SDK

willclarktech:
Enables eg VSCode extension

### DIFF
--- a/.yarn/sdks/typescript/lib/tsserver.js
+++ b/.yarn/sdks/typescript/lib/tsserver.js
@@ -13,6 +13,9 @@ const moduleWrapper = tsserver => {
   const {isAbsolute} = require(`path`);
   const pnpApi = require(`pnpapi`);
 
+  const isVirtual = str => str.match(/\/(\$\$virtual|__virtual__)\//);
+  const normalize = str => str.replace(/\\/g, `/`).replace(/^\/?/, `/`);
+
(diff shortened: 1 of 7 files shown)


.yarn/sdks/typescript/lib/tsserver.js
```diff
@@ -96,9 +114,9 @@ const moduleWrapper = tsserver => {
         parsedMessage != null &&
         typeof parsedMessage === `object` &&
         parsedMessage.arguments &&
-        parsedMessage.arguments.hostInfo === `vscode`
+        typeof parsedMessage.arguments.hostInfo === `string`
       ) {
-        isVSCode = true;
+        hostInfo = parsedMessage.arguments.hostInfo;
```
webmaster128, line 119: Why is this file changing whout a change in the lockfile?

webmaster128 (approved):

.yarn/sdks/typescript/lib/tsserver.js
willclarktech: I believe because I re-ran `yarn dlx @yarnpkg/pnpify --sdk base`, which will have downloaded the latest version of that package. See eg https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-pnpify/sources/sdks/base.ts#L37

webmaster128: Thanks

PR #833, opened 2021-06-15
Remove default fees

### LINKED ISSUE #827: Restructure fees
The current setup with default fee tables doesn't make much sense. It's too simplistic a model of the complexity involved in choosing an appropriate fee, the defaults only work with a `ucosm` denomination, and customisation requires a whole new client, which is not convenient. The following applies to all signing clients:

- [ ] Remove default fee tables
- [ ] Remove the gas price setting
- [ ] Make fee required everywhere where a transaction is created
- [ ] Provide a function that calculates a `StdFee` given a `GasPrice` and a gas limit

willclarktech:
Closes #827

### DIFF
--- a/scripts/wasmd/send_first.js
+++ b/scripts/wasmd/send_first.js
@@ -25,7 +25,7 @@ async function main() {
   const client = await SigningStargateClient.connectWithSigner(rpcUrl, wallet, { prefix: prefix });
   const recipient = Bech32.encode(prefix, Random.getBytes(20));
   const amount = coins(226644, "ucosm");
-  const fee = calculateFee(80_000, GasPrice.fromString("0.025ucosm"));
+  const fee = calculateFee(80_000, "0.025ucosm");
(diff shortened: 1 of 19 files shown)


scripts/wasmd/send_first.js
```diff
@@ -20,8 +25,9 @@ async function main() {
   const client = await SigningStargateClient.connectWithSigner(rpcUrl, wallet, { prefix: prefix });
   const recipient = Bech32.encode(prefix, Random.getBytes(20));
   const amount = coins(226644, "ucosm");
+  const fee = calculateFee(80_000, GasPrice.fromString("0.025ucosm"));
```
webmaster128, line 28:
```suggestion
  const fee = calculateFee(80_000, "0.025ucosm");
```

What about accepting `GasPrice | string` to make this simple case more compact?

webmaster128 (approved):
🎖

webmaster128 (approved):
Thanks!

PR #834, opened 2021-06-15
Add CLI example for generating Cosmos HD path test vectors

webmaster128:


webmaster128:
This was needed for the proposed spec: https://github.com/cosmos/cips/pull/11. Unfortunately it never got any traction 😢. RIP

PR #840, opened 2021-07-09
fix: Add ibc, liquidity pool (and more) denom support to parseCoins()

clockworkgr:
As per: https://github.com/cosmos/cosmos-sdk/blob/master/types/coin.go#L637

### DIFF
--- a/packages/amino/src/coins.ts
+++ b/packages/amino/src/coins.ts
@@ -28,7 +28,7 @@ export function parseCoins(input: string): Coin[] {
     .split(",")
     .filter(Boolean)
     .map((part) => {
-      const match = part.match(/^([0-9]+)([a-zA-Z]+)/);
+      const match = part.match(/^([0-9]+)([a-zA-Z][a-zA-Z0-9/-]{2,127})$/);
       if (!match) throw new Error("Got an invalid coin string");
       return {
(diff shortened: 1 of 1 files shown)


webmaster128:
Thanks for bringing this up. Where/how do you use `parseCoins`? It seems to be unused internally.

We duplicate that logic in `GasPrice.fromString`, which is implemented twice with different patterns in launchpad and stargate.

clockworkgr:
I use it in quite a few places since our app uses the `1234uatom` format a fair bit.

Tbh I can't remember where/when I first came across the existence of parseCoins but I've been using it since for that purpose

webmaster128:
If you only need one, could you use `coin(value, "denom")` or `coins(value, "denom")` instead? The first one returns a `Coin` the second one a `Coin[]`. This contains much fewer logic and checks.

In order to make `parseCoins` Stargate ready, we should have a Stargate specific version:
- Remove the re-export `parseCoins` from `packages/proto-signing/src/index.ts`
- Re-implement `parseCoins` in `packages/proto-signing/src/coins.ts` and export in `packages/proto-signing/src/index.ts`

webmaster128:
@clockworkgr could you have a look at https://github.com/cosmos/cosmjs/pull/845? I think this is what you are looking for.

PR #847, opened 2021-07-21
Implement ARD-038 data signing

webmaster128:


mikedotexe:
Thank you for doing this. I'm going to try it out and see if I can borrow logic. Whether or not it works for me, thanks :)

webmaster128:
@mikedotexe please note that ADR-036 is not finalized. The current state seems to be completely discareded. Nobody is owning the spec process right now. This is one of a million ways to implement the functionality, but we're missing a spec.

mikedotexe:
> @mikedotexe please note that ADR-036 is not finalized. The current state seems to be completely discareded. Nobody is owning the spec process right now. This is one of a million ways to implement the functionality, but we're missing a spec.

Thank you, I hadn't noticed that the status was "draft" here:
https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-036-arbitrary-signature.md

I wasn't able to use the code from here, but did find that I could use `signAmino` code I found in some unit tests elsewhere.

Basically…

```js
    const signDoc = {
      msgs: [{
        type: 'starry-login',
        value: this.state.messageToSign
      }],
      fee: {
        amount: [],
        // Note: this needs to be 0 gas to comply with ADR36, but Keplr current throws an error. See: https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-036-arbitrary-signature.md#decision
        gas: "1" },
      chain_id: this.chainId,
      memo: "You are powerful and capable, friend.",
      account_number: "0",
      sequence: "0",
    };

    try {
      const { signed, signature } = await this.signer.signAmino(this.accounts[0].address, signDoc);
      // TODO: here's where do we a post request to our backend, localhost:5000 with info
      //   we'll copy the code stuffs below
      const valid = await Secp256k1.verifySignature(
        Secp256k1Signature.fromFixedLength(fromBase64(signature.signature)),
        sha256(serializeSignDoc(signed)),
        this.accounts[0].pubkey,
      );
      console.log('valid', valid);
      // Clear the messages area
      this.setState({message: ''});
    } catch (e) {
    …
```

I also noticed that Keplr doesn't allow 0 gas to be attached yet. I'll be making an issue and PR for that as soon as I can get the local version to build so I can test it.

I want to thank you for your response and concern. Your code here and elsewhere has been invaluable and thanks for following up. We're building a fun Discord bot for our project, wish us luck!

rabm2k7:
Is this ever going to be merged?

webmaster128:
> Is this ever going to be merged?

Unfortunately my last info is that there is not agreement on a spec and I don't see a point in merging some draft state.

What could work is find a few parties to develop the current draft into a spec that is independent of Cosmos SDK and has a different name.

PR #851, opened 2021-07-22
Add field `ibcPortId` to `Contract`

### LINKED ISSUE #836: Expose ibc port in cosmwasm queries
We don't expose the full contract info in the queries. [`getContractInfo`](https://github.com/cosmos/cosmjs/blob/main/packages/cosmwasm-stargate/src/queries/wasm.ts#L87-L90) returns [`ContractInfo`](https://github.com/cosmos/cosmjs/blob/06fbc34f72f12c30a396c3ca296f80eca9fa60b0/packages/cosmwasm-launchpad/src/lcdapi/wasm.ts#L45-L53) which is missing a number of fields.

This is what is [available in wasm `v0.16.0`](https://github.com/CosmWasm/wasmd/blob/v0.16.0/proto/cosmwasm/wasm/v1beta1/types.proto#L71-L93).

`ibc_port` is especially useful for me.

webmaster128:
Closes #836


ethanfrey (approved):
Looks good.
I'd love to see a positive test case, but I guess there are no ibc contracts in cosmjs test cases now and that would be a lot to add one for a minor test

webmaster128:
> I'd love to see a positive test case

Me too. At which point is a port ID assigned to a contract by wasmd? Would it be enough to instantiate the ibs reflect contract?

packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
```diff
@@ -291,20 +292,33 @@ describe("CosmWasmClient", () => {
...
+      expect(contract).toEqual(
         jasmine.objectContaining({
-          address: deployedHackatom.instances[2].address,
-          codeId: deployedHackatom.codeId,
-          creator: alice.address0,
-          label: deployedHackatom.instances[2].label,
-          admin: alice.address1,
+          address: deployedIbcReflect.instances[0],
+          codeId: deployedIbcReflect.codeId,
+          ibcPortId: "wasm.wasm1vjecguu37pmd577339wrdp208ddzymku8yy0te",
```
webmaster128, line 321: Now we get a port ID!

ethanfrey:
> 

> Me too. At which point is a port ID assigned to a contract by wasmd? Would it be enough to instantiate the ibs reflect contract?

Yes. But looks like you found that out already.

ethanfrey (approved):
Good stuff. Thanks.

Happy to see this merged

PR #853, opened 2021-07-26
Fix AminoMsgTransfer/AminoHeight encoding

webmaster128:


packages/stargate/src/aminotypes.spec.ts
```diff
@@ -353,6 +353,70 @@ describe("AminoTypes", () => {
...
+      });
+      const expected: AminoMsgTransfer = {
+        type: "cosmos-sdk/MsgTransfer",
+        value: {
+          source_port: "testport",
+          source_channel: "testchannel",
+          token: coin(1234, "utest"),
+          sender: "cosmos1pkptre7fdkl6gfrzlesjjvhxhlc3r4gmmk8rs6",
+          receiver: "cosmos10dyr9899g6t0pelew4nvf4j5c3jcgv0r73qga5",
+          timeout_height: {}, // 🤷‍♂️
```
ethanfrey, line 413: 😄

packages/stargate/src/aminotypes.spec.ts
```diff
@@ -353,6 +353,70 @@ describe("AminoTypes", () => {
...
+      });
+      const expected: AminoMsgTransfer = {
+        type: "cosmos-sdk/MsgTransfer",
+        value: {
+          source_port: "testport",
+          source_channel: "testchannel",
+          token: coin(1234, "utest"),
+          sender: "cosmos1pkptre7fdkl6gfrzlesjjvhxhlc3r4gmmk8rs6",
+          receiver: "cosmos10dyr9899g6t0pelew4nvf4j5c3jcgv0r73qga5",
+          timeout_height: {
```
ethanfrey, line 381: Is there a difference between this and `timeout_height: {}`?

I guess JSON (at least) would encode them the same. Is there some other subtle difference that might appear in JS land?

packages/stargate/src/signingstargateclient.spec.ts
```diff
@@ -321,6 +322,86 @@ describe("SigningStargateClient", () => {
...
+          faucet.address0,
+          faucet.address1,
+          coin(1234, "ucosm"),
+          "fooPort",
+          "fooChannel",
+          { revisionHeight: Long.fromNumber(123), revisionNumber: Long.fromNumber(456) },
+          Math.floor(Date.now() / 1000) + 60,
+          memo,
+        );
+        // CheckTx must pass but the execution must fail in DeliverTx due to invalid channel/port
```
ethanfrey, line 344: Would you get another response if CheckTx fails? Eg. would `sendIbcTokens` throw on CheckTx error?

Is there any way you can ensure this is due to DeliverTx not CheckTx?

ethanfrey (approved):
Looks good. Two minor questions

packages/stargate/src/aminotypes.spec.ts
webmaster128: In JS and Jasmine testing there is a difference between `{ foo: undefined }` and `{}`. However, when serializing them to JSON they become the same.

It is written that way because the implementation always sets the `revision_height`/`revision_number` properties. We could change that but this would create a lot of hard to read code.

packages/stargate/src/signingstargateclient.spec.ts
webmaster128: Yes, `await client.sendIbcTokens` throws and error on submission fails (CheckTx) but not on execution fails (DeliverTx).

PR #857, opened 2021-07-28
Add ADR-028 implementation and test CosmWasm contract addresses

webmaster128:


packages/proto-signing/src/adr28.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { Bech32 } from "@cosmjs/encoding";
+import { Uint64 } from "@cosmjs/math";
+
+import { moduleAddressRaw } from "./adr28";
+
+describe("adr28", () => {
+  describe("moduleAddressRaw", () => {
+    it("can be used for wasm contract address derivation", () => {
+      function contractAddressRaw(codeId: number, contractId: number): Uint8Array {
```
ethanfrey, line 9: very cool. glad to see this cross-test

ethanfrey: did you want to expose this function not just in tests? (but i guess in cosmwasm package?)

The `slice(0, 20)` is unique to our 0.42-compatible code. The 0.43 sdk has 32 byte addresses :scream: (actually a mix between 20 and 32 I think) hope that doesn't break other tests

ethanfrey (approved):
Nice implementation

packages/proto-signing/src/adr28.spec.ts
webmaster128: This is just to test the generic ADR-28 implementation. Turns out no CosmWasm client needts to know how contract addresses are generated. They just use them. So I don't see a point in exposing that.

webmaster128:
Closing as not needed. See `function instantiate2Address` where some of that functionality is implemented as well.

PR #876, opened 2021-09-04
Update SigningCosmWasmClient.execute to accept multiple messages

ben2x4:
Proposing a change to the inputs of `SigningCosmWasmClient.execute` to accept multiple transactions. I think this functionality is important to expose and altering the input to an array is the cleanest implementation.

If we want to avoid introducing breaking change this could be implemented with execute msg having the type of `Record<string, unknown> |   Array<Record<string, unknown>>` or as a separate function, ie `SigningCosmWasmClient.executeMultiple`. 

Would love to discuss! I have been signing multi-msg transactions with `SigningCosmWasmClient.signAndBroadcast` directly and it is quite unergonomic.

### DIFF
--- a/packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
+++ b/packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
@@ -339,21 +339,21 @@ export class SigningCosmWasmClient extends CosmWasmClient {
   public async execute(
     senderAddress: string,
     contractAddress: string,
-    msg: Record<string, unknown>,
+    msgs: Array<Record<string, unknown>>,
     fee: StdFee,
     memo = "",
     funds?: readonly Coin[],
   ): Promise<ExecuteResult> {
-    const executeContractMsg: MsgExecuteContractEncodeObject = {
+    const executeContractMsgs: Array<MsgExecuteContractEncodeObject> = msgs.map((m) => {return {
       typeUrl: "/cosmwasm.wasm.v1.MsgExecuteContract",
       value: MsgExecuteContract.fromPartial({
         sender: senderAddress,
         contract: contractAddress,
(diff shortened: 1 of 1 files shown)


webmaster128:
Thank you for your contribution. The diff makes sense to me.

However, I wonder about the use case. Why do you need to have multiple smart contract interactions from the same sender to the same contract in a single transaction? Seems like a pretty advanced nieche case to me. But happy to learn more.

packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
```diff
@@ -339,21 +339,21 @@ export class SigningCosmWasmClient extends CosmWasmClient {
...
   ): Promise<ExecuteResult> {
-    const executeContractMsg: MsgExecuteContractEncodeObject = {
+    const executeContractMsgs: Array<MsgExecuteContractEncodeObject> = msgs.map((m) => {return {
       typeUrl: "/cosmwasm.wasm.v1.MsgExecuteContract",
       value: MsgExecuteContract.fromPartial({
         sender: senderAddress,
         contract: contractAddress,
-        msg: toUtf8(JSON.stringify(msg)),
+        msg: toUtf8(JSON.stringify(m)),
         funds: [...(funds || [])],
```
webmaster128, line 353: The funds line here is a problem now. The user would send the funds with every message, so an input of 5 messages and 20 utkn would cost 100 utkn. This is very unintuative.

ben2x4: ah got it, will look at this

webmaster128:
This will be implemented and shipped soon as part of 0.29.0. See https://github.com/cosmos/cosmjs/issues/1072

webmaster128:

PR #884, opened 2021-09-29
#882 Decode multisig pubkey

### LINKED ISSUE #882: decode multisig acc
function `decodeBech32Pubkey` can't decode multisig keys of next format `bostrompub1ytql0csgqgfzd666axrjzqm4p586gn8vrm6958zd4c9lje8xeahzcf35mwsjw3gaxn7xnhe9eufzd666axrjzqcc9h5ms6x96eg23e4hxghnvwn7xwtuhgeupvg0lgppn7d6dkyaxc66dhat`

seems multisig branch is missed https://github.com/cosmos/cosmjs/blob/b2279928c6d46a568e619cafddc6c1bbd17ec1b0/packages/amino/src/encoding.ts#L37

hleb-albau:
resolves #882

### DIFF
--- a/packages/amino/src/encoding.ts
+++ b/packages/amino/src/encoding.ts
@@ -6,6 +6,8 @@ import {
   isEd25519Pubkey,
   isMultisigThresholdPubkey,
   isSecp256k1Pubkey,
+  isSinglePubkey,
+  MultisigThresholdPubkey,
   Pubkey,
   pubkeyType,
   Secp256k1Pubkey,
@@ -61,6 +63,9 @@ export function decodeAminoPubkey(data: Uint8Array): Pubkey {
       type: pubkeyType.sr25519,
       value: toBase64(rest),
     };
+  } else if (arrayContentStartsWith(data, pubkeyAminoPrefixMultisigThreshold)) {
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define
+    return decodeMultisigPubkey(data);
   } else {
     throw new Error("Unsupported public key type. Amino data starts with: " + toHex(data.slice(0, 5)));
   }
@@ -75,6 +80,53 @@ export function decodeAminoPubkey(data: Uint8Array): Pubkey {
 export function decodeBech32Pubkey(bechEncoded: string): Pubkey {
   const { data } = Bech32.decode(bechEncoded);
   return decodeAminoPubkey(data);
+}
+
+/**
+ * Decodes a multisig pubkey to type object.
+ * Pubkey structure [ prefix + const + threshold + loop:(const + pubkeyLength + pubkey            ) ]
+ *                  [   4b   + 1b    +   1b      + loop:(1b    +    1b        + pubkeyLength bytes) ]
+ * @param data encoded pubkey
+ */
+function decodeMultisigPubkey(data: Uint8Array): MultisigThresholdPubkey {
+  if (data.length < 4 + 1 + 1) {
+    // verify that we can read at least threshold
+    throw new Error("Invalid multisig data length.");
+  }
+  let rest = data.slice(pubkeyAminoPrefixSecp256k1.length);
+  rest = rest.slice(1); // removing 0x08;
+  const threshold = rest[0];
+  rest = rest.slice(1); // removing threshold
+
+  const pubkeys = [];
+  while (rest.length > 0) {
+    if (rest.length < 1 + 1) {
+      // verify that we can read at least pubkeyLength
+      throw new Error("Invalid multisig data length.");
+    }
+    rest = rest.slice(1); // removing 0x12
+    const pubkeyLength = rest[0];
+    rest = rest.slice(1); // removing pubkey length
+    if (rest.length < pubkeyLength) {
+      // verify that we can read pubkey
+      throw new Error("Invalid multisig data length.");
+    }
+    const encodedPubkey = rest.slice(0, pubkeyLength);
+    rest = rest.slice(pubkeyLength); // removing pubkey
+    const pubkey = decodeAminoPubkey(encodedPubkey);
+    if (!isSinglePubkey(pubkey)) {
+      throw new Error("Unsupported multisig participant pubkey type");
+    }
+    pubkeys.push(pubkey);
+  }
+
+  return {
+    type: pubkeyType.multisigThreshold,
+    value: {
+      threshold: threshold.toString(),
+      pubkeys: pubkeys,
+    },
+  };
 }
 
 /**
--- a/packages/amino/src/encoding.ts
+++ b/packages/amino/src/encoding.ts
@@ -64,7 +64,7 @@ export function decodeAminoPubkey(data: Uint8Array): Pubkey {
     };
   } else if (arrayContentStartsWith(data, pubkeyAminoPrefixMultisigThreshold)) {
     // eslint-disable-next-line @typescript-eslint/no-use-before-define
-    return decodeMultisigPubkey(Array.from(data));
+    return decodeMultisigPubkey(data);
   } else {
     throw new Error("Unsupported public key type. Amino data starts with: " + toHex(data.slice(0, 5)));
   }
@@ -83,14 +83,17 @@ export function decodeBech32Pubkey(bechEncoded: string): Pubkey {
 
 /**
  * Uvarint decoder for Amino.
- * Decoding numbers > 127 is not supported here. Please tell those lazy CosmJS maintainers to port the binary.Varint implementation from the Go standard library and write some tests.
- *
  * @see https://github.com/tendermint/go-amino/blob/8e779b71f40d175/decoder.go#L64-76
  * @returns varint as number, and bytes count occupied by varaint
  */
 function decodeUvarint(reader: number[]): [number, number] {
   if (reader.length < 1) {
     throw new Error("Can't decode varint. EOF");
+  }
+  if (reader[0] > 127) {
+    throw new Error(
+      "Decoding numbers > 127 is not supported here. Please tell those lazy CosmJS maintainers to port the binary.Varint implementation from the Go standard library and write some tests.",
+    );
   }
   return [reader[0], 1];
 }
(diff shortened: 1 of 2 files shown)


packages/amino/src/encoding.ts
```diff
@@ -30,6 +32,49 @@ const pubkeyAminoPrefixSr25519 = fromHex("0dfb1005" + "20" /* fixed length */);
...
+ * @param data encoded pubkey
+ */
+function decodeMultisigPubkey(data: Uint8Array): MultisigThresholdPubkey {
+  if (data.length < 4 + 1 + 1) { // verify that we can read at least threshold
+    throw new Error("Invalid multisig data length.");
+  }
+  let rest = data.slice(pubkeyAminoPrefixSecp256k1.length);
+  rest = rest.slice(1); // removing 0x08;
+  const threshold = rest[0];
+  rest = rest.slice(1); // removing threshold
```
webmaster128: This is not constant 1 byte but a variable length int. Could you create a decoder similar to `encodeUvarint` for clarity? It is okay if ot only supports values up to 127

packages/amino/src/encoding.ts
```diff
@@ -30,6 +32,49 @@ const pubkeyAminoPrefixSr25519 = fromHex("0dfb1005" + "20" /* fixed length */);
...
+
+  const pubkeys = [];
+  while (rest.length > 0) {
+    if (rest.length < 1 + 1) { // verify that we can read at least pubkeyLength
+      throw new Error("Invalid multisig data length.");
+    }
+    rest = rest.slice(1); // removing 0x12
+    const pubkeyLength = rest[0];
+    rest = rest.slice(1) // removing pubkey length
```
webmaster128: Also varint, as above

packages/amino/src/encoding.ts
```diff
@@ -30,6 +32,49 @@ const pubkeyAminoPrefixSr25519 = fromHex("0dfb1005" + "20" /* fixed length */);
...
+  rest = rest.slice(1); // removing 0x08;
+  const threshold = rest[0];
+  rest = rest.slice(1); // removing threshold
+
+  const pubkeys = [];
+  while (rest.length > 0) {
+    if (rest.length < 1 + 1) { // verify that we can read at least pubkeyLength
+      throw new Error("Invalid multisig data length.");
+    }
+    rest = rest.slice(1); // removing 0x12
```
webmaster128: Can we check this is 0x12? This is some sort of type identifier and we should assert it is what we expect.

packages/amino/src/encoding.ts
```diff
@@ -30,6 +32,49 @@ const pubkeyAminoPrefixSr25519 = fromHex("0dfb1005" + "20" /* fixed length */);
...
+ * Pubkey structure [ prefix + const + threshold + loop:(const + pubkeyLength + pubkey            ) ]
+ *                  [   4b   + 1b    +   1b      + loop:(1b    +    1b        + pubkeyLength bytes) ]
+ * @param data encoded pubkey
+ */
+function decodeMultisigPubkey(data: Uint8Array): MultisigThresholdPubkey {
+  if (data.length < 4 + 1 + 1) { // verify that we can read at least threshold
+    throw new Error("Invalid multisig data length.");
+  }
+  let rest = data.slice(pubkeyAminoPrefixSecp256k1.length);
+  rest = rest.slice(1); // removing 0x08;
```
webmaster128: Can we check this is 0x08? This is some sort of type identifier and we should assert it is what we expect.

packages/amino/src/encoding.ts
```diff
@@ -77,6 +82,53 @@ export function decodeBech32Pubkey(bechEncoded: string): Pubkey {
...
+ * Pubkey structure [ prefix + const + threshold + loop:(const + pubkeyLength + pubkey            ) ]
+ *                  [   4b   + 1b    +   1b      + loop:(1b    +    1b        + pubkeyLength bytes) ]
+ * @param data encoded pubkey
+ */
+function decodeMultisigPubkey(data: Uint8Array): MultisigThresholdPubkey {
+  if (data.length < 4 + 1 + 1) {
+    // verify that we can read at least threshold
+    throw new Error("Invalid multisig data length.");
+  }
+  let rest = data.slice(pubkeyAminoPrefixSecp256k1.length);
```
webmaster128, line 96: The way you use `rest` here feels like it would be better to convert the `Uint8Array` to a mutable array: `Array.from(data)`. Then you can use `shift` and `splice` to get and delete in one call.

packages/amino/src/encoding.ts
hleb-albau: ups, I expected  uint8 here, but you are right - it is varint!

packages/amino/src/encoding.ts
hleb-albau: Golang developer detected :)

hleb-albau:
@webmaster128 ready for reaview

packages/amino/src/encoding.ts
```diff
@@ -61,6 +62,9 @@ export function decodeAminoPubkey(data: Uint8Array): Pubkey {
       type: pubkeyType.sr25519,
       value: toBase64(rest),
     };
+  } else if (arrayContentStartsWith(data, pubkeyAminoPrefixMultisigThreshold)) {
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define
+    return decodeMultisigPubkey(Array.from(data));
```
webmaster128, line 67: Please move implementation up such that we don't need to disable this lint. I know this is typically done different in many other code bases but here we decided to keep the link and have the dependencies above.

packages/amino/src/encoding.ts
```diff
@@ -77,6 +81,84 @@ export function decodeBech32Pubkey(bechEncoded: string): Pubkey {
...
+  return [reader[0], 1];
+}
+
+/**
+ * Decodes a multisig pubkey to type object.
+ * Pubkey structure [ prefix + const + threshold + loop:(const + pubkeyLength + pubkey            ) ]
+ *                  [   4b   + 1b    +  varint   + loop:(1b    +    varint    + pubkeyLength bytes) ]
+ * @param reader encoded pubkey
+ */
+function decodeMultisigPubkey(reader: number[]): MultisigThresholdPubkey {
```
webmaster128, line 104: Let's keep the input a `data: Uint8Array` here in the function signature. Then do `const reader = Array.from(data)` internally. Otherwiese we would have to check that every `number` is an int between 0 and 255 and nothing crazy like negative, float or NaN.

packages/amino/src/encoding.ts
```diff
@@ -77,6 +81,84 @@ export function decodeBech32Pubkey(bechEncoded: string): Pubkey {
...
+  // verify that we can at least start reading threshold
+  if (reader.length < pubkeyAminoPrefixMultisigThreshold.length + 1 + 1) {
+    throw new Error("Invalid multisig data length.");
+  }
```
webmaster128, line 108: This check is now redundant with the reading below and can be removed.

packages/amino/src/encoding.ts
```diff
@@ -77,6 +81,84 @@ export function decodeBech32Pubkey(bechEncoded: string): Pubkey {
...
+ * Decoding numbers > 127 is not supported here. Please tell those lazy CosmJS maintainers to port the binary.Varint implementation from the Go standard library and write some tests.
+ *
+ * @see https://github.com/tendermint/go-amino/blob/8e779b71f40d175/decoder.go#L64-76
+ * @returns varint as number, and bytes count occupied by varaint
+ */
+function decodeUvarint(reader: number[]): [number, number] {
+  if (reader.length < 1) {
+    throw new Error("Can't decode varint. EOF");
+  }
+  return [reader[0], 1];
```
webmaster128, line 95: We need to check that `reader[0]` is <= 127. Higher values indicate that more bytes are following.

packages/amino/src/encoding.ts
```diff
@@ -77,6 +81,84 @@ export function decodeBech32Pubkey(bechEncoded: string): Pubkey {
...
+  }
+
+  // remove multisig amino prefix;
+  const prefixFromReader = reader.splice(0, pubkeyAminoPrefixMultisigThreshold.length);
+  if (!arrayContentStartsWith(prefixFromReader, pubkeyAminoPrefixMultisigThreshold)) {
+    throw new Error("Invalid multisig prefix.");
+  }
+
+  // remove 0x08 threshold prefix;
+  if (reader.shift() != 0x08) {
```
webmaster128, line 117: Much nicer than before!

packages/amino/src/encoding.ts
```diff
@@ -77,6 +81,84 @@ export function decodeBech32Pubkey(bechEncoded: string): Pubkey {
...
+  const [threshold, thresholdBytesLength] = decodeUvarint(reader);
+  reader.splice(0, thresholdBytesLength);
+
+  // read participants pubkeys
+  const pubkeys = [];
+  while (reader.length > 0) {
+    // verify that we can at least start reading pubkeyLength
+    if (reader.length < 1 + 1) {
+      throw new Error("Invalid multisig data length.");
+    }
```
webmaster128, line 131: Redundant with the next two checked reads. This can be removed.

webmaster128 (commented):
🐎

packages/amino/src/encoding.ts
hleb-albau: Here is a problem. These two functions invoke each other. Thus we if a place one on top, another will produce a warning. That's why I suppressed the warning.

hleb-albau: Also, I read SDK code, and participants of multisig can be another multisig. So we have some sort of recursion here.

packages/amino/src/encoding.ts
hleb-albau: great note

hleb-albau:
@webmaster128 next round of fixes done

packages/amino/src/encoding.ts
webmaster128: Ah, I see, thanks for clarification!

webmaster128 (approved):

PR #892, opened 2021-10-04
Update signingstargateclient.ts

blackpanther1881:
removing extra space in messageTimeout

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- @cosmjs/stargate: remove extra space in messageTimeout registry.
+
 ## [0.26.1] - 2021-09-30
 
 ### Added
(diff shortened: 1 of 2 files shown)


puneet2019:
@webmaster128 can we also add `MsgCreateVestingAccount` to the default registry? 
`["/cosmos.vesting.v1beta1.MsgCreateVestingAccount", vestingTx.MsgCreateVestingAccount]`

webmaster128:
@puneet2019 I think so, yes. But let's track this independently. Ideally we write a test for it first to avoid bugs like this one.

webmaster128 (commented):
Thanks. Could you add a CHNAGLOG entry for this in the `### Fixed` section of `## [Unreleased]`?

blackpanther1881:
@webmaster128 added CHNAGLOG

webmaster128:
Could you do that in the same PR to have everything together? Also please note you put it in the 0.26.1 section that is already released.

webmaster128 (approved):

PR #893, opened 2021-10-04
Add custom ledger app name support

calvinkei:
Currently LaunchpadLedger forces users to use Cosmos ledger app. However there are cosmos-sdk based chains uses their own ledger app as they have different coin type (eg, Band, Terra, Kava, Desmos)

This PR solves the problem by allowing users to input an option `ledgerAppName` to pass the ledger app name check

### DIFF
--- a/packages/ledger-amino/src/launchpadledger.ts
+++ b/packages/ledger-amino/src/launchpadledger.ts
@@ -24,7 +24,7 @@ function unharden(hdPath: HdPath): number[] {
 
 const cosmosHdPath = makeCosmoshubPath(0);
 const cosmosBech32Prefix = "cosmos";
-const cosmosLedgerAppName = "cosmos";
+const cosmosLedgerAppName = "Cosmos";
 const requiredCosmosAppVersion = "1.5.3";
 
 export interface LaunchpadLedgerOptions {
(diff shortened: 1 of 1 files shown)


packages/ledger-amino/src/launchpadledger.ts
```diff
@@ -24,30 +24,35 @@ function unharden(hdPath: HdPath): number[] {
...
 const cosmosHdPath = makeCosmoshubPath(0);
 const cosmosBech32Prefix = "cosmos";
+const cosmosLedgerAppName = "cosmos";
 const requiredCosmosAppVersion = "1.5.3";
 
 export interface LaunchpadLedgerOptions {
   readonly hdPaths?: readonly HdPath[];
   readonly prefix?: string;
   readonly testModeAllowed?: boolean;
+  readonly ledgerAppName?: string
```
webmaster128, line 34:
```suggestion
  /**
   * The name of the app the user must have opened on the Ledger.
   * This allows you to use this connector with forks of the Cosmos Hub Ledger app.
   * Support is provided on an best effort basis and only as long as those forks do not
   * significantly differ from the original app.
   *
   * Defaults to "cosmos".
   */
  readonly ledgerAppName?: string
```

webmaster128 (commented):
Interesting. Thanks for this PR.

Could you go through the whole file and check other relevant places? E.g. `assert(this.app, "Cosmos Ledger App is not connected");` could be generalized. The required app version should also become configurable as I would not expect forks to follow the same versioning.

webmaster128:
Thanks a lot for your contribution and upstreaming the changes you need instead of forking the codebase.

This work was merged as part of #908.

Issue #903, opened 2021-10-17
Generate and store salt


shapkarin:
Hello. It there any way to make salt configurable? 

https://github.com/cosmos/cosmjs/blob/f198f0acc425051f0ebd71489d993d7851b48475/packages/proto-signing/src/wallet.ts#L15

I found a way to replace that file. And put salt to the env variable:
```js
new webpack.NormalModuleReplacementPlugin(
  new RegExp('@cosmjs/proto-signing/build/wallet.js'),
  path.resolve(
    __dirname,
    'replaced/wallet.js'
  )
)
```
Of course someone can find the value in a builded js, but it looks harder.

webmaster128:
> Hello. It there any way to make salt configurable?

Heyho! Why do you want to do that? I'd be curious about a detailed motivation and usecase description. If it makes sense to do, it should be trivial to make it configurable through proper APIs.

> Of course someone can find the value in a builded js, but it looks harder.

Are you sure you're on the right path here? The salt is not a secret value. It's public and that's perfectly fine. It just makes rainbow attacks harder.

shapkarin:
While wallet development, I decided that If `It just makes rainbow attacks harder`, why not make it a bit more harder with "hidden" salt from env variable.

webmaster128:
Okay, so one scenario could be this: All CosmJS users use the same salt. One can come up with a rainbow table of passwords that are all hashed with one salt in order to attack CosmJS users. Now what you can do is override the salt in each application. Still no need for env variables and hiding. Then you need application specific tables instead of lib specific tables, making them much harder to generate or less attractive.

I'm unsure if this is worth it. Changing the argon 2 parameters should be enough to ger unique hashes and at the same time tune the difficulty of the KDF. Then if you push this computation to a WebWorker to not freeze the UI you can go really high with your params. Development effort is much more worth there. Also this heavily reduces the risk that an encrypted wallet cannot be decrypted anymore because you remain compatible with standard CosmJS tooling.

shapkarin:
> Then you need application specific tables instead of lib specific tables, making them much harder to generate or less attractive.

That was the goal. Instead of `lib specific tables` has different `application specific tables` so one table can't be used to any app.

webmaster128:
Yeah. This can be done and there is nothing wrong with it. But I highly recommend not making this a priority and don't implement costy and risky hacks for it.

I leave this ticket open to make the salt configurable. This is easy to implement.

webmaster128:
[...]

PR #906, opened 2021-10-19
Test wallet encryption improvements

shapkarin:
related issue #903


packages/crypto/src/libsodium.ts
```diff
@@ -36,10 +38,11 @@ export function isArgon2idOptions(thing: unknown): thing is Argon2idOptions {
 export class Argon2id {
   public static async execute(
     password: string,
-    salt: Uint8Array,
+    saltInput: Uint8Array,
     options: Argon2idOptions,
   ): Promise<Uint8Array> {
     await sodium.ready;
+    const salt: Uint8Array = saltInput || Random.getBytes(sodium.crypto_pwhash_SALTBYTES);
```
shapkarin, line 45:
```suggestion
const salt: Uint8Array = saltInput || sodium.randombytes_buf(sodium.crypto_pwhash_SALTBYTES)
```

shapkarin: that may be placed to the `wallet.ts executeKdf`

shapkarin: Will implement it as it was implemented for the `wallet.encrypt`

PR #915, opened 2021-10-29
Use Buffer from module

sunnya97:
When I try to use some of the cosmjs crypto libraries in the browser, not in node, the RIPEMD hash function doesn't work.

```
(node:54788) [MODULE_NOT_FOUND] Error: ripemd160 tried to access buffer. While this module is usually interpreted as a Node builtin, your resolver is running inside a non-Node resolution context where such builtins are ignored. Since buffer isn't otherwise declared in ripemd160's dependencies, this makes the require call ambiguous and unsound.
```

Using this solution here fixed it:  https://stackoverflow.com/a/55226877/3299736

### DIFF
--- a/packages/crypto/src/ripemd.ts
+++ b/packages/crypto/src/ripemd.ts
@@ -1,3 +1,4 @@
+import { Buffer } from "buffer";
 import RIPEMD160 from "ripemd160";
 
 import { HashFunction } from "./hash";


webmaster128:
Thanks for bringing this up. The change makese sense, but even better to not rely on the non-JS type `Buffer` at all. See https://github.com/cosmos/cosmjs/pull/916.

I wonder how many of our dependencies simply assume there is a global Buffer available, such that we have the same problem in other places. Which bundeler are you using to bundle commonjs modules like CosmJS for the browser? With Webpack 4 you got a Buffer injected by default. In Webpack 5 you need to do that explicitly via:

```
    plugins: [
      new webpack.ProvidePlugin({
        Buffer: ["buffer", "Buffer"],
      }),
    ],
```

webmaster128:
The error might even point to a problem in ripemd160: In https://github.com/crypto-browserify/ripemd160/blob/v2.0.2/index.js#L2 `buffer` is required but it is [not declared as a dependency](https://github.com/crypto-browserify/ripemd160/blob/v2.0.2/package.json#L28-L31).

sunnya97:
I am using webpack 5

I actually just copied the webpack config from the ledger-amino package, which seems to have disabled buffer fallback

https://github.com/osmosis-labs/cosmjs/blob/main/packages/ledger-amino/webpack.demo.config.js#L19

webmaster128:
Okay, so we agree the root of this issue us a missing dependency in ripemd160 and you can work around it but explicitly enabling a Buffer implementation in the webpack 5 config (that we got by default in webpack 4 by the way), right?

sunnya97:
I suppose.  Sorry, not too familiar with webpack and stuff.  Wouldn't it be easier to just include this import?

webmaster128:
Starting with CosmJS 0.28.0 we'll use @noble/hashes for ahh the hashing stuff, which des not cause Buffer related issues anymore.

PR #916, opened 2021-10-29
Remove unneeded Buffer from Ripemd160

webmaster128:


webmaster128:
Works with the type system but not at runtime :/

PR #922, opened 2021-11-07
Add redelegateTokens to singingstargateclient.ts

msteiner96:


webmaster128:
Nice, thanks. Do you need this released as 0.26.x any time soon?

msteiner96:
Not necessarily. While building my tool I just realized it was missing and added it :)

webmaster128:
Could you add a test for that? I would not require it for merge because we don't have tests for the other staking methods as well. But it would be great to increase coverage.

msteiner96:
Sure thing! Just checked that other tests are missing too so I skipped that :D Gimme some time, gonna add that

webmaster128:
For inspiration: the query tests for staking and distribution might help for checking if the operation was successful:
- https://github.com/cosmos/cosmjs/blob/main/packages/stargate/src/queries/staking.spec.ts
- https://github.com/cosmos/cosmjs/blob/main/packages/stargate/src/queries/distribution.spec.ts

msteiner96:
@webmaster128 To write a test for redelegation, I need 2 validators to test. But in `packages/stargate/src/testutils.spec.ts` there's only one defined. Do you want me to extend that?

webmaster128:
Oh, good point. The problem is the validator does not only need to be added in the test file. Instead it need to exists in the test network which is running as a backend for the tests. The validator is created here: https://github.com/cosmos/cosmjs/blob/v0.26.4/scripts/simapp42/template/.simapp/config/genesis.json#L533-L558.

We can probably not run the network with multiple validators without massive overhead in the docker setups. But maybe we can register a second inactive validator?

webmaster128:
We give up the idea of those wrapper function in general. We can never keep up with all the possible messages this way.

PR #931, opened 2021-11-17
Add simulation support

### LINKED ISSUE #828: Explore simulation for help in choosing gas limits
After https://github.com/cosmos/cosmjs/issues/827

webmaster128:
Closes #828

### DIFF
--- a/packages/cli/examples/simulate.ts
+++ b/packages/cli/examples/simulate.ts
@@ -1,4 +1,4 @@
-import { makeCosmoshubPath } from "@cosmjs/amino";
+import { coins, makeCosmoshubPath } from "@cosmjs/amino";
 import { DirectSecp256k1HdWallet } from "@cosmjs/proto-signing";
 import {
   assertIsBroadcastTxSuccess,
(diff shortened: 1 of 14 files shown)


packages/cosmwasm-stargate/src/signingcosmwasmclient.spec.ts
```diff
@@ -71,6 +71,30 @@ describe("SigningCosmWasmClient", () => {
...
+        typeUrl: "/cosmwasm.wasm.v1.MsgExecuteContract",
+        value: MsgExecuteContract.fromPartial({
+          sender: alice.address0,
+          contract: deployedHackatom.instances[0].address,
+          msg: toUtf8(`{"release":{}}`),
+          funds: [],
+        }),
+      };
+      const memo = "Go go go";
+      const gasUsed = await client.simulate(alice.address0, [executeContractMsg], memo);
```
ethanfrey, line 91: very cool!

ethanfrey (commented):
Very useful helper.

I like to not auto-set gas inside the signer, but let a caller easily get simulation estimates and set gas in the app

webmaster128:
> I like to not auto-set gas inside the signer, but let a caller easily get simulation estimates and set gas in the app

Yeah, this is the next question. I think the flow is very inconvenient. I'm thinking about an "auto" option that does the simulation internally and use a constant multiplier but still allow a custom gas limit. I'll write an example that shows how the current diff can be used.

webmaster128:
@ethanfrey could you have a look at `packages/cli/examples/simulate.ts`? There you see how manul gas is easy if you create the messages manually anyways (like e.g. ts-relayer), but annoying if you use higher level methods like "sendTokens" or CosmWasm "execute" that create the messages internally. The API we have now allows both. The only limitation is that the "auto" multiplier is fixed to 1.3.

packages/cli/examples/simulate.ts
```diff
@@ -0,0 +1,123 @@
...
+    value: {
+      fromAddress: account.address,
+      toAddress: recipient,
+      amount: amount,
+    },
+  };
+  const memo = "With simulate";
+  const gasEstimation = await client.simulate(account.address, [sendMsg], memo);
+  const fee = calculateFee(Math.round(gasEstimation * 1.3), gasPrice);
+  const result = await client.sendTokens(account.address, recipient, amount, fee, memo);
```
ethanfrey, line 62: Wow, this next to last argument can be:

* "auto"
* number -> multiplier after simulation
* `StdFee` -> caller fixed gas/fee

The top two using a gas price set in the constructor.

Is that correct?

packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
```diff
@@ -367,7 +391,7 @@ export class SigningCosmWasmClient extends CosmWasmClient {
     senderAddress: string,
     recipientAddress: string,
     amount: readonly Coin[],
-    fee: StdFee,
+    fee: StdFee | "auto" | number,
```
ethanfrey, line 394: Okay, nice arg here... very flexible and I hope the caller uses TS

ethanfrey (approved):
nice stuff

packages/cli/examples/simulate.ts
webmaster128: Exactly

PR #937, opened 2021-11-19
Awaiting broadcast of async transactions fails 

Cashmaney:
When trying to send an async transaction via Tendermint34Client it will return an error when waiting for broadcastTxAsync to return.

Can be triggered by calling 

```javascript
const result = await tmClient.broadcastTxAsync({tx: txBytes});
```

Which results in:

```
e: TypeError: Cannot read property 'decodeBroadcastTxSync' of undefined
(node:38652) UnhandledPromiseRejectionWarning: ReferenceError: result is not defined
    at run (C:\Users\Itzik\WebstormProjects\signingtest\main_sg.js:62:28)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
(node:38652) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1
)
(node:38652) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Even though the broadcast itself is successful. This small PR fixes this issue

Tested with Secret Network Testnet running Tendermint version v0.34.14

### DIFF
--- a/packages/tendermint-rpc/src/tendermint34/adaptor/responses.ts
+++ b/packages/tendermint-rpc/src/tendermint34/adaptor/responses.ts
@@ -829,7 +829,7 @@ export class Responses {
   }
 
   public static decodeBroadcastTxAsync(response: JsonRpcSuccessResponse): responses.BroadcastTxAsyncResponse {
-    return this.decodeBroadcastTxSync(response);
+    return decodeBroadcastTxSync(response.result as RpcBroadcastTxSyncResponse);
   }
 
   public static decodeBroadcastTxCommit(


webmaster128:
Nice find! This affects pretty much all versions of CosmJS and can easily be reproduced in tests. It seels like this function was never used and **never tested**. Shame on us. It also shows that you should not use classes for namespacing in JavaScript. We don't write Java.

I'll write a patch that for 0.26.x and add the missing tests.

webmaster128:
Fix released as part of 0.26.5

PR #960, opened 2021-12-06
Replace ripemd160 dependencies with @noble/hashes

webmaster128:


webmaster128:
In order to remove ripemd160 from the dependency tree for most users, we also need to update the hashing dependencies of @confio/ics23.

PR #966, opened 2021-12-16
Re-implement bip39

webmaster128:
This partially reverts https://github.com/cosmos/cosmjs/pull/804 (going back to our own implementation) and inlines much more bip39 functionality.


webmaster128:
This should reduce the bundle size a lot as seen in https://github.com/cosmos/cosmjs/issues/904:
137797752-1822dba0-8fda-4116-91fc-693270c96b39

JakeHartnell:
What else is needed here? Would be great to reduce build size. How best to help get this out of Draft?

webmaster128:
This is basically ready. I just got side-tracked and did not consider this a release blocker for 0.27.0.

One thing I wanted to add is a pure-js implementation of pbkdf2 in order to support environments that do not have subtle or the Node.js crypto module. But we can also do a cry test and see if someone complains.

PR #972, opened 2022-01-09
added show address in ledger signer

blackpanther1881:


puneet2019:
@webmaster128 The function is required by the ledger team as one of the requirements to publish the ledger live app.
https://developers.ledger.com/docs/nano-app/publish-introduction/
> The application is compatible with a wallet app with a GUI (a CLI is not enough)

It shows the address in the ledger device. Had added it to our fork  [here](https://github.com/persistenceOne/ledger-amino/blob/main/src/ledgersigner.ts#L41). Since cosmjs then did not have support for other coin-types. 
Would like to have it in cosmjs and remove the dependency on the fork.

packages/ledger-amino/src/launchpadledger.ts
```diff
@@ -166,6 +167,17 @@ export class LaunchpadLedger {
     await this.verifyCosmosAppIsOpen();
   }
 
+  public async verifyAddress(hdPath: HdPath): Promise<AddressAndPublicKeyResponse | ErrorResponse> {
+    await this.verifyDeviceIsReady();
+    assert(this.app, `${this.ledgerAppName} Ledger App is not connected`);
```
webmaster128, line 172: `this.app` is set in the constructor and is non-optional. So we don't need the check here.

Also after `await this.verifyDeviceIsReady();` multiple things are checked already based on `this.app`.

packages/ledger-amino/src/launchpadledger.ts
```diff
@@ -4,7 +4,8 @@ import { fromUtf8 } from "@cosmjs/encoding";
 import { assert } from "@cosmjs/utils";
 import Transport from "@ledgerhq/hw-transport";
 import CosmosApp, {
-  AppInfoResponse,
+  AddressAndPublicKeyResponse,
+  AppInfoResponse, ErrorResponse,
```
webmaster128, line 8: Where is `ErrorResponse` defined? I did not find anything in https://github.com/Zondax/ledger-cosmos-js.

Is this the same we create as `export interface LedgerAppErrorResponse` a few lines below?

packages/ledger-amino/src/launchpadledger.ts
```diff
@@ -166,6 +167,17 @@ export class LaunchpadLedger {
     await this.verifyCosmosAppIsOpen();
   }
 
+  public async verifyAddress(hdPath: HdPath): Promise<AddressAndPublicKeyResponse | ErrorResponse> {
+    await this.verifyDeviceIsReady();
+    assert(this.app, `${this.ledgerAppName} Ledger App is not connected`);
+
+    const hdPathToUse = hdPath;
```
webmaster128, line 174: As in the other methods, `hdPath` should be optional for consistency because the `LaunchpadLedger` instance already has paths configured.

There might be a good reason to remove the default, but then we should do it for all methods.

packages/ledger-amino/src/launchpadledger.ts
```diff
@@ -166,6 +167,17 @@ export class LaunchpadLedger {
     await this.verifyCosmosAppIsOpen();
   }
 
+  public async verifyAddress(hdPath: HdPath): Promise<AddressAndPublicKeyResponse | ErrorResponse> {
```
webmaster128, line 170: The errors are converted to exceptions, right? So this becomes

```suggestion
  public async verifyAddress(hdPath: HdPath): Promise<AddressAndPublicKeyResponse> {
```

packages/ledger-amino/src/launchpadledger.ts
```diff
@@ -200,3 +212,4 @@ export class LaunchpadLedger {
     }
   }
 }
+
```
webmaster128, line 215: Why this diff? Anyways, linter will remove it

packages/ledger-amino/src/ledgersigner.ts
```diff
@@ -37,6 +38,10 @@ export class LedgerSigner implements OfflineAminoSigner {
     return this.accounts;
   }
 
+  public async showAddress(path: HdPath): Promise<AddressAndPublicKeyResponse | ErrorResponse> {
```
webmaster128, line 41:
```suggestion
  public async showAddress(path: HdPath): Promise<AddressAndPublicKeyResponse> {
```

webmaster128 (commented):
Looks good overall, thank you.

Could you run `yarn && yarn lint-fix` to take care of formatting?

I'm not sure if we really want to expose this method in the signer implementation. Maybe we better provide public access to `ledger: LaunchpadLedger` to allow accessing all the detailed methods. But let's clean up some basics first.

vishrutsingh:
@webmaster128 could you please review the latest changes. Thank you.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to
 - @cosmjs/amino: Added `StdTx`, `isStdTx` and `makeStdTx` and removed them from
   @cosmjs/launchpad. They are re-exported in @cosmjs/launchpad for backwards
   compatibility.
+- @cosmjs/ledger-amino: Added `showAddress` in ledgerSigner.
```
webmaster128, line 25:
```suggestion
- @cosmjs/ledger-amino: Added `showAddress` in LedgerSigner.
```

packages/ledger-amino/src/ledgersigner.ts
```diff
@@ -37,6 +38,10 @@ export class LedgerSigner implements OfflineAminoSigner {
     return this.accounts;
   }
 
+  public async showAddress(path: HdPath): Promise<AddressAndPublicKeyResponse> {
```
webmaster128, line 41: This should get its own return type that we can make public. `AddressAndPublicKeyResponse` depends on an external dependency, uses sname_case fields and is not exported from @cosmjs/ledger-amino.

Ideally we have this type on @cosmjs/amino-ledger:

```ts
export interface AddressAndPubkey {
    readonly address: string;
    readonly pubkey: Secp256k1Pubkey;
}
```

webmaster128 (commented):
Thanks.

This still needs a linter run: `yarn lint-fix`

packages/ledger-amino/src/ledgersigner.ts
vishrutsingh: @webmaster128 should this be done by declaring a new module for AddressAndPubkey or changing the existing interface?

webmaster128: The existing interface cannot be changed because it is the format we get from `cosmos-ledger-js`, which is an external dependency.

You can just create the new interface as mentioned above in the same file.

vishrutsingh: @webmaster128 ive added the interface to the signer itself, please let me know if this is the approach?

packages/ledger-amino/src/launchpadledger.ts
```diff
@@ -166,6 +167,16 @@ export class LaunchpadLedger {
...
   }
 
+  public async verifyAddress(hdPath?: HdPath): Promise<AddressAndPublicKeyResponse> {
+    await this.verifyDeviceIsReady();
+
+    const hdPathToUse = hdPath || this.hdPaths[0];
+    // ledger-cosmos-js hardens the first three indices
+    const response = await this.app.showAddressAndPubKey(unharden(hdPathToUse), this.prefix);
+    this.handleLedgerErrors(response);
+    return response as AddressAndPublicKeyResponse;
```
webmaster128, line 177: We trun all errors to exceptions in the first line. So we know there is no error anymore in the result and we can use `AddressAndPubkey` instead of `AddressAndPublicKeyResponse`.

packages/ledger-amino/src/ledgersigner.ts
```diff
@@ -37,6 +43,19 @@ export class LedgerSigner implements OfflineAminoSigner {
     return this.accounts;
   }
 
+  public async showAddress(path: HdPath): Promise<AddressAndPubkey> {
+    const response = await this.ledger.verifyAddress(path);
+
+    if (response.error_message) {
+      throw new Error(response.error_message);
+    }
```
webmaster128, line 51: This is not needed if the lower level methods thows on error

webmaster128:
For some reason, circle CI does not build and test this anymore :/ I will check and finalize after the interface change in LaunchpadLedger

packages/ledger-amino/src/ledgersigner.ts
```diff
@@ -37,6 +43,19 @@ export class LedgerSigner implements OfflineAminoSigner {
...
+
+    return {
+      address: response.address,
+      pubkey: response.compressed_pk,
```
webmaster128, line 55: This line does not compile because compressed_pk is a Buffer and pubkey is a struct

webmaster128:
Thank you for your contribution. I'll wrap this up now in a separate PR.

I'm closing here to avoid duplicate work.

webmaster128:
Could you have a look at https://github.com/cosmos/cosmjs/pull/1008? I want to release this with CosmJS 0.27.1 very soon. This integrates the new functionality also in the node.js and web demo, so it can easily be tested.

PR #981, opened 2022-01-15
create StargateClient with proxy rpc provider

daoauth:
Stargate client connect with  proxy rpcClient like a ethereum and metamask.

```typescript
const proxyProvider = new ProxyRpcClinet();
const client = await StargateClient.connect(proxyProvider)';
```

### DIFF
--- a/packages/tendermint-rpc/src/tendermint34/tendermint34client.ts
+++ b/packages/tendermint-rpc/src/tendermint34/tendermint34client.ts
@@ -21,7 +21,7 @@ export class Tendermint34Client {
    */
   public static async connect(url: string | RpcClient): Promise<Tendermint34Client> {
     let rpcClient: RpcClient;
-    if (typeof url === 'string') {
+    if (typeof url === "string") {
       const useHttp = url.startsWith("http://") || url.startsWith("https://");
       rpcClient = useHttp ? new HttpClient(url) : new WebsocketClient(url);
     } else {
--- a/packages/tendermint-rpc/src/tendermint34/tendermint34client.ts
+++ b/packages/tendermint-rpc/src/tendermint34/tendermint34client.ts
@@ -19,9 +19,14 @@ export class Tendermint34Client {
    *
    * Uses HTTP when the URL schema is http or https. Uses WebSockets otherwise.
    */
-  public static async connect(url: string): Promise<Tendermint34Client> {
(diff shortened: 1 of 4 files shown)


webmaster128:
Thank you. Diff looks plausible. Could you explain the use case why you need this?

daoauth:
> Thank you. Diff looks plausible. Could you explain the use case why you need this?

I am developing a multi-chain wallet that supports Cosmos, Ethereum, and many other blockchains. When developing the DAPP, I thought it would be good to set it up comfortably like Ethereum in Cosmos. With simple setup, you can use the stable rpc supported by the wallet together.

```typescript
// https://gist.github.com/webmaster128/8444d42a7eceeda2544c8a59fbd7e1d9
const rpcEndpoint = "https://rpc.my_tendermint_rpc_node";
const client = await SigningStargateClient.connectWithSigner(rpcEndpoint, wallet);
```

```typescript
// https://medium.com/valist/how-to-connect-web3-js-to-metamask-in-2020-fee2b2edf58a
const proxyProvider = window.cosmos; // injection from extension wallet
const client = await StargateClient.connect(proxyProvider);
```

packages/tendermint-rpc/src/tendermint34/tendermint34client.ts
```diff
@@ -19,9 +19,14 @@ export class Tendermint34Client {
    *
    * Uses HTTP when the URL schema is http or https. Uses WebSockets otherwise.
    */
-  public static async connect(url: string): Promise<Tendermint34Client> {
-    const useHttp = url.startsWith("http://") || url.startsWith("https://");
-    const rpcClient = useHttp ? new HttpClient(url) : new WebsocketClient(url);
+  public static async connect(url: string | RpcClient): Promise<Tendermint34Client> {
```
catrias, line 22: I suggest renaming the parameter. `rpcClient = url` looks weird

webmaster128:
Thanks a lof for your contribution @daoauth. After careful consideration of the current ecosystem and the path in front of us (looking at [grpc/grpc-web](https://github.com/cosmos/cosmjs/issues/885) as well as upcoming Tendermint RPC versions with and without WebSocket support), I think it's much safer and future proof to not try to standardize endpoints in any way. Keeping the interface pivate and even removing the HttpClient/WebsocketClient exports from the public interface better prepares us for this path.

I hope there are other ways to solve your problems. If not, feel free to open a ticket.

PR #999, opened 2022-01-20
Fix ambiguity in the IBC/TransferMsg/TimeoutTimestamp docs

assafmo:
Fix ambiguity in the IBC/TransferMsg/TimeoutTimestamp docs. 

Also see https://github.com/cosmos/ibc-go/pull/715

### DIFF
--- a/packages/stargate/src/aminomsgs.ts
+++ b/packages/stargate/src/aminomsgs.ts
@@ -377,7 +377,7 @@ export interface AminoMsgTransfer extends AminoMsg {
      */
     readonly timeout_height: AminoHeight;
     /**
-     * Timeout timestamp (in nanoseconds). The timeout is disabled when set to 0.
+     * Timeout timestamp in absolute nanoseconds since unix epoch. The timeout is disabled when set to 0.
      *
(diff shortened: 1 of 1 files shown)


packages/stargate/src/aminomsgs.ts
```diff
@@ -377,7 +377,7 @@ export interface AminoMsgTransfer extends AminoMsg {
      */
     readonly timeout_height: AminoHeight;
     /**
-     * Timeout timestamp (in nanoseconds). The timeout is disabled when set to 0.
+     * Timeout timestamp in absolute nanoseconds since unix epoch. The timeout is disabled when set to 0.
```
webmaster128, line 380:
```suggestion
     * Timeout timestamp in nanoseconds since Unix epoch. The timeout is disabled when set to 0.
```

webmaster128 (commented):
Thanks

packages/stargate/src/aminomsgs.ts
assafmo: Cool! I didn't know you can suggest changes. Thanks.

webmaster128 (approved):

PR #1013, opened 2022-01-26
Upgrade log4js

webmaster128:


ethanfrey:
I first read Log4J and was 😱 
Glad this is not full of holes....

PR #1015, opened 2022-01-26
Remove Launchpad client

### LINKED ISSUE #947: Remove Launchpad client
The Launchpad client in @cosmjs/launchpad connects to Cosmos SDK 0.37-0.39 networks. Nobody seems to be running such networks anymore. Let's remove the code.

If you need Launchpad support, please let me know.

webmaster128:
Closes #947

- [x] Remove Launchpad support from faucet
- [x] Rewrite `examples/delegate.ts` to use the stargate backend
- [x] Rewrite `examples/local_faucet.ts` to use the stargate backend and adjust `packages/cli/README.md`
- [x] Remove Launchpad support from cli
- [x] Remove launchpad tests from CI
- [x] Remove launchpad package
- [x] Remove launchpad scripts
- [x] Update references in docs


webmaster128 (commented):
Code changes look good.

But wh are there so many new zip files if you only removed dependenciey? Also why did you change `checksumBehavior`?

msteiner96:
> Code changes look good.
> 
> But wh are there so many new zip files if you only removed dependenciey? Also why did you change `checksumBehavior`?

Had to delete yarn cache and `yarn install` again because tests didn't work anymore (checksum errors). If there's a better solution, please let me know. 

`checksumBehavior`was changed for local debugging reasons. Changed it back now.

ethanfrey:
Why is `.yarn/cache` committed to the repo anyway?

webmaster128:
> Why is `.yarn/cache` committed to the repo anyway?

That's the best practice way to handle dependencies in yarn 2+. This way, you always have all dependencies available in your repo and don't need to download them. They are zipped and stored in git LFS. "cache" is a questionable name though.

ethanfrey:
> That's the best practice way to handle dependencies in yarn 2+. This way, you always have all dependencies available in your repo and don't need to download them.

Let's download the internet in case we need it 😄

webmaster128:
I ran into all kind of mess trying to check out this PR. I could fix it as follows:

```
git checkout -b rm-launchpad-v2
wget https://patch-diff.githubusercontent.com/raw/cosmos/cosmjs/pull/1015.diff
git apply --exclude=".yarn/cache/*" 1015.diff
```

Then you can `yarn && yarn build` with no change to the zip files at all. Will follow-up with a fresh PR.

PR #1022, opened 2022-01-31
Add height/gasWanted/gasUsed values to all result types of CosmWasm client

webmaster128:
- [x] Implementation
- [x] Tests
- [x] CHANGELOG


webmaster128 (commented):
Thanks

Issue #980, opened 2022-01-14
Cosmjs failes to parse blocks with double sign evidence 


JoeKash:
On vega-testnet when we call `stargateClient.getBlock(height)` for block '7900625'

We get a parsing error from tendermint34 package -
```
Error: Value must not be undefined 
at assertSet (node_modules/@cosmjs/tendermint-rpc/build/tendermint34/encodings.js:13:15) 
at Object.assertNotEmpty (node_modules/@cosmjs/tendermint-rpc/build/tendermint34/encodings.js:107:5) 
at decodeEvidence (node_modules/@cosmjs/tendermint-rpc/build/tendermint34/adaptor/responses.js:329:55)
```

We are running v0.27

webmaster128:
Thanks for the report. Can you provide a Tendermint RPC endpoint that we can use to reproduce?

JoeKash:
Thanks @webmaster128 I think cosmos has a public testnet rpc -
https://rpc.testnet.cosmos.network

webmaster128:
https://rpc.testnet.cosmos.network/block?height=7900625 tells me _height 7900625 must be less than or equal to the current blockchain height 1250979_. So those seem to be different networks.

webmaster128:
There is https://vega-rpc.interchain.io from https://github.com/cosmos/vega-test/tree/master/public-testnet but it tells me _"height 7900625 is not available, lowest height is 8197001"_.

ethanfrey:
I second the use of `any` and linking to this issue.
I bet this will change in tendermint 0.35 and again in 0.36 and let's just leave it `any` til someone has trouble with that

JoeKash:
Thanks for digging this up @webmaster128 

We mostly care about the client not throwing an exception we can work around not having type safety so type `any` should work great.

webmaster128:
This is addressed in https://github.com/cosmos/cosmjs/pull/1012. Would love if someone could have a look at that PR.

webmaster128:
This @cosmjs/cli script can be used to reproduce the bug an confirm the fix:

```ts
import { StargateClient } from "@cosmjs/stargate";
import { Tendermint34Client } from "@cosmjs/tendermint-rpc";

const rpcEndpoint = "http://198.50.215.1:36657"

// High level client
const client = await StargateClient.connect(rpcEndpoint);
for (const height of [7900624, 7900625]) {
  console.log(`Getting block ${height} …`)
  const block = await client.getBlock(7900626);
[...]

PR #1023, opened 2022-02-01
Update Block Interface to make evidence non-optional

### LINKED ISSUE #1011: Make Block.evidence non-optional in Tendermint client
In

```
export interface Block {
  readonly header: Header;
  /**
   * For the block at height 1, last commit is not set.
   */
  readonly lastCommit: Commit | null;
  readonly txs: readonly Uint8Array[];
  readonly evidence?: readonly Evidence[];
}
```

the `evidence` field is always set, even if an empty array. There is no need to differentiate between an empty array and undefined.

msteiner96:
Resolves #1011 

`export type Evidence = any;` unchanged as mentioned in #980 .

### DIFF
--- a/packages/tendermint-rpc/src/tendermint34/responses.ts
+++ b/packages/tendermint-rpc/src/tendermint34/responses.ts
@@ -227,8 +227,7 @@ export interface Block {
    */
   readonly lastCommit: Commit | null;
   readonly txs: readonly Uint8Array[];
-  // This field becomes non-optional in 0.28 (https://github.com/cosmos/cosmjs/issues/1011)
-  readonly evidence?: readonly Evidence[];
+  readonly evidence: readonly Evidence[];
 }
 
 /**


webmaster128 (commented):
API change looks good. Could fix the implementation (see CI fail)?

Can you add an entry to CHANGELOG.md for this?

packages/tendermint-rpc/src/tendermint34/adaptor/responses.ts
```diff
@@ -743,8 +743,8 @@ interface RpcBlock {
   };
   // It's currently unclear why the deep nesting is requied.
   // See https://github.com/tendermint/tendermint/issues/7697.
-  readonly evidence?: {
-    readonly evidence?: readonly RpcEvidence[];
+  readonly evidence: {
+    readonly evidence: readonly RpcEvidence[];
```
webmaster128, line 747: Ah, this should remain unchanged. The private type `RpcBlock` is what we expect from the Tendermint API. Array encoding coming from Go can be strange and unexpected. Empty arrays are often omitted.

Better turn undefined to an empty array in this code:

```ts
evidence: data.evidence?.evidence,
```

webmaster128 (approved):

PR #1024, opened 2022-02-01
Let Registry constructor argument override default types

### LINKED ISSUE #994: Let Registry constructor argument override default types
With the addition systen, the constructor argument and the `register` method do the same and there is no way to get rid of `Coin` and `MsgSend`.

msteiner96:
Resolves #994

### DIFF
--- a/packages/proto-signing/src/registry.ts
+++ b/packages/proto-signing/src/registry.ts
@@ -95,12 +95,13 @@ export class Registry {
    * change the `customTypes` argument to override the default types if set.
    * See https://github.com/cosmos/cosmjs/issues/994
    */
-  public constructor(customTypes: Iterable<[string, GeneratedType]> = []) {
+  public constructor(customTypes?: Iterable<[string, GeneratedType]>) {
     const { cosmosCoin, cosmosMsgSend } = defaultTypeUrls;
-    this.types = new Map<string, GeneratedType>([
+      this.types = customTypes ? new Map<string, GeneratedType>([
+        ...customTypes
+      ]) : new Map<string, GeneratedType>([
       [cosmosCoin, Coin],
       [cosmosMsgSend, MsgSend],
-      ...customTypes,
     ]);
   }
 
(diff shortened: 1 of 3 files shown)


packages/proto-signing/src/registry.ts
```diff
@@ -95,13 +95,14 @@ export class Registry {
...
-      [cosmosCoin, Coin],
-      [cosmosMsgSend, MsgSend],
-      ...customTypes,
-    ]);
+    this.types = customTypes
+      ? new Map<string, GeneratedType>([...customTypes])
+      : new Map<string, GeneratedType>([
+          [cosmosCoin, Coin],
+          [cosmosMsgSend, MsgSend],
+        ]);
```
webmaster128, line 105: 👍

packages/stargate/src/signingstargateclient.ts
```diff
@@ -72,6 +72,7 @@ import { calculateFee, GasPrice } from "./fee";
 import { DeliverTxResponse, StargateClient } from "./stargateclient";
 
 export const defaultRegistryTypes: ReadonlyArray<[string, GeneratedType]> = [
+  ["/cosmos.base.v1beta1.Coin", Coin],
```
webmaster128, line 75: Here we also need `/cosmos.bank.v1beta1.MsgSend`. The `MsgMultiSend` we have below is an advanced version of that (multiple inputs and multiple outputs).

packages/proto-signing/src/registry.ts
```diff
@@ -95,13 +95,14 @@ export class Registry {
    * change the `customTypes` argument to override the default types if set.
    * See https://github.com/cosmos/cosmjs/issues/994
    */
-  public constructor(customTypes: Iterable<[string, GeneratedType]> = []) {
+  public constructor(customTypes?: Iterable<[string, GeneratedType]>) {
```
webmaster128, line 98: Can you read through the doc comment and update the parts that change with this PR?

webmaster128 (approved):

PR #1028, opened 2022-02-07
Set wasmd to v.0.23.0

### LINKED ISSUE #1017: Upgrade wasmd to 0.22

msteiner96:
Resolves #1017

### DIFF
--- a/scripts/wasmd/env
+++ b/scripts/wasmd/env
@@ -1,5 +1,5 @@
 # Choose from https://hub.docker.com/r/cosmwasm/wasmd/tags
 REPOSITORY="cosmwasm/wasmd"
-VERSION="v0.21.0"
+VERSION="v0.23.0"
 
 CONTAINER_NAME="wasmd"


webmaster128:
Could you have a look at the CI fails? The problem seems simple to resolve:

```
out of gas in location: WriteFlat; gasWanted: 80000, gasUsed: 80293: out of gas
    at assertIsDeliverTxSuccess (/home/circleci/project/packages/stargate/build/stargateclient.js:32:15)
    at main (/home/circleci/project/scripts/wasmd/send_first.js:31:3)
```

We use 80_000 in a few places as the default fee for a bank send. This seems to be a little low. We can just increase it consistently in all places. Many wallets use 200_000 for that these days, which is plenty.

webmaster128:
Now the expected contract addresses fail in tests. This is because in the upgrade from Cosmos SDK 0.42 to 0.45, contract addresses were changed from a 20 byte format to a 32 byte format. This only affects the addresses of instantiated contracts, not the regular user accounts.

You see the new addresses when running `./scripts/wasmd/start.sh && ./scripts/wasmd/init.sh` locally. They need to be updated in the testing constants.

PR #1039, opened 2022-02-14
Adding errors to unsupported TypeUrls

### LINKED ISSUE #1026: Add authz and feegrant messages to default types
authz and feegrant [were](https://github.com/confio/cosmjs-types/pull/17) [added](https://github.com/confio/cosmjs-types/pull/26) to the code generator already. They add the new message types:
- cosmos.authz.v1beta1.MsgGrant
- cosmos.authz.v1beta1.MsgExec
- cosmos.authz.v1beta1.MsgRevoke
- cosmos.feegrant.v1beta1.MsgGrantAllowance
- cosmos.feegrant.v1beta1.MsgRevokeAllowance

They should be added to the Stargate registry.

For the AminoTypes it seems like no Amino JSON signing is implemented on-chain. So we should create a specific error message for that, explaining why it does not work.

msteiner96:
Resolves #1026


packages/stargate/src/aminotypes.ts
```diff
@@ -539,6 +548,12 @@ export class AminoTypes {
   }
 
   public toAmino({ typeUrl, value }: EncodeObject): AminoMsg {
+    if (this.unsupportedTypes.includes(typeUrl)) {
+      throw new Error(
+        `The message type '${typeUrl}' cannot be signed using the Amino JSON sign mode because this is not implemented on-chain.`,
+      );
+    }
```
webmaster128, line 555: Nice error handling

packages/stargate/src/aminotypes.ts
```diff
@@ -526,6 +526,15 @@ export interface AminoTypesOptions {
...
   private readonly register: Record<string, AminoConverter>;
 
+  // Those message types cannot be signed using the Amino JSON
+  private readonly unsupportedTypes: string[] = [
+    "cosmos.authz.v1beta1.MsgGrant",
+    "cosmos.authz.v1beta1.MsgExec",
+    "cosmos.authz.v1beta1.MsgRevoke",
+    "cosmos.feegrant.v1beta1.MsgGrantAllowance",
+    "cosmos.feegrant.v1beta1.MsgRevokeAllowance",
+  ];
```
webmaster128, line 536: Can we move those to the config instead of hardcode them?

E.g. if we change

```diff
export class AminoTypes {
-  private readonly register: Record<string, AminoConverter>;
+  private readonly register: Record<string, AminoConverter | "not_implemented_on_chain">;
```

then we can set them explicitly to `"not_implemented_on_chain"`. This allows users
1. to override the value when they do implement it on chain
2. Add their own types with `"not_implemented_on_chain"`

webmaster128 (commented):
A test or two that trigget the new error would be nice

packages/stargate/src/aminotypes.ts
msteiner96: Sure, but what config do you mean? Don't we already have such an error handling for unsupported/not implemented amino messages with 

```ts
      throw new Error(
        `Type URL '${typeUrl}' does not exist in the Amino message type register. ` +
          "If you need support for this message type, you can pass in additional entries to the AminoTypes constructor. " +
          "If you think this message type should be included by default, please open an issue at https://github.com/cosmos/cosmjs/issues.",
      );
```
?

webmaster128: I mean right now you hardcode `private readonly unsupportedTypes: string[] = [`. This could be made part of the configurable part, i.e.

```ts
function createDefaultTypes(prefix: string): Record<string, AminoConverter | "not_implemented_on_chain"> {
  return {
    // authz

    "/cosmos.authz.v1beta1.MsgGrant": "not_implemented_on_chain", // this setting produces a different error message than a missing typeUrl
    "/cosmos.authz.v1beta1.MsgExec": "not_implemented_on_chain",
    "/cosmos.authz.v1beta1.MsgRevoke": "not_implemented_on_chain",

    // bank

    "/cosmos.bank.v1beta1.MsgSend": {
      aminoType: "cosmos-sdk/MsgSend",
      toAmino: ({ fromAddress, toAddress, amount }: MsgSend): AminoMsgSend["value"] => ({
        from_address: fromAddress,
        to_address: toAddress,
        amount: [...amount],
      }),
      fromAmino: ({ from_address, to_address, amount }: AminoMsgSend["value"]): MsgSend => ({
        fromAddress: from_address,
        toAddress: to_address,
        amount: [...amount],
      }),
    },
```

packages/stargate/src/aminotypes.ts
```diff
@@ -539,6 +548,12 @@ export class AminoTypes {
...
 
   public toAmino({ typeUrl, value }: EncodeObject): AminoMsg {
+    if (this.unsupportedTypes.includes(typeUrl)) {
+      throw new Error(
+        `The message type '${typeUrl}' cannot be signed using the Amino JSON sign mode because this is not implemented on-chain.`,
+      );
+    }
+
     const converter = this.register[typeUrl];
     if (!converter) {
```
webmaster128, line 558: This would become

```suggestion
    const converter = this.register[typeUrl];
    if (converter === "not_implemented_on_chain") {
      // error message from above
    } else if (!converter) {
```

webmaster128 (commented):
Looks good so far. I see two things as open:
- Could you update this to latest main? (#1046 changed the code a bit)
- Could you rename "no implemented" to "not supported" in the error text and config type? This seems to be matching a bit better when the type is implemented but buggy.

PR #1040, opened 2022-02-14
Implement Query SlashingExtension

### LINKED ISSUE #927: Add support for slashing queries in stargate client
Stargate client is missing the support for slashing queries.

msteiner96:
Resolves #927 

- [x] Implementation
- [x] Tests


packages/stargate/src/stargateclient.ts
```diff
@@ -139,7 +141,7 @@ export interface PrivateStargateClient {
 export class StargateClient {
   private readonly tmClient: Tendermint34Client | undefined;
   private readonly queryClient:
-    | (QueryClient & AuthExtension & BankExtension & StakingExtension & TxExtension)
+    | (QueryClient & AuthExtension & BankExtension & StakingExtension & TxExtension & SlashingExtension)
```
webmaster128, line 144: I would only add the extensions that are needed for the higher level functions in the stargate client. The query client here is private and I think callers should create their own, depending on their needs and backend availabilities.

webmaster128 (approved):

PR #1042, opened 2022-02-14
Faucet can only be used once a day per address

### LINKED ISSUE #962: Faucet limit from draining
Hey,

Is there currently a way in the faucet to limit faucet users from draining it, like IP rate limiter per day or something like that, which would prevent somebody from spamming the faucet?

msteiner96:
Resolves #962

### DIFF
--- a/packages/faucet/src/api/webserver.ts
+++ b/packages/faucet/src/api/webserver.ts
@@ -14,14 +14,9 @@ export interface ChainConstants {
   readonly chainId: string;
 }
 
-export interface AddressEntry {
-  address: string;
-  date: number;
-}
-
 export class Webserver {
   private readonly api = new Koa();
-  private readonly addressCounter: AddressEntry[] = [];
+  private readonly addressCounter = new Map<string, Date>();
 
   public constructor(faucet: Faucet, chainConstants: ChainConstants) {
     this.api.use(cors());
@@ -70,9 +65,9 @@ export class Webserver {
             throw new HttpError(400, "Address is not in the expected format for this chain.");
           }
 
-          const addressUsed = this.addressCounter.find((x) => x.address === address);
-          if (addressUsed !== undefined) {
-            if (addressUsed.date + 24 * 3600 > Date.now()) {
+          const entry: Date | undefined = this.addressCounter.get(address);
+          if (entry !== undefined) {
+            if (entry.getDate() + 24 * 3600 > Date.now()) {
               throw new HttpError(
                 405,
                 "Too many request from the same address. Blocked to prevent draining. Please wait 24h and try it again!",
@@ -80,7 +75,7 @@ export class Webserver {
             }
           }
 
-          const availableTokens = await faucet.availableTokens();
+          const availableTokens: string[] = await faucet.availableTokens();
           const matchingDenom = availableTokens.find((availableDenom) => availableDenom === denom);
           if (matchingDenom === undefined) {
             throw new HttpError(422, `Token is not available. Available tokens are: ${availableTokens}`);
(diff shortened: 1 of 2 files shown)


packages/faucet/src/api/webserver.ts
```diff
@@ -14,8 +14,14 @@ export interface ChainConstants {
   readonly chainId: string;
 }
 
+export interface AddressEntry {
+  address: string;
+  date: number;
```
webmaster128, line 19: I think this makes the code a bit more intuitive to read

```suggestion
  lastUsed: Date;
```

packages/faucet/src/api/webserver.ts
```diff
@@ -73,6 +88,8 @@ export class Webserver {
 
           try {
             await faucet.credit(address, matchingDenom);
+            // Count addresses to prevent draining
+            this.addressCounter.push({ address: address, date: Date.now() });
```
webmaster128, line 92: This adds the same address multiple times to the array. But `const addressUsed = this.addressCounter.find((x) => x.address === address);` will only search for the first entry. So after 24h, no request will be blocked anymore.

We probably make our lives easier using a Map instead.

packages/faucet/src/api/webserver.ts
```diff
@@ -58,21 +59,32 @@ export class Webserver {
...
+          if (entry !== undefined) {
+            if (entry.getDate() + 24 * 3600 > Date.now()) {
+              throw new HttpError(
+                405,
+                "Too many request from the same address. Blocked to prevent draining. Please wait 24h and try it again!",
+              );
+            }
+          }
+
+          const availableTokens: string[] = await faucet.availableTokens();
```
webmaster128, line 78: No need to change this line as part of this PR, right?

```suggestion
          const availableTokens = await faucet.availableTokens();
```

packages/faucet/src/api/webserver.ts
```diff
@@ -58,21 +59,32 @@ export class Webserver {
...
           const creditBody = RequestParser.parseCreditBody(requestBody);
-
           const { address, denom } = creditBody;
 
           if (!isValidAddress(address, constants.addressPrefix)) {
             throw new HttpError(400, "Address is not in the expected format for this chain.");
           }
 
-          const availableTokens = await faucet.availableTokens();
+          const entry: Date | undefined = this.addressCounter.get(address);
```
webmaster128, line 68: Do you need to make this type explicit? It should be infered to `Date | undefined`

packages/faucet/src/api/webserver.ts
msteiner96: Was my IDE somehow... Gonna delete it

packages/faucet/src/api/webserver.ts
```diff
@@ -58,13 +59,22 @@ export class Webserver {
...
-
           const { address, denom } = creditBody;
 
           if (!isValidAddress(address, constants.addressPrefix)) {
             throw new HttpError(400, "Address is not in the expected format for this chain.");
           }
 
+          const entry = this.addressCounter.get(address);
+          if (entry !== undefined) {
+            if (entry.getDate() + 24 * 3600 > Date.now()) {
```
webmaster128, line 70: >  The getDate() method returns the day of the month for the specified date according to local time. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDate

Seems like this is not what we want here

webmaster128 (commented):
Could you add a CHANGELOG entry?

packages/faucet/src/api/webserver.ts
msteiner96: I'm stupid. Got confused. It's `getTime()` ofc

msteiner96:
> Could you add a CHANGELOG entry?

-- Done

webmaster128 (approved):

PR #1065, opened 2022-03-01
Adding faucet dockerfile

msteiner96:
Usage:

`docker build -f packages/faucet/Dockerfile .`


packages/faucet/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM node:14-alpine
```
webmaster128, line 1:
```suggestion
# Build:
#   docker build -t confio/cosmos-faucet:manual -f packages/faucet/Dockerfile .
# Run:
#   docker run --rm confio/cosmos-faucet:manual version

FROM node:14-alpine
```

packages/faucet/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+RUN apk add --update --no-cache alpine-sdk linux-headers build-base gcc libusb-dev python3 eudev-dev && ln -sf python3 /usr/bin/python
+RUN python3 -m ensurepip
+RUN pip3 install --no-cache --upgrade pip setuptools
+
+RUN yarn install && yarn run build
+
+EXPOSE 8000
+ENTRYPOINT ["/app/packages/faucet/bin/cosmos-faucet"]
+CMD ["node", "/app/packages/faucet/bin/cosmos-faucet"]
```
webmaster128, line 14: You only need one and since the user should not override the prefix, we use entrypoint

```suggestion
ENTRYPOINT ["yarn", "node", "/app/packages/faucet/bin/cosmos-faucet"]
```

webmaster128: Also note the yarn there

webmaster128:
Ah, could you please remove the file `docs/faucet.Dockerfile` which is superseded by this approach?

webmaster128 (commented):
Okay, this seesm to work.

Next up: the resulting image is 950 MB large because it contains all the things we install for the compilation. What we should do now is create two stages: a build stage and a runtime stage. The dependencies are only needed for the build stage.

You can see that here:
1. the build stage installes the dependencies: https://github.com/iov-one/iov-faucet/blob/master/Dockerfile#L1-L5
2. The runtime stage only copies what is needed at runtime, starting from a fresh image without app the apk and python installations: https://github.com/iov-one/iov-faucet/blob/master/Dockerfile#L7-L17

webmaster128 (commented):
650 MB – nice improvement.

With this setup there is more optimization potential, but no need to dig into that now.

webmaster128:
I added the build and publish instructions to https://github.com/cosmos/cosmjs/wiki/Maintenance#how-to-publish-to-npm-yarn-2-cosmjs-026. This can be automated better later on.

Issue #830, opened 2021-06-10
Build and publish faucet docker


webmaster128:
To make the faucet binary easily available to users, it would be nice to build and publish to DockerHub as part of the CI for tag builds.

The currect [Dockerfile template](https://github.com/cosmos/cosmjs/blob/main/docs/faucet.Dockerfile) uses the faucet from npm. We would not do that here. Instead we'd bundle some kind of local build.

webmaster128:
The faucet is published to https://hub.docker.com/r/confio/faucet. Unfortunately this is an ARM image because it was created on my ARM machine. We should build it on x86_64, which is the primary host (Linux servers).

PR #1068, opened 2022-03-02
Update faucet repo name to confio/faucet [skip ci]

webmaster128:
Part of #830

Let's use a repo name that is a little shorter


msteiner96 (approved):
LGTM

PR #1086, opened 2022-03-11
 getContracts(): Support more than 100 per code ID

### LINKED ISSUE #1077: getContracts(): Support more than 100 instances per code ID
See https://github.com/cosmos/cosmjs/blob/v0.28.0-rc1/packages/cosmwasm-stargate/src/cosmwasmclient.ts#L337-L341

msteiner96:
Resolves #1077

### DIFF
--- a/packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
+++ b/packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
@@ -275,9 +275,9 @@ describe("CosmWasmClient", () => {
       const expectedAddresses = deployedHackatom.instances.map((info) => info.address);
 
       // Test first 3 instances we get from scripts/wasmd/init.sh. There may me more than that in the result.
-      expect(result[0]).toEqual(expectedAddresses[0]);
+      expect(result[0]).toEqual(expectedAddresses[2]);
       expect(result[1]).toEqual(expectedAddresses[1]);
-      expect(result[2]).toEqual(expectedAddresses[2]);
+      expect(result[2]).toEqual(expectedAddresses[0]);
     });
   });
 
--- a/packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
+++ b/packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
@@ -275,9 +275,9 @@ describe("CosmWasmClient", () => {
       const expectedAddresses = deployedHackatom.instances.map((info) => info.address);
 
       // Test first 3 instances we get from scripts/wasmd/init.sh. There may me more than that in the result.
-      expect(result[0]).toEqual(expectedAddresses[2]);
+      expect(result[0]).toEqual(expectedAddresses[0]);
       expect(result[1]).toEqual(expectedAddresses[1]);
-      expect(result[2]).toEqual(expectedAddresses[0]);
+      expect(result[2]).toEqual(expectedAddresses[2]);
     });
   });
 
--- a/packages/cosmwasm-stargate/src/cosmwasmclient.ts
+++ b/packages/cosmwasm-stargate/src/cosmwasmclient.ts
@@ -342,20 +342,14 @@ export class CosmWasmClient {
    */
   public async getContracts(codeId: number): Promise<readonly string[]> {
     const allContracts = [];
-
-    try {
-      let startAtKey: Uint8Array | undefined = undefined;
-      do {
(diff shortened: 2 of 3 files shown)


packages/cosmwasm-stargate/src/cosmwasmclient.ts
```diff
@@ -334,10 +334,30 @@ export class CosmWasmClient {
...
+        const { contracts, pagination }: QueryContractsByCodeResponse =
+          await this.forceGetQueryClient().wasm.listContractsByCodeId(codeId, startAtKey);
+        const loadedContracts = contracts || [];
+        loadedContracts.reverse();
+        allContracts.unshift(...loadedContracts);
+        startAtKey = pagination?.nextKey;
+      } while (startAtKey?.length !== 0);
+    } catch (_e: any) {
+      throw new Error(_e);
+    }
```
webmaster128, line 358: No need to catch here are all if we throw the same thing later on. The `try {` is only needed in the React app.

packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
```diff
@@ -275,9 +275,9 @@ describe("CosmWasmClient", () => {
       const expectedAddresses = deployedHackatom.instances.map((info) => info.address);
 
       // Test first 3 instances we get from scripts/wasmd/init.sh. There may me more than that in the result.
-      expect(result[0]).toEqual(expectedAddresses[0]);
+      expect(result[0]).toEqual(expectedAddresses[2]);
       expect(result[1]).toEqual(expectedAddresses[1]);
-      expect(result[2]).toEqual(expectedAddresses[2]);
+      expect(result[2]).toEqual(expectedAddresses[0]);
```
webmaster128, line 280: I wonder if we really want to change the order here. It was a UI decision to sort from new to old in the code explorer. But here? Let's use the ordering we had before (ascending by creation time, the same way we get it from the node). The app can change the order if it wants to.

packages/cosmwasm-stargate/src/cosmwasmclient.ts
```diff
@@ -334,10 +334,30 @@ export class CosmWasmClient {
...
+    const allContracts = [];
+
+    try {
+      let startAtKey: Uint8Array | undefined = undefined;
+      do {
+        const { contracts, pagination }: QueryContractsByCodeResponse =
+          await this.forceGetQueryClient().wasm.listContractsByCodeId(codeId, startAtKey);
+        const loadedContracts = contracts || [];
+        loadedContracts.reverse();
+        allContracts.unshift(...loadedContracts);
```
webmaster128, line 353:
```suggestion
        allContracts.push(...loadedContracts);
```

This probably fixes the order

packages/cosmwasm-stargate/src/cosmwasmclient.ts
```diff
@@ -334,10 +334,30 @@ export class CosmWasmClient {
...
+        startAtKey = pagination?.nextKey;
+      } while (startAtKey?.length !== 0);
```
webmaster128, line 355: This code was a mistake by me. If `startAtKey` is undefined, the expression becomes truthy. We somehow need to map undefined and 0 to the same case.

webmaster128 (commented):
Two minor 💅, otherwise 👍

packages/cosmwasm-stargate/src/cosmwasmclient.ts
```diff
@@ -334,10 +334,24 @@ export class CosmWasmClient {
...
+   */
   public async getContracts(codeId: number): Promise<readonly string[]> {
-    // TODO: handle pagination - accept as arg or auto-loop
-    const { contracts } = await this.forceGetQueryClient().wasm.listContractsByCodeId(codeId);
-    return contracts;
+    const allContracts = [];
+    let startAtKey: Uint8Array | undefined = undefined;
+    do {
+      const { contracts, pagination }: QueryContractsByCodeResponse =
+        await this.forceGetQueryClient().wasm.listContractsByCodeId(codeId, startAtKey);
```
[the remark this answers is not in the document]
msteiner96: Nope - `'pagination' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.ts`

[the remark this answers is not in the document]
webmaster128: Okay, I can reproduce it but I don't understand what the error means and why we get it. Anyways, let's move on.

PR #1087, opened 2022-03-11
Add waiting time to timeout error message

### LINKED ISSUE #1066: Add waiting time to timeout error message
image

In an error case like this it would be interesting to understand how long the client waited for the inclusion of a tx. That allows for interpretation how realistic it is that the tx will make it into a block. E.g. if a chain has 3 second block times and the error says "after 60 seconds" you know that 20 blocks have been created without your transaction. If it says "after 10 seconds" and you have a 6 second block time, that can mean you did not make it in the first block but it is still likely the tx will show up.

msteiner96:
Resolves #1066

### DIFF
--- a/packages/cosmwasm-stargate/src/cosmwasmclient.ts
+++ b/packages/cosmwasm-stargate/src/cosmwasmclient.ts
@@ -270,7 +270,7 @@ export class CosmWasmClient {
         throw new TimeoutError(
           `Transaction with ID ${txId} was submitted but was not yet found on the chain. You might want to check later. There was a wait of ${
             timeoutMs / 10000
-          } seconds. Approx. ${timeoutMs / pollIntervalMs} blocks were created during this time.`,
+          } seconds.`,
           txId,
         );
       }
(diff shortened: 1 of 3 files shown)


packages/cosmwasm-stargate/src/cosmwasmclient.ts
```diff
@@ -264,7 +264,9 @@ export class CosmWasmClient {
     const pollForTx = async (txId: string): Promise<DeliverTxResponse> => {
       if (timedOut) {
         throw new TimeoutError(
-          `Transaction with ID ${txId} was submitted but was not yet found on the chain. You might want to check later.`,
+          `Transaction with ID ${txId} was submitted but was not yet found on the chain. You might want to check later. There was a wait of ${
+            timeoutMs / 10000
+          } seconds. Approx. ${timeoutMs / pollIntervalMs} blocks were created during this time.`,
```
webmaster128, line 269: We don't have any information on the block time unfortunately. The `pollIntervalMs` is only the time after which the client polls again to check for the block.

The two information we do have is: time out, time after which we gave up. This can be e.g. 10 seconds timeout and 12 seconds for the runtime. The time after we gave up is not available directly now. We have to store a start time in line 260.

msteiner96: So isn't the first sentence enough? `There was a wait of ${
            timeoutMs / 10000
          } seconds.`

webmaster128: Yeah, that's enough and simple

webmaster128 (approved):

PR #1102, opened 2022-03-29
feat: allow to specify an AccountParser inside StargateClient

RiccardoM:
This PR introduces a new `StargateClientOptions` interface that allows setting custom options for `StargateClient`.  

The most important option is the `accountParser` option, which must be an `AccountParser` instance and allows to provide the `StargateClient` a custom way of decoding Cosmos accounts. This is particularly useful to solve problems when reading account from chains that support custom account types (eg. the Desmos chain). This can solve problems like the one REStake is having: https://github.com/eco-stake/restake/issues/334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- @cosmjs/stargate: Added the ability to specify a custom account parser for `StargateClient`
+
 ### Fixed
 
 - @cosmjs/proto-signing: Add missing runtime dependencies @cosmjs/encoding and
--- a/packages/stargate/src/accounts.ts
+++ b/packages/stargate/src/accounts.ts
@@ -36,8 +36,13 @@ function accountFromBaseAccount(input: BaseAccount): Account {
 }
 
 /**
- * Takes an `Any` encoded account from the chain and extracts some common
- * `Account` information from it. This is supposed to support the most relevant
+ * Represents a generic function that takes an `Any` encoded account from the chain
+ * and extracts some common `Account` information from it.
+ */
+export type AccountParser = (any: Any) => Account;
+
+/**
+ * Basic implementation of AccountParser. This is supposed to support the most relevant
  * common Cosmos SDK account types. If you need support for exotic account types,
  * you'll need to write your own account decoder.
  */
--- a/packages/stargate/src/stargateclient.ts
+++ b/packages/stargate/src/stargateclient.ts
@@ -146,7 +146,7 @@ export class StargateClient {
     | (QueryClient & AuthExtension & BankExtension & StakingExtension & TxExtension)
     | undefined;
   private chainId: string | undefined;
-  private readonly accountParser: AccountParser | undefined;
+  private readonly accountParser: AccountParser;
 
   public static async connect(
     endpoint: string,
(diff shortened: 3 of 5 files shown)


RiccardoM:
> accountFromAny

Do you have an example of how you would implement this?

webmaster128:
> Do you have an example of how you would implement this?

Let's get in the user's perspective, imagine someone wants to create a client that supports `/desmos.profiles.v1beta1.Profile`. An implamentation then looks like this:

```ts
import { Profile } from "@desmos/generated/types/accounts";

function myParser(input: Any): Account {
  const { typeUrl, value } = input;

  if (typeUrl === "/desmos.profiles.v1beta1.Profile") {
      // version 1
      const baseAccount = Profile.decode(value)?..baseAccount;
      assert(baseAccount);
      return accountFromBaseAccount(baseAccount); // accountFromBaseAccount is not public

     // version 2
    return {
      address: address,
      pubkey: pubkey,
      accountNumber: uint64FromProto(accountNumber).toNumber(), // uint64FromProto is not public
      sequence: uint64FromProto(sequence).toNumber(),
    }
  } else {
     throw new Error("Found unsupported account type"); // is this desired? Do you want a fallback to standard Cosmos SDK accounts?
  }
```

So basically it's unclear to me if it is currently possible to write a good implementation of the new interface.

webmaster128:
> I've actually implemented like the following:

Looks good. I missed that `accountFromAny` is public already. Using `accountFromBaseAccount` directly would be better, but if we don't need to make it public, this is also good.

In this case I consider this feature-complete. Can you add a CHANGELOG entry? Then 🐎🐎🐎

RiccardoM:
> In this case I consider this feature-complete. Can you add a CHANGELOG entry? Then racehorseracehorseracehorse

Added, let me know if it can be improved

packages/stargate/src/accounts.ts
```diff
@@ -36,8 +36,13 @@ function accountFromBaseAccount(input: BaseAccount): Account {
 }
 
 /**
- * Takes an `Any` encoded account from the chain and extracts some common
- * `Account` information from it. This is supposed to support the most relevant
+ * Represents a generic function that takes an `Any` encoded account from the chain
+ * and extracts some common `Account` information from it.
+ */
+export type AccountParser = (any: Any) => Account;
```
webmaster128, line 42: Ah, we probably want to make that public as it is part of a public interface, so it needs to be added to `index.ts` of the package.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- @cosmjs/stargate: Added the ability to specify a custom account parser for `StargateClient`
```
webmaster128, line 11: 👍

packages/stargate/src/stargateclient.ts
```diff
@@ -136,19 +136,27 @@ export interface PrivateStargateClient {
   readonly tmClient: Tendermint34Client | undefined;
 }
 
+export interface StargateClientOptions {
```
webmaster128, line 139: This symbol also needs to be added to `index.ts`

webmaster128 (commented):
Looks good. Just the exports left

packages/stargate/src/accounts.ts
RiccardoM: Done :+1:

packages/stargate/src/stargateclient.ts
RiccardoM: Done :+1:

webmaster128 (approved):

PR #1114, opened 2022-04-07
Use correct data format in vesting aminomessage

### LINKED ISSUE #1074: Add support for vesting messages (MsgCreateVestingAccount)
There seems to be one message only: https://github.com/cosmos/cosmos-sdk/blob/v0.45.1/x/auth/vesting/types/msgs.go

This came up in https://github.com/cosmos/cosmjs/pull/1062 but the author decided to just be selfish and fork away instead of creating a trivial PR to add this type to the codebase.

msteiner96:
Closes #1074


webmaster128 (commented):
okay, looking at https://github.com/cosmos/cosmos-sdk/pull/11019#pullrequestreview-935153832 we have another case of `"not_supported_by_chain"`. Can you change the amino converter implementation to that (see `packages/stargate/src/modules/authz/aminomessages.ts` for an example)? Then the sirect sign mode test remains but the Amino sign mode test needs to be removed.

packages/stargate/src/signingstargateclient.ts
```diff
@@ -61,7 +59,6 @@ export const defaultRegistryTypes: ReadonlyArray<[string, GeneratedType]> = [
   ...govTypes,
   ...stakingTypes,
   ...ibcTypes,
-  ...vestingTypes,
```
webmaster128, line 64: Let's keep those here for sign mode direct support

packages/stargate/src/signingstargateclient.ts
```diff
@@ -102,7 +99,6 @@ function createDefaultTypes(prefix: string): AminoConverters {
     ...createStakingAminoConverters(prefix),
     ...createIbcAminoConverters(),
     ...createFreegrantAminoConverters(),
-    ...createVestingAminoConverters(),
```
webmaster128, line 105: This should also remain here, such that the new `"not_supported_by_chain"` value is found, which gives a better error message

packages/stargate/src/modules/vesting/aminomessages.ts
```diff
@@ -1,56 +1,7 @@
...
-        end_time,
-        delayed,
-      }: AminoMsgCreateVestingAccount["value"]): MsgCreateVestingAccount => ({
-        fromAddress: from_address,
-        toAddress: to_address,
-        amount: [...amount],
-        endTime: Long.fromString(end_time),
-        delayed: delayed,
-      }),
-    },
```
webmaster128, line 55: Bye bye 👋 We'll bring back the code when Cosmos SDK 0.46 is released and deployed.

webmaster128 (approved):

PR #1123, opened 2022-04-13
Update HACKING.md with updated contracts directory

davideitel:
Contracts directory used an old/incorrect path. Fixed to reflect the current code.


webmaster128 (approved):
Thank you

PR #1124, opened 2022-04-13
Add getter to get staked balance

### LINKED ISSUE #1116: Add getter to get staked balance
Then you call getBalance, you don't get staked amounts. It would be good to have a simple method that does this.

An implementation I tested looks like this:

```ts
export function addCoins(lhs: Coin, rhs: Coin): Coin {
  if (lhs.denom !== rhs.denom) throw new Error("Trying to add two coins with different demoms");
  return {
    amount: Decimal.fromAtomics(lhs.amount, 0).plus(Decimal.fromAtomics(rhs.amount, 0)).atomics,
    denom: lhs.denom,
  };
}


  const { delegationResponses, pagination } = await queryClient.staking.delegatorDelegations(
    address,
  );
  assert(pagination?.nextKey?.length === 0); // TODO: support multiple pages
  const staked = delegationResponses.reduce((acc: Coin | null, resp) => {
    // console.log(resp.balance);
    assert(resp.balance);
    if (acc) {
      return addCoins(acc, resp.balance);
    } else {
      return resp.balance;
    }
  }, null);
  return staked;
```

msteiner96:
Resolves #1116

### DIFF
--- a/packages/stargate/src/stargateclient.ts
+++ b/packages/stargate/src/stargateclient.ts
@@ -1,10 +1,13 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import { toHex } from "@cosmjs/encoding";
 import { Uint53 } from "@cosmjs/math";
+import { Decimal } from "@cosmjs/math";
 import { Tendermint34Client, toRfc3339WithNanoseconds } from "@cosmjs/tendermint-rpc";
-import { sleep } from "@cosmjs/utils";
+import { assert, sleep } from "@cosmjs/utils";
 import { MsgData } from "cosmjs-types/cosmos/base/abci/v1beta1/abci";
 import { Coin } from "cosmjs-types/cosmos/base/v1beta1/coin";
+import { QueryDelegatorDelegationsResponse } from "cosmjs-types/cosmos/staking/v1beta1/query";
+import { DelegationResponse } from "cosmjs-types/cosmos/staking/v1beta1/staking";
 
 import { Account, accountFromAny, AccountParser } from "./accounts";
 import {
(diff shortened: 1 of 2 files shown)


packages/stargate/src/stargateclient.ts
```diff
@@ -1,10 +1,13 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import { toHex } from "@cosmjs/encoding";
 import { Uint53 } from "@cosmjs/math";
+import { Decimal } from "@cosmjs/math";
```
webmaster128, line 4:
```suggestion
import { Decimal, Uint53 } from "@cosmjs/math";
```

packages/stargate/src/stargateclient.ts
```diff
@@ -273,6 +276,31 @@ export class StargateClient {
...
+    const allDelegations = [];
+    let startAtKey: Uint8Array | undefined = undefined;
+    do {
+      const { delegationResponses, pagination }: QueryDelegatorDelegationsResponse =
+        await this.forceGetQueryClient().staking.delegatorDelegations(address, startAtKey);
+
+      const loadedDelegations = delegationResponses || [];
+      allDelegations.push(...loadedDelegations);
+      startAtKey = pagination?.nextKey;
+    } while (startAtKey?.length !== 0 && startAtKey !== undefined);
```
webmaster128, line 289: You can swap those to make it a bit easier to understand

```suggestion
    } while (startAtKey !== undefined && startAtKey.length !== 0);
```

This should compile because `startAtKey` is known to be defined after `&&`.

packages/stargate/src/stargateclient.ts
```diff
@@ -273,6 +276,31 @@ export class StargateClient {
...
+
+    const sumValues = allDelegations.reduce(
+      (previousValue: Coin | null, currentValue: DelegationResponse): Coin => {
+        assert(currentValue.balance);
```
webmaster128, line 293:
```suggestion
        // Safe because field is set to non-nullable (https://github.com/cosmos/cosmos-sdk/blob/v0.45.3/proto/cosmos/staking/v1beta1/staking.proto#L295)
        assert(currentValue.balance);
```

packages/stargate/src/stargateclient.ts
```diff
@@ -417,4 +445,15 @@ export class StargateClient {
...
+  /**
+   * Function to sum up coins with type Coin
+   */
+  private addCoins(lhs: Coin, rhs: Coin): Coin {
+    if (lhs.denom !== rhs.denom) throw new Error("Trying to add two coins with different demoms");
+    return {
+      amount: Decimal.fromAtomics(lhs.amount, 0).plus(Decimal.fromAtomics(rhs.amount, 0)).atomics,
+      denom: lhs.denom,
+    };
+  }
```
webmaster128, line 458: Let's move this to `packages/amino/src/coins.ts` and write some tests

webmaster128 (commented):
Nice. Can you add a CHANGELOG entry for `addCoins` and the new method? This this can be shipped with 0.28.4.

webmaster128 (approved):

PR #1147, opened 2022-05-16
Implement Sign Mode Textual

webmaster128:
The textual implementtion is tracked in https://github.com/cosmos/cosmos-sdk/issues/11970. This is the CosmJS implementation including shared test vectors.

Note to SDK developers: heavy rebasing is happening here. Be warned.


packages/proto-signing/src/testdata/coins.json
```diff
@@ -0,0 +1,50 @@
...
+      "uatom": { "denom": "atom", "exponent": 6 }
+    },
+    "0.000001 AA, 0.000002 BB, 0.000003 atom"
+  ],
+  [
+    [
+      { "amount": "8", "denom": "_uxc1" },
+      { "amount": "7", "denom": "_uxc" },
+      { "amount": "6", "denom": "_uxb" },
+      { "amount": "5", "denom": "_uxa" },
```
amaurym, line 32: I got an error here on the SDK: these are Coins, so their denom can't start with `_`. Maybe we can add the `_` inside the denom?

webmaster128: Sure, done

webmaster128:
As far as I can tell, sign mode textual died before it was even finished. At least that was my interpretation of a bunch of in-person conversations with me, Dan from Cosmology, Marko from Binary and Juan from Zondax in May 2024. I don't know what the status of signing in Cosmos is. I think these days nobody works on this stuff anymore.

I will close this PR as it is certainly irrelevant as long as there is no proper spec of the sign mode.

PR #1159, opened 2022-06-02
Feat: option to add fee payer account

arnabghose997:
This PR involves allowing users to set an account as FeePayer, as it is needed for the transaction type `/osmos.authz.v1beta1.MsgExec` to execute. 

Reference Issue: https://github.com/cosmos/cosmjs/issues/1155

### DIFF
--- a/packages/cosmwasm-stargate/src/cosmwasmclient.searchtx.spec.ts
+++ b/packages/cosmwasm-stargate/src/cosmwasmclient.searchtx.spec.ts
@@ -79,7 +79,8 @@ async function sendTokens(
     },
   ];
   const gasLimit = 200000;
-  const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], feeAmount, gasLimit);
+  const feePayer = "";
+  const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], feeAmount, gasLimit, feePayer);
 
   const chainId = await client.getChainId();
   const signDoc = makeSignDoc(txBodyBytes, authInfoBytes, chainId, accountNumber);
--- a/packages/cosmwasm-stargate/src/cosmwasmclient.searchtx.spec.ts
+++ b/packages/cosmwasm-stargate/src/cosmwasmclient.searchtx.spec.ts
@@ -79,8 +79,8 @@ async function sendTokens(
     },
   ];
   const gasLimit = 200000;
-  const feeGranter = "";
-  const feePayer = "";
+  const feeGranter = undefined;
+  const feePayer = undefined;
   const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], feeAmount, gasLimit, feeGranter, feePayer);
 
   const chainId = await client.getChainId();
--- a/packages/cosmwasm-stargate/src/cosmwasmclient.searchtx.spec.ts
+++ b/packages/cosmwasm-stargate/src/cosmwasmclient.searchtx.spec.ts
@@ -79,8 +79,9 @@ async function sendTokens(
     },
   ];
   const gasLimit = 200000;
+  const feeGranter = "";
   const feePayer = "";
-  const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], feeAmount, gasLimit, feePayer);
+  const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], feeAmount, gasLimit, feeGranter, feePayer);
 
   const chainId = await client.getChainId();
   const signDoc = makeSignDoc(txBodyBytes, authInfoBytes, chainId, accountNumber);
--- a/packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
+++ b/packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
@@ -203,8 +203,15 @@ describe("CosmWasmClient", () => {
       };
       const txBodyBytes = registry.encode(txBody);
       const gasLimit = Int53.fromString(fee.gas).toNumber();
+      const feeGranter = "";
       const feePayer = "";
-      const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], fee.amount, gasLimit, feePayer);
+      const authInfoBytes = makeAuthInfoBytes(
+        [{ pubkey, sequence }],
+        fee.amount,
+        gasLimit,
+        feeGranter,
+        feePayer,
+      );
       const signDoc = makeSignDoc(txBodyBytes, authInfoBytes, chainId, accountNumber);
       const { signed, signature } = await wallet.signDirect(alice.address0, signDoc);
       const txRaw = TxRaw.fromPartial({
--- a/packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
+++ b/packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
@@ -203,7 +203,7 @@ describe("CosmWasmClient", () => {
       };
       const txBodyBytes = registry.encode(txBody);
       const gasLimit = Int53.fromString(fee.gas).toNumber();
-      const feePayer = ""
+      const feePayer = "";
       const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], fee.amount, gasLimit, feePayer);
       const signDoc = makeSignDoc(txBodyBytes, authInfoBytes, chainId, accountNumber);
       const { signed, signature } = await wallet.signDirect(alice.address0, signDoc);
--- a/packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
+++ b/packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
@@ -203,8 +203,8 @@ describe("CosmWasmClient", () => {
       };
       const txBodyBytes = registry.encode(txBody);
       const gasLimit = Int53.fromString(fee.gas).toNumber();
-      const feeGranter = "";
-      const feePayer = "";
+      const feeGranter = undefined;
+      const feePayer = undefined;
       const authInfoBytes = makeAuthInfoBytes(
         [{ pubkey, sequence }],
         fee.amount,
--- a/packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
+++ b/packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
@@ -203,7 +203,8 @@ describe("CosmWasmClient", () => {
       };
       const txBodyBytes = registry.encode(txBody);
       const gasLimit = Int53.fromString(fee.gas).toNumber();
-      const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], fee.amount, gasLimit);
+      const feePayer = ""
+      const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], fee.amount, gasLimit, feePayer);
       const signDoc = makeSignDoc(txBodyBytes, authInfoBytes, chainId, accountNumber);
       const { signed, signature } = await wallet.signDirect(alice.address0, signDoc);
       const txRaw = TxRaw.fromPartial({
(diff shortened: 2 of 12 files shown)


packages/proto-signing/src/signing.ts
```diff
@@ -34,13 +34,15 @@ export function makeAuthInfoBytes(
...
   gasLimit: number,
+  feePayer: string,
   signMode = SignMode.SIGN_MODE_DIRECT,
 ): Uint8Array {
   const authInfo = {
     signerInfos: makeSignerInfos(signers, signMode),
     fee: {
       amount: [...feeAmount],
       gasLimit: Long.fromNumber(gasLimit),
+      granter: feePayer,
```
webmaster128, line 45: Granter and payer are different things. I think we need both fields here.

The type definition is:

```ts
/**
 * Fee includes the amount of coins paid in fees and the maximum
 * gas to be used by the transaction. The ratio yields an effective "gasprice",
 * which must be above some miminum to be accepted into the mempool.
 */
export interface Fee {
  /** amount is the amount of coins to be paid as a fee */
  amount: Coin[];
  /**
   * gas_limit is the maximum gas that can be used in transaction processing
   * before an out of gas error occurs
   */
  gasLimit: Long;
  /**
   * if unset, the first signer is responsible for paying the fees. If set, the specified account must pay the fees.
   * the payer must be a tx signer (and thus have signed this field in AuthInfo).
   * setting this field does *not* change the ordering of required signers for the transaction.
   */
  payer: string;
  /**
   * if set, the fee payer (either the first signer or the value of the payer field) requests that a fee grant be used
   * to pay fees instead of the fee payer's own balance. If an appropriate fee grant does not exist or the chain does
   * not support fee grants, this will fail
   */
  granter: string;
}
```

packages/proto-signing/src/signing.ts
```diff
@@ -34,13 +34,15 @@ export function makeAuthInfoBytes(
   signers: ReadonlyArray<{ readonly pubkey: Any; readonly sequence: number }>,
   feeAmount: readonly Coin[],
   gasLimit: number,
+  feePayer: string,
```
webmaster128, line 37: Let's give this some proper optional type: `string | undefined`

webmaster128 (commented):
Nice start. I think there is some confision with payer vs. granter. Please consult the Cosmos SDK team if you have questions what the difference is. I never used this tech.

packages/proto-signing/src/signing.ts
arnabghose997: While performing `x/authz` exec transaction, I observed it was only the `granter` field that was being populated, and not the `payer` field. Hence, I didn't include the `payer` field.

Should we also `payer` field in `StdFee` interface, along with `granter`?

packages/proto-signing/src/signing.ts
arnabghose997: Sure, I will incorporate this change.

packages/proto-signing/src/signing.ts
webmaster128: Yeah, let's add both. This avoids confusion and users can choose what they want/need to set.

arnabghose997:
`payer` and `granter` fields are added as part of `StdFee` interface.

packages/cosmwasm-stargate/src/testutils.spec.ts
```diff
@@ -224,13 +224,17 @@ export class ModifyingDirectSecp256k1HdWallet extends DirectSecp256k1HdWallet {
     }));
     const modifiedFeeAmount = coins(3000, "ucosm");
     const modifiedGasLimit = 333333;
+    const modifiedFeeGranter = "";
+    const modifiedFeePayer = "";
```
webmaster128, line 228:
```suggestion
    const modifiedFeeGranter = undefined;
    const modifiedFeePayer = undefined;
```

packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
```diff
@@ -596,7 +600,15 @@ export class SigningCosmWasmClient extends CosmWasmClient {
     };
     const txBodyBytes = this.registry.encode(txBody);
     const gasLimit = Int53.fromString(fee.gas).toNumber();
-    const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], fee.amount, gasLimit);
+    const feePayer = fee.payer == undefined ? "" : fee.payer;
+    const feeGranter = fee.granter == undefined ? "" : fee.granter;
```
webmaster128, line 604: Those lines are not needed anymore as undefined is now handled internally

packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
```diff
@@ -559,10 +559,14 @@ export class SigningCosmWasmClient extends CosmWasmClient {
     const signedTxBodyBytes = this.registry.encode(signedTxBody);
     const signedGasLimit = Int53.fromString(signed.fee.gas).toNumber();
     const signedSequence = Int53.fromString(signed.sequence).toNumber();
+    const signedFeeGranter = signed.fee.granter == undefined ? "" : signed.fee.granter;
+    const signedFeePayer = signed.fee.payer == undefined ? "" : signed.fee.payer;
```
webmaster128, line 563: Those lines are not needed anymore as undefined is now handled internally

packages/cosmwasm-stargate/src/cosmwasmclient.spec.ts
```diff
@@ -203,7 +203,15 @@ describe("CosmWasmClient", () => {
       };
       const txBodyBytes = registry.encode(txBody);
       const gasLimit = Int53.fromString(fee.gas).toNumber();
-      const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], fee.amount, gasLimit);
+      const feeGranter = "";
+      const feePayer = "";
```
webmaster128, line 207:
```suggestion
      const feeGranter = undefined;
      const feePayer = undefined;
```

packages/cosmwasm-stargate/src/cosmwasmclient.searchtx.spec.ts
```diff
@@ -79,7 +79,9 @@ async function sendTokens(
     },
   ];
   const gasLimit = 200000;
-  const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], feeAmount, gasLimit);
+  const feeGranter = "";
+  const feePayer = "";
```
webmaster128, line 83:
```suggestion
  const feeGranter = undefined;
  const feePayer = undefined;
```

packages/stargate/src/signingstargateclient.ts
```diff
@@ -362,10 +362,14 @@ export class SigningStargateClient extends StargateClient {
     const signedTxBodyBytes = this.registry.encode(signedTxBodyEncodeObject);
     const signedGasLimit = Int53.fromString(signed.fee.gas).toNumber();
     const signedSequence = Int53.fromString(signed.sequence).toNumber();
+    const signedFeeGranter = signed.fee.granter == undefined ? "" : signed.fee.granter;
+    const signedFeePayer = signed.fee.payer == undefined ? "" : signed.fee.payer;
```
webmaster128, line 366: Those lines are not needed anymore as undefined is now handled internally

packages/stargate/src/signingstargateclient.ts
```diff
@@ -399,7 +403,15 @@ export class SigningStargateClient extends StargateClient {
     };
     const txBodyBytes = this.registry.encode(txBodyEncodeObject);
     const gasLimit = Int53.fromString(fee.gas).toNumber();
-    const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], fee.amount, gasLimit);
+    const feeGranter = fee.granter == undefined ? "" : fee.granter;
+    const feePayer = fee.payer == undefined ? "" : fee.payer;
```
webmaster128, line 407: Those lines are not needed anymore as undefined is now handled internally

packages/stargate/src/stargateclient.searchtx.spec.ts
```diff
@@ -74,7 +74,9 @@ async function sendTokens(
     },
   ];
   const gasLimit = 200000;
-  const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], feeAmount, gasLimit);
+  const feeGranter = "";
+  const feePayer = "";
```
webmaster128, line 78:
```suggestion
  const feeGranter = undefined;
  const feePayer = undefined;
```

packages/stargate/src/stargateclient.spec.ts
```diff
@@ -360,7 +360,15 @@ describe("StargateClient", () => {
       const { accountNumber, sequence } = (await client.getSequence(address))!;
       const feeAmount = coins(2000, "ucosm");
       const gasLimit = 200000;
-      const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], feeAmount, gasLimit);
+      const feeGranter = "";
+      const feePayer = "";
```
webmaster128, line 364:
```suggestion
      const feeGranter = undefined;
      const feePayer = undefined;
```

packages/stargate/src/stargateclient.spec.ts
```diff
@@ -417,7 +425,16 @@ describe("StargateClient", () => {
       const { accountNumber, sequence } = (await client.getSequence(address))!;
       const feeAmount = coins(2000, "ucosm");
       const gasLimit = 200000;
-      const authInfoBytes = makeAuthInfoBytes([{ pubkey, sequence }], feeAmount, gasLimit, sequence);
+      const feeGranter = "";
+      const feePayer = "";
```
webmaster128, line 429:
```suggestion
      const feeGranter = undefined;
      const feePayer = undefined;
```

packages/stargate/src/testutils.spec.ts
```diff
@@ -234,13 +234,17 @@ export class ModifyingDirectSecp256k1HdWallet extends DirectSecp256k1HdWallet {
     }));
     const modifiedFeeAmount = coins(3000, "ucosm");
     const modifiedGasLimit = 333333;
+    const modifiedFeeGranter = "";
+    const modifiedFeePayer = "";
```
webmaster128, line 238:
```suggestion
    const modifiedFeeGranter = undefined;
    const modifiedFeePayer = undefined;
```

webmaster128 (commented):
Looks good overall, thanks.

Do you have an idea how to test the new code?

arnabghose997:
The sequence of transactions for a complete authz+feegrant are as follows:

1) `/cosmos.authz.v1beta1.MsgGrant`
2) `/cosmos.feegrant.v1beta1.MsgGrantAllowance`
3) `/cosmos.authz.v1beta1.MsgExec`

The test should execute the above messages sequentially. We can consider `packages/stargate/src/modules/authz/queries.spec.ts` as a reference.

Before proceeding with writing this test, I have a query. As mentioned in `HACKING.md`, we need to run few scripts and set flags to perform complete test suite run. In the current version of `HACKING.md` present on `main` branch, there is mention of running launchpad blockchain, although its corresponding scripts aren't present in the `main` branch. So, I wanted to know about the scipts we need to run and the flags that needs to set instead of launchpad?

dckc:
Anything I can do to help this along?

webmaster128:
> Anything I can do to help this along?

Thank you @dckc. I am hesitant to merge this without the feature being tested. I don't want this feature to fail in the user's environment. Also having a test case serves as an always up-to-date example how to use the feature. Writing tests may detect flaws in the API. If you could add tests on top of that PR, I think we can get this out.

Please note that on latest main we have Cosmos SDK 0.44 and 0.46 (simapp44/simapp46) test chains. simapp42 was removed. With those chains it should be possible to test the whole flow.

webmaster128:
I'm finishing this work in #1262. The test is in https://github.com/cosmos/cosmjs/pull/1262/commits/c959343cacd868ed60891cbd7c92a1a2f24448ba. Turns out no `/cosmos.authz.v1beta1.MsgGrant` and `/cosmos.authz.v1beta1.MsgExec` is needed to test that as you can use feegrant for a simple bank send.

PR #1160, opened 2022-06-06
EthSecp

zakarialounes:
Please also see: https://github.com/confio/cosmjs-types/pull/30

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to
   changes all 64 bit int fields from type `long` to `bigint`. As part of the
   upgrade, the types do not require the `long` and `protobufjs` anymore.
   ([#1484])
+- all: `gasWanted`/`gasUsed` fields were changed from type `number` to `bigint`
+  to supported cases where users put very high gas values in there ([#1465]).
 - Drop support for Node.js 14 and add support for Node.js 20. ([#1421])
 - @cosmjs/tendermint-rpc: Remove `Adaptor` abstractions which are not needed
   anymore by haing a dedicated client for each backend.
@@ -34,6 +36,7 @@ and this project adheres to
   `.getCometClient`/`.forceGetCometClient`.
 
 [#1421]: https://github.com/cosmos/cosmjs/issues/1421
+[#1465]: https://github.com/cosmos/cosmjs/issues/1465
 [#1484]: https://github.com/cosmos/cosmjs/pull/1484
 
 ### Deprecated
--- a/packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
+++ b/packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
@@ -78,8 +78,8 @@ export interface UploadResult {
   /** Transaction hash (might be used as transaction ID). Guaranteed to be non-empty upper-case hex */
   readonly transactionHash: string;
   readonly events: readonly Event[];
-  readonly gasWanted: number;
-  readonly gasUsed: number;
+  readonly gasWanted: bigint;
+  readonly gasUsed: bigint;
 }
 
 /**
@@ -112,8 +112,8 @@ export interface InstantiateResult {
   /** Transaction hash (might be used as transaction ID). Guaranteed to be non-empty upper-case hex */
   readonly transactionHash: string;
   readonly events: readonly Event[];
-  readonly gasWanted: number;
-  readonly gasUsed: number;
+  readonly gasWanted: bigint;
+  readonly gasUsed: bigint;
 }
 
 /**
@@ -126,8 +126,8 @@ export interface ChangeAdminResult {
   /** Transaction hash (might be used as transaction ID). Guaranteed to be non-empty upper-case hex */
   readonly transactionHash: string;
   readonly events: readonly Event[];
-  readonly gasWanted: number;
-  readonly gasUsed: number;
+  readonly gasWanted: bigint;
+  readonly gasUsed: bigint;
 }
 
 export interface MigrateResult {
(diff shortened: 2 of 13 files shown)


keupsonite:
cc: @fedekunze

liangping:
Hope we can merge it ASAP.

webmaster128:
Is `tendermint/PubKeyEthSecp256k1` a type that exists in Tendermint? At least it uses a Tendermint namespace. But I don't see it in the codebase. I would like to avoid adding code for forked Tendermint/Cosmos SDK tech.

liangping:
It's introduced By EVMOS, we need it if we want to integrate evmos stuff into cosmjs.

keupsonite:
> Is `tendermint/PubKeyEthSecp256k1` a type that exists in Tendermint? At least it uses a Tendermint namespace. But I don't see it in the codebase. I would like to avoid adding code for forked Tendermint/Cosmos SDK tech.

@webmaster128 So this PR will not be accepted?

StregoFren:
I think the underlying problem is that comsjs currently can't be extended in Application Land to support this. In some way I think this is very special for the type of an Account. If it would be just messages, we could add them to the registry.

I think short term, it might make sense to support evmos accounts in cosmjs. If there are too many of those incidents, it might make more sense to allow users hooking into the account parsing.

jolube:
@StregoFren @webmaster128 In fact it is introduced by Ethermint: https://github.com/tharsis/ethermint

The open source library is used and maintained by the Evmos core team, in addition to being used in for a number of other chains; [Cronos](https://github.com/crypto-org-chain/cronos), [Kava](https://github.com/Kava-Labs/kava), [GenesisL1](https://github.com/alpha-omega-labs/genesisd), etc.

Supporting Ethermint types in cosmjs will ensure support for a growing number of chains in the Cosmos ecosystem that rely on Ethermint's EVM.

packages/stargate/src/signingstargateclient.ts
```diff
@@ -346,7 +351,11 @@ export class SigningStargateClient extends StargateClient {
     if (!accountFromSigner) {
       throw new Error("Failed to retrieve account from signer");
     }
-    const pubkey = encodePubkey(encodeSecp256k1Pubkey(accountFromSigner.pubkey));
+    const pubkey = encodePubkey(
+      chainId.startsWith("evmos_9001-")
```
jiujiteiro, line 355: Is there a more general approach over the evmos chain id? Other chains may be using the EthSecp keys from an ethermint integration. It would be nice if they can also use the cosm library in this case.

keupsonite: When this is the case then we can make the code more complex so that it is manageable. No need to make it more complex for the moment as long as it is not useful.

jiujiteiro: we already have other chains using the evm module from ethermint other than evmos. A few were mentioned here already (cronos, kava etc), and I am working on one myself. Given that I would love to see ethermint types supported in cosmjs in a way that these other chains could leverage.

keupsonite: Oh I didn't know that. So I'm going to rework this so it's scalable and get back to you when it's right. Thanks for the heads up.

webmaster128:
I have no issue adding support for evmos types where this is possible in a clean way. But looking at the diffs people send for evmos support, it feels like big time hackery in the deepest core of CosmJS. If evmos changes those things, it is no surprise Cosmos SDK client libraries don't work anymore.

Well certainly not add chain ID checks to change internal logic. There should also be a different Amino prefix for the different pubkey type. `AccountData` does not support the ethermint pubkey type, but this could be changed.

Adding ethermint types in [cosmjs-types](https://github.com/confio/cosmjs-types) seems useful. But not in `src/cosmos/crypto/ethsecp256k1/keys.ts` as there is a separate ethermint protobuf package already.

dimiandre:
So as dapp developers, how can we interact with evmos? 

is there any hacky way to do that dAPP side instead of cosmjs side?

keupsonite:
As soon as I have some time I will update the code to import the ethermint types directly into cosmjs (without having to go through the chain-id, but via the coinType). Before that I'll update the code with my last changes (chain-id updated) and I'll publish the package on npm so that it's possible to take advantage of it now (since there is absolutely no chance that's the Amino prefix will be changed imho). I let you know.

packages/amino/src/encoding.ts
```diff
@@ -45,13 +61,15 @@ export function encodeEd25519Pubkey(pubkey: Uint8Array): Ed25519Pubkey {
 // Prefixes listed here: https://github.com/tendermint/tendermint/blob/d419fffe18531317c28c29a292ad7d253f6cafdf/docs/spec/blockchain/encoding.md#public-key-cryptography
 // Last bytes is varint-encoded length prefix
 const pubkeyAminoPrefixSecp256k1 = fromHex("eb5ae987" + "21" /* fixed length */);
+const pubkeyAminoPrefixEthSecp256k1 = fromHex("eb5ae987" + "21" /* fixed length */);
```
webmaster128, line 64: Did you test this prefix? Is this the one used in address derivation?

This is wrong in two ways:
1. eb5ae987 refers to tendermint/PubKeySecp256k1
2. 0x21 means that data of length 33 bytes is following

webmaster128:
Thank you for this PR. It was a very helpful contribution to https://github.com/cosmos/cosmjs/issues/1351. For now it is undecided if Ethermint is in scope of this project.

packages/amino/src/encoding.ts
keupsonite: EthSecp256 use the same prefix and length (please see https://github.com/cosmos/cosmjs/pull/1160#issuecomment-1264779586). 
Works and is available for trial via https://ezstaking.tools/.

webmaster128: Thank you, good to know. Then this is broken in Ethermint. Possibly we could work around that by removing the reverse operation bytes -> Pubkey. Let's see ...

vincentysc:
Hey~ Im interested in the ethermint support too. Im willing to contribute and push this forward. So the blockers is how to ideal with the same prefix of `secp256k1` and `ethsecp256k1`. And the chain-id issue can be solved by using the `coin_type=60`. Those are the blockers. Am I right?

PR #1166, opened 2022-06-08
use protobufjs as a devDep only

mrnerdhair:
Currently, `protobufjs` is used only for type information and running unit tests. Using `import type` allows it to be demoted from `dependencies` to `devDependencies`, reducing this package's footprint significantly (and sidestepping various complaints generated by naive interpretations of the output of `npm audit` to boot).


packages/proto-signing/src/registry.ts
```diff
@@ -4,7 +4,7 @@ import { Coin } from "cosmjs-types/cosmos/base/v1beta1/coin";
 import { TxBody } from "cosmjs-types/cosmos/tx/v1beta1/tx";
 import { Any } from "cosmjs-types/google/protobuf/any";
 import Long from "long";
-import protobuf from "protobufjs";
+import type protobuf from "protobufjs";
```
webmaster128, line 7: Nice one, not not know this 💎

webmaster128 (approved):
Nice change, thank you!

But please note that the runtime dependency still exists in ts-proto generated code (see https://github.com/confio/cosmjs-types/blob/v0.5.0/package.json).

PR #1178, opened 2022-06-15
fix: decoding validator update from slashing

### LINKED ISSUE #1177: Error while decoding validator updates due to slashing
As seen in the [results for Cosmos block 10883046](https://rpc.cosmos.network/block_results?height=10883046), when a validator is slashed the `validator_updates` entry does not have a `power` attribute. The decoder for this should be changed to make this property optional, as it is currently required and cannot be parsed.

Full error:
```
Error: Value must not be undefined
    at assertSet (.../node_modules/@cosmjs/tendermint-rpc/build/tendermint34/encodings.js:13:15)
    at assertNotEmpty (.../node_modules/@cosmjs/tendermint-rpc/build/tendermint34/encodings.js:107:5)
    at decodeValidatorUpdate (.../node_modules/@cosmjs/tendermint-rpc/build/tendermint34/adaptor/responses.js:122:79)
    at Array.map (<anonymous>)
    at decodeBlockResults (.../node_modules/@cosmjs/tendermint-rpc/build/tendermint34/adaptor/responses.js:130:58)
    at decodeBlockResults (.../node_modules/@cosmjs/tendermint-rpc/build/tendermint34/adaptor/responses.js:384:16)
    at Tendermint34Client.doCall (.../node_modules/@cosmjs/tendermint-rpc/build/tendermint34/tendermint34client.js:306:16)
```

Related to #1142.

blorgon1:
Closes #1177.

### DIFF
--- a/packages/tendermint-rpc/src/tendermint34/adaptor/responses.ts
+++ b/packages/tendermint-rpc/src/tendermint34/adaptor/responses.ts
@@ -257,13 +257,14 @@ function decodeConsensusParams(data: RpcConsensusParams): responses.ConsensusPar
 // for block results
 interface RpcValidatorUpdate {
   readonly pub_key: RpcPubkey;
+  // When omitted, this means zero (see https://github.com/cosmos/cosmjs/issues/1177#issuecomment-1160115080)
   readonly power?: string;
 }
 
 export function decodeValidatorUpdate(data: RpcValidatorUpdate): responses.ValidatorUpdate {
   return {
     pubkey: decodePubkey(assertObject(data.pub_key)),
-    votingPower: data.power ? Integer.parse(data.power) : undefined,
+    votingPower: Integer.parse(data.power ?? 0),
   };
 }
 
--- a/packages/tendermint-rpc/src/tendermint34/adaptor/responses.ts
+++ b/packages/tendermint-rpc/src/tendermint34/adaptor/responses.ts
@@ -257,13 +257,13 @@ function decodeConsensusParams(data: RpcConsensusParams): responses.ConsensusPar
 // for block results
 interface RpcValidatorUpdate {
   readonly pub_key: RpcPubkey;
-  readonly power: string;
+  readonly power?: string;
 }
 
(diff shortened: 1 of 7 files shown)


packages/tendermint-rpc/src/tendermint34/responses.ts
```diff
@@ -356,7 +356,7 @@ export interface Validator {
 
 export interface ValidatorUpdate {
   readonly pubkey: ValidatorPubkey;
-  readonly votingPower: number;
+  readonly votingPower?: number;
```
webmaster128, line 359: Let's keep this non-optional

packages/tendermint-rpc/src/tendermint35/adaptor/responses.ts
```diff
@@ -259,13 +259,13 @@ function decodeConsensusParams(data: RpcConsensusParams): responses.ConsensusPar
 // for block results
 interface RpcValidatorUpdate {
   readonly pub_key: RpcPubkey;
-  readonly power: string;
+  readonly power?: string;
```
webmaster128, line 262:
```suggestion
  // When omitted, this means zero (see https://github.com/cosmos/cosmjs/issues/1177#issuecomment-1160115080)
  readonly power?: string;
```

packages/tendermint-rpc/src/tendermint34/adaptor/responses.ts
```diff
@@ -257,13 +257,13 @@ function decodeConsensusParams(data: RpcConsensusParams): responses.ConsensusPar
 // for block results
 interface RpcValidatorUpdate {
   readonly pub_key: RpcPubkey;
-  readonly power: string;
+  readonly power?: string;
```
webmaster128, line 260:
```suggestion
  // When omitted, this means zero (see https://github.com/cosmos/cosmjs/issues/1177#issuecomment-1160115080)
  readonly power?: string;
```

webmaster128 (commented):
Thank you for the fix!

Can you use something like `votingPower: Integer.parse(data.power ?? 0),` in the implementation to convert back the omitted field to zero?

webmaster128 (approved):
Thank you!

PR #1188, opened 2022-06-22
Fix `Decimal` can be constructed with a negative atomics

tien:


webmaster128 (commented):
Nice catch! The atomics string should checked more strictly.

Can you use

```ts
    if (!atomics.match(/^[0-9]+$/)) {
      throw new Error("Invalid string format. Only non-negative integers in decimal representation suppored.");
    }
```

instead?

tien:
@webmaster128 sure thing, I have updated the PR accordingly.

webmaster128 (approved):
Thank you!

Will merge as part of #1195

webmaster128:

PR #1198, opened 2022-06-28
Fix allow optional attributes on decodeAttributes

holyxiaoxin:
Empty events emitted will result in missing attributes field in events object, which is a legit scenario.

```json
{
   "type": "cosmos.module.EmittedEvent"
}
```
Which causes decodeAttributes assertArray throw an exception `UnhandledPromiseRejectionWarning: Error: Value must not be undefined`. I'm proposing to make this field optional.


webmaster128:
Thank you for bringing this up. Could you please have a look in https://github.com/cosmos/cosmjs/pull/1199 for an alternative implementation?

webmaster128:

PR #1199, opened 2022-06-28
Fix decoding of events with no attributes

webmaster128:
Closes #1198


holyxiaoxin:
Could you also help publish this next tag? thanks.

webmaster128:
> Could you also help publish this next tag? thanks.

What about backporting this to 0.28.x instead? Or do you need other 0.29 features?

holyxiaoxin:
> > Could you also help publish this next tag? thanks.
> 
> What about backporting this to 0.28.x instead? Or do you need other 0.29 features?

backporting to 0.28.x is good! thanks

webmaster128:
Published as part of 0.28.10

PR #1211, opened 2022-07-07
timeout check uses microseconds

donkey-donkey:
the time delay for requests was not working.
the time comparison is done in micro seconds so the amount added to the time needs to be in microseconds. 
I added in a * 1000 and it works now.


webmaster128:
In https://github.com/cosmos/cosmjs/pull/1218 you see a version of your patch that makes code a bit more future-proof. It will be shipped as a 0.28.x patch release this week.

webmaster128:
Done in #1218

PR #1218, opened 2022-07-12
Fix cooldown value

webmaster128:
Closes #1211

Thanks a lot to @donkey-donkey for bringing this up and providing a patch!


webmaster128:
Shipped as part of 0.28.11

PR #1219, opened 2022-07-14
Minor broken link when reading on NPM's site

mikedotexe:
When visiting the site:
https://www.npmjs.com/package/@cosmjs/cli

Near the bottom there's a broken link, perhaps because it used a relative path. Just set that to the long-form URL so it won't 404 on anyone.

Then also a little typo.


webmaster128 (approved):
Thank you

PR #1241, opened 2022-08-16
fix: return tx response data without ignoring it

youngjoon-lee:
The `DeliverTxRepsonse` interface already has a [data](https://github.com/cosmos/cosmjs/blob/e514144e48ff74c8741f08c86c876bc54ffdf361/packages/stargate/src/stargateclient.ts#L102) field so that it can contain tx result data. But, it doesn't get the `data` and set it into the `DeliverTxResponse` when querying tx results internally. 

So, I fixed it to read `data` from tx results properly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,11 +44,13 @@ and this project adheres to
 - @cosmjs/tendermint-rpc: Fix `key` and `value` type in `RpcAbciQueryResponse`
   to also include the `null` option.
 - @cosmjs/tendermint-rpc: Fix decoding events without attributes ([#1198]).
+- @cosmjs/stargate: Return tx response data without ignoring it ([#1241]).
 
 [#1170]: https://github.com/cosmos/cosmjs/issues/1170
 [#1177]: https://github.com/cosmos/cosmjs/issues/1177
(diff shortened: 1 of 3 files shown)


youngjoon-lee:
Hi @webmaster128 ,
It seems that CI jobs keep failing because of timeout from downloading docker images or initializing tendermint node. Since I don't have a permission to re-run jobs, I've pushed small commits to re-trigger CI jobs. Could you re-run jobs or suggest any better way?

webmaster128:
Sorry for keeping you wait so long. I just wanted to implement this change in a new PR but ran into this upcoming Cosmos SDK change, which opens up a larger API design question: https://github.com/cosmos/cosmjs/issues/1305#issuecomment-1291116184. Whoever finds this PR, feel free to follow #1305 for updates on this matter.

TheMarstonConnell:
@webmaster128 We're looking to get the response from a Txn using cosmjs on our chain running SDK 45, does this PR being closed and unmerged mean there is no way to do so?

webmaster128:
Looking at https://github.com/cosmos/cosmjs/issues/1305 it seems like I gave up on adding new features for 0.45 chains only. I guess the data is not available in the high level client.

PR #1242, opened 2022-08-16
fix(account): Allow getAccount to work with account 0.

SylvestreG:
Account 0 can exist on chain. This will allow the few owners of account 0 to use cosmjs :)


webmaster128 (commented):
`account` is of type `Any | null`. How does this change fix an issue with Account number 0?

SylvestreG:
Sorry I did not dig enough, I was thinking that the account was potentially a number. I've got some issue performing wasm request on cosmjs using a wallet with the account number 0. 

Sorry for the noise, I close the PR and will dig further.

webmaster128:
Hmm, we has and solved some issues with account number 0 before. Not impossible there is something left. Could you provide a detailed description what you do and which error you get (in a new ticket)?

PR #1253, opened 2022-08-30
Implement determinstic contract addresses

webmaster128:
Implementation for ~https://github.com/CosmWasm/wasmd/issues/942~ https://github.com/CosmWasm/wasmd/pull/1014.

The output is available here: https://gist.github.com/webmaster128/e4d401d414bd0e7e6f70482f11877fbe


packages/cosmwasm-stargate/src/instantiate2.ts
```diff
@@ -0,0 +1,69 @@
...
+ * Predictable address generation for the MsgInstantiateContract2
+ * introduced with wasmd 0.29.
+ */
+export function instantiate2Address(
+  checksum: Uint8Array,
+  creator: string,
+  salt: Uint8Array,
+  prefix: string,
+): string {
+  return _instantiate2AddressIntermediate(checksum, creator, salt, null, prefix).address;
```
loin3, line 68: Why msg is always set to null when instantiate2 while Wasmd v0.31.x include msgs when creating address?

webmaster128: See https://medium.com/cosmwasm/dev-note-3-limitations-of-instantiate2-and-how-to-deal-with-them-a3f946874230

webmaster128: I added https://github.com/cosmos/cosmjs/pull/1398 for future readers

loin3: I got it. 
However, according to the link you provided, it seems that fix_msg should be null, not that instantiate msg should be null and current code is making instantiate msg to null.

loin3: To be honest, I am not familiar with wasmd, so I'm not sure how to use instantiate2. 
If you have a chance, could you make a test code that compares the address obtained with instantiate2Address and the address of the contract instantiated with instantiate2 in the script/wasmd chain?

webmaster128: Yeah, it's still early for instantiate2 and not much resources are out there. Feel free to subscribe to https://github.com/cosmos/cosmjs/issues/1399 to see
- Pre-calculation
- Actual instantiation
in CosmJS. The tests for this ticket will make a few things clearer I guess.

webmaster128: Oh, nice! I **strongly** advise for removing the msg input field there. Nobody should need to use it.

PR #1260, opened 2022-09-07
CReate initial Module type to collect symbols by module

webmaster128:


webmaster128:
Potentially a good idea but currently not planned. This requires a fresh start. In the lights of https://github.com/cosmos/cosmjs/issues/1649#issuecomment-3636040875 I close this.

PR #1278, opened 2022-10-03
Add query*Block functions that report height

kriskowal:
This change introduces parallel implementations of queryVerified and queryUnverified that return an object containing both the value and the height it was found at.  This would allow us at Agoric to avoid a round trip through abciInfo to discover the current height before calling queryVerified.

(Please feel free to treat this as an issue-with-code.)

### DIFF
--- a/packages/stargate/src/queryclient/queryclient.ts
+++ b/packages/stargate/src/queryclient/queryclient.ts
@@ -2,7 +2,7 @@
 import { iavlSpec, ics23, tendermintSpec, verifyExistence, verifyNonExistence } from "@confio/ics23";
 import { toAscii, toHex } from "@cosmjs/encoding";
 import { firstEvent } from "@cosmjs/stream";
-import { tendermint34, Tendermint34Client } from "@cosmjs/tendermint-rpc";
+import { tendermint34, Tendermint34Client, AbciQueryResponse } from "@cosmjs/tendermint-rpc";
 import { arrayContentEquals, assert, assertDefined, isNonNullObject, sleep } from "@cosmjs/utils";
 import { ProofOps } from "cosmjs-types/tendermint/crypto/proof";
 import { Stream } from "xstream";
@@ -23,6 +23,12 @@ export interface ProvenQuery {
   readonly key: Uint8Array;
   readonly value: Uint8Array;
   readonly proof: ProofOps;
+  readonly height: number;
+}
+
+export interface UnprovenQuery {
+  readonly key: Uint8Array;
+  readonly value: Uint8Array;
   readonly height: number;
 }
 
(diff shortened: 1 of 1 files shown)


kriskowal:
cc @webmaster128. Please let us know if this is a direction you’d entertain. Normally I’d open an issue first, but I think code speaks in this case.

packages/stargate/src/queryclient/queryclient.ts
```diff
@@ -575,6 +586,15 @@ export class QueryClient {
...
   ): Promise<Uint8Array> {
+    const { value } = await this.queryUnverifiedBlock(path, request, desiredHeight);
+    return value;
+  }
+
+  public async queryUnverifiedBlock(
+    path: string,
+    request: Uint8Array,
+    desiredHeight?: number,
+  ): Promise<AbciQueryResponse> {
```
webmaster128, line 597: Which values

webmaster128:
Thank you. This makes sense overall. Just a few thoughts:
1. It would be good to have a dedicated return type for both functions to not rely on tendermint-rpc details. Which fields of AbciQueryResponse do you need?
2. I would go one step further and let `queryVerified`/`queryUnverified` return the data you need instead of maintaining 2 more additional functions.
3. Whats the reasoning behind the "Block" term? I don't understand it yet.

For 2. we need a migration path but we can break the API in 0.30. How urgent is the diff for you?

kriskowal:
I’ve not reasoned hard about the `Block` suffix; please consider it a placeholder. `Envelope`,`WithHeight`, or any suitable name for a data + height struct should suffice. There’s no ideal name, of course, and you’re right that we could present the fatter result object under the same function name, except for the migration story.

If the notion is that replace the existing methods in the fullness of time, we could also greatly simplify the migration story if we just changed the name of the method to something equally sensible, like `queryProven` and `queryUnproven`. Then it’s straight-forward to deprecate `queryVerified` with the message “Please switch to using `queryProven` and extract the `data` field of the result. For queries with an unspecified height, the result will also include `height`”. Then in the next breaking-change version-bump release, remove the old methods.

webmaster128:
Step 1 is here: https://github.com/cosmos/cosmjs/pull/1328. @kriskowal could you check if the new method serves your needs?

turadg:
@webmaster128 that works for our unverified case. Are you planning changes to support `queryVerified` returning [the height it gets from queryRawProof](https://github.com/cosmos/cosmjs/blob/74346b2ce6/packages/stargate/src/queryclient/queryclient.ts#L509)?

webmaster128:
Yeah, I do. Unfortunately I'm currently busy with CosmWasm.

This is more an ABI beauty challenge than anything else. But naming matters. I see step 2 as creating a something along the lines of `queryStorageVerified(...)` to make a clear differentiation between "query ABCI" and "query storage".

webmaster128:
See https://github.com/cosmos/cosmjs/pull/1337 for the second part of this. I called the new function `queryStoreVerified` as it can only query a store, not any ABCI query. Also I added a note to the limitations: It verifies the responses based in unverified block headers. But changing that would be a massive project.

Reviews welcome.

webmaster128:
Closing in favour of #1328 and #1337. Sorry for letting you wait so long. But now I'm quite happy with the API that explaisn much better what is actually queried. Thank you for your contribution and reviews!

webmaster128:
Released as part of 0.29.5

PR #1281, opened 2022-10-07
fix: fixed the cosmwasm msg types

dadamu:
The typescript file of messages using new macro `cw_serde` then generated by codegen are no longer fix `Record<string, unknown>`. The msg types should be `JsonObject`, instead.

The reason is as follows:
1. schema json changed to have value `"additionalProperties": false` by this [change](https://github.com/CosmWasm/cw-nfts/commit/6dfb0c6f020b417620708a615f19828929f01cdb#diff-4244bdf6e15adf9c0a9e27dd191357efb818a9d2a25eff4d10a9b1fa1aeccd46R24) in `cw_nfts`
2. codegen ts structure loses `[k: string]: unknown`, it can be found at [cw721-base-instantiate-with-cw-serde](https://github.com/desmos-labs/contract-utils/blob/paul/DCD-83/add-remarkables-utils/types/contracts/cw721-poap/instantiate_msg.d.ts) and [cw721-poap-instantiate-without-cw-serde](https://github.com/desmos-labs/contract-utils/blob/paul/DCD-83/add-remarkables-utils/types/contracts/cw721-base/instantiate_msg.d.ts)
3. without `[k: string]: unknown` field, message is no longer to match `Record<string, unknown>`

### DIFF
--- a/packages/cosmwasm-stargate/src/cosmwasmclient.ts
+++ b/packages/cosmwasm-stargate/src/cosmwasmclient.ts
@@ -71,7 +71,7 @@ export interface ContractCodeHistoryEntry {
   /** The source of this history entry */
   readonly operation: "Genesis" | "Init" | "Migrate";
   readonly codeId: number;
-  readonly msg: Record<string, unknown>;
+  readonly msg: Object;
 }
 
 /** Use for testing only */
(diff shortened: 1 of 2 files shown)


webmaster128:
Interesting. I came across a this issue recently as well when writing a `interface MyManualMessageTypeDefinition { ... }`. Wanted to change this to `any`. But maybe this fix is better.

webmaster128:
I picked up this work and extended it a bit in https://github.com/cosmos/cosmjs/pull/1281. Thanks a lot!

PR #1282, opened 2022-10-08
Fix aminomessages/createWasmAminoConverters

honeymaro:
In some cases, `funds` may not exist.


webmaster128:
Why should the field not exist? It is required and not marked as `omitempty` in Go. What's the use case for this?

webmaster128:
Closing since `funds` is not optional. It should always be set. Use an empty array if no funds should be set.

Feel free to open a ticket if there are cases where the field is not set. Then we have to go for bug hunting, possibly here or in other code bases.

PR #1300, opened 2022-10-20
Batch RPC requests

codehans:
Proof of concept re https://github.com/cosmos/cosmjs/issues/1298

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,10 @@ and this project adheres to
 - @cosmjs/utils: Add `isDefined` which checks for `undefined` in a
   TypeScript-friendly way.
 - @cosmjs/stargate: Add missing `{is,}MsgBeginRedelegateEncodeObject`,
-  `{is,MsgCreateValidatorEncodeObject}` and
-  `{is,MsgEditValidatorEncodeObject}`.
+  `{is,MsgCreateValidatorEncodeObject}` and `{is,MsgEditValidatorEncodeObject}`.
 - @cosmjs/tendermint-rpc: Add `HttpBatchClient`, which implements `RpcClient`,
   supporting batch RPC requests ([#1300]).
-  
+
 [#1300]: https://github.com/cosmos/cosmjs/pull/1300
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to
   TypeScript-friendly way.
 - @cosmjs/stargate: Add missing `{is,}MsgBeginRedelegateEncodeObject`,
   `{is,MsgCreateValidatorEncodeObject}` and
-  `{is,MsgEditValidatorEncodeObject}`.\
+  `{is,MsgEditValidatorEncodeObject}`.
 - @cosmjs/tendermint-rpc: Add `HttpBatchClient`, which implements `RpcClient`,
   supporting batch RPC requests ([#1300]).
   
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,10 @@ and this project adheres to
 - @cosmjs/utils: Add `isDefined` which checks for `undefined` in a
   TypeScript-friendly way.
 - @cosmjs/stargate: Add missing `{is,}MsgBeginRedelegateEncodeObject`,
-  `{is,MsgCreateValidatorEncodeObject}` and `{is,MsgEditValidatorEncodeObject}`.
+  `{is,MsgCreateValidatorEncodeObject}` and
+  `{is,MsgEditValidatorEncodeObject}`.\
+- @cosmjs/tendermint-rpc: Add `HttpBatchClient`, which implements `RpcClient`,
+  supporting batch RPC requests
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,9 @@ and this project adheres to
   `{is,MsgCreateValidatorEncodeObject}` and
   `{is,MsgEditValidatorEncodeObject}`.\
 - @cosmjs/tendermint-rpc: Add `HttpBatchClient`, which implements `RpcClient`,
-  supporting batch RPC requests
+  supporting batch RPC requests ([#1300]).
+  
+[#1300]: https://github.com/cosmos/cosmjs/pull/1300
 
 ### Fixed
 
--- a/packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
+++ b/packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
@@ -58,7 +58,8 @@ export class HttpBatchClient implements RpcClient {
   }
 
   private validate(): void {
-    if (this.options.batchSizeLimit < 1) throw new Error("batchSizeLimit < 1");
+    if (!this.options.batchSizeLimit || this.options.batchSizeLimit < 1)
+      throw new Error("batchSizeLimit < 1");
   }
 
   private async tick(): Promise<void> {
--- a/packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
+++ b/packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
@@ -8,7 +8,7 @@ import {
 import { http, HttpEndpoint } from "./httpclient";
 import { hasProtocol, RpcClient } from "./rpcclient";
 
-export interface HttpClientOptions {
+export interface HttpBatchClientOptions {
   dispatchInterval: number;
   batchSizeLimit: number;
 }
--- a/packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
+++ b/packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
@@ -4,8 +4,8 @@ import {
   JsonRpcSuccessResponse,
   parseJsonRpcResponse,
 } from "@cosmjs/json-rpc";
-import { http, HttpEndpoint } from "./httpclient";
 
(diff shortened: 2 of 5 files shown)


packages/tendermint-rpc/src/rpcclients/httpclient.ts
```diff
@@ -65,29 +65,67 @@ export interface HttpEndpoint {
   readonly headers: Record<string, string>;
 }
 
+export interface HttpClientOptions {
+  dispatchInterval: number;
+}
+
+export const defaultHttpClientOptions: HttpClientOptions = { dispatchInterval: 0 };
+
 export class HttpClient implements RpcClient {
```
webmaster128, line 74: Let's make this a second client please (which we then can use by default once tested and stabilized). This keeps us flexible testing both implementations and going back if we run into trouble for some cases.

packages/tendermint-rpc/src/rpcclients/httpclient.ts
```diff
@@ -65,29 +65,67 @@ export interface HttpEndpoint {
...
   protected readonly url: string;
   protected readonly headers: Record<string, string> | undefined;
+  protected readonly options: HttpClientOptions;
 
-  public constructor(endpoint: string | HttpEndpoint) {
+  private stack: Array<{
```
webmaster128, line 79: This should be a queue (first in, first out), not a stack (first in, last out). You already implement it that way. But the variable name should be adjusted.

packages/tendermint-rpc/src/rpcclients/httpclient.ts
```diff
@@ -65,29 +65,67 @@ export interface HttpEndpoint {
...
-    return response;
+    return new Promise((resolve, reject) => {
+      this.stack.push({ request, resolve, reject });
+    });
+  }
+
+  private async tick(): Promise<void> {
+    // Avoid race conditions
+    const stack = this.stack;
+    this.stack = [];
```
webmaster128, line 110: Here I'd limit the number of jobs that are processed. Sending 100s or 1000s of requests in one request is probably not a good idea. Maybe `.shift` 20 items and leave the rest in the queue.

packages/tendermint-rpc/src/rpcclients/httpclient.ts
```diff
@@ -65,29 +65,67 @@ export interface HttpEndpoint {
   readonly headers: Record<string, string>;
 }
 
+export interface HttpClientOptions {
+  dispatchInterval: number;
+}
+
+export const defaultHttpClientOptions: HttpClientOptions = { dispatchInterval: 0 };
```
webmaster128, line 72: This runs a CPU burner loop, right?

I think something like 20ms should be a sane interval time as this delay is still small compared to the network roundtrip.

codehans: oof yes nice catch

codehans:
@webmaster128 PR updated. httpbatchclient had a dependency on the `http` in `httpclient` now, can pull that out into something shared if need be. figured that if batch may become the default then this may be unnecessary

packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
```diff
@@ -0,0 +1,74 @@
...
+  isJsonRpcErrorResponse,
+  JsonRpcRequest,
+  JsonRpcSuccessResponse,
+  parseJsonRpcResponse,
+} from "@cosmjs/json-rpc";
+
+import { http, HttpEndpoint } from "./httpclient";
+import { hasProtocol, RpcClient } from "./rpcclient";
+
+export interface HttpClientOptions {
```
webmaster128, line 11:
```suggestion
export interface HttpBatchClientOptions {
```

packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
```diff
@@ -0,0 +1,74 @@
...
+      this.url = hasProtocol(endpoint) ? endpoint : "http://" + endpoint;
+    } else {
+      this.url = endpoint.url;
+      this.headers = endpoint.headers;
+    }
+    setInterval(() => this.tick(), options.dispatchInterval);
+  }
+
+  public disconnect(): void {
+    // nothing to be done
```
webmaster128, line 42: This would be a good place to stop the interval. `setInterval` returns some sort of ID and there is an API to stop it.

packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  public async execute(request: JsonRpcRequest): Promise<JsonRpcSuccessResponse> {
+    return new Promise((resolve, reject) => {
+      this.queue.push({ request, resolve, reject });
+    });
+  }
+
+  private async tick(): Promise<void> {
+    // Avoid race conditions
+    const queue = this.queue.splice(0, this.options.batchSizeLimit);
```
webmaster128, line 53: Can you add a check that `batchSizeLimit` is set and >= 1? Just throw if that's not the case.

I would like to avoid running into special behaviour for the decond argument of `splice`:

>  If deleteCount is 0 or negative, no elements are removed. In this case, you should specify at least one new element (see below).

packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import { http, HttpEndpoint } from "./httpclient";
+import { hasProtocol, RpcClient } from "./rpcclient";
+
+export interface HttpClientOptions {
+  dispatchInterval: number;
+  batchSizeLimit: number;
+}
+
+export const defaultHttpClientOptions: HttpClientOptions = { dispatchInterval: 20, batchSizeLimit: 20 };
```
webmaster128, line 16:
```suggestion
export const defaultHttpBatchClientOptions: HttpClientOptions = { dispatchInterval: 20, batchSizeLimit: 20 };
```

webmaster128 (commented):
Beautiful

packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
```diff
@@ -0,0 +1,87 @@
...
+      this.headers = endpoint.headers;
+    }
+    this.timer = setInterval(() => this.tick(), options.dispatchInterval);
+    this.validate();
+  }
+
+  public disconnect(): void {
+    this.timer && clearInterval(this.timer);
+    this.timer = undefined;
+  }
```
webmaster128, line 52: Nice idempotent implementation

packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  public async execute(request: JsonRpcRequest): Promise<JsonRpcSuccessResponse> {
+    return new Promise((resolve, reject) => {
+      this.queue.push({ request, resolve, reject });
+    });
+  }
+
+  private validate(): void {
+    if (this.options.batchSizeLimit < 1) throw new Error("batchSizeLimit < 1");
```
webmaster128, line 61: Let's check for truthiness first. There are those falsy values of `number`, like `NaN`.

```
> NaN < 1
false
```

CHANGELOG.md
```diff
@@ -17,7 +17,10 @@ and this project adheres to
 - @cosmjs/utils: Add `isDefined` which checks for `undefined` in a
   TypeScript-friendly way.
 - @cosmjs/stargate: Add missing `{is,}MsgBeginRedelegateEncodeObject`,
-  `{is,MsgCreateValidatorEncodeObject}` and `{is,MsgEditValidatorEncodeObject}`.
+  `{is,MsgCreateValidatorEncodeObject}` and
+  `{is,MsgEditValidatorEncodeObject}`.\
```
webmaster128, line 21:
```suggestion
  `{is,MsgEditValidatorEncodeObject}`.
```

CHANGELOG.md
```diff
@@ -17,7 +17,10 @@ and this project adheres to
...
+- @cosmjs/tendermint-rpc: Add `HttpBatchClient`, which implements `RpcClient`,
+  supporting batch RPC requests
```
webmaster128, line 23:
```suggestion
  supporting batch RPC requests ([#1300]).
  
[#1300]: https://github.com/cosmos/cosmjs/pull/1300
```

CHANGELOG.md
```diff
@@ -17,7 +17,12 @@ and this project adheres to
 - @cosmjs/utils: Add `isDefined` which checks for `undefined` in a
   TypeScript-friendly way.
 - @cosmjs/stargate: Add missing `{is,}MsgBeginRedelegateEncodeObject`,
-  `{is,MsgCreateValidatorEncodeObject}` and `{is,MsgEditValidatorEncodeObject}`.
+  `{is,MsgCreateValidatorEncodeObject}` and
+  `{is,MsgEditValidatorEncodeObject}`.
```
webmaster128, line 21: Hmm, still a diff in the old entry.

Can you run `yarn format-text` in the repo root?

webmaster128 (approved):
Nice stuff!

Any last changes from your side? Otherwise I merge and ship with the next patch release. Then we can start testing it in real apps.

codehans:
> Nice stuff!
> 
> Any last changes from your side? Otherwise I merge and ship with the next patch release. Then we can start testing it in real apps.

nothing from my side, all good!

actually just deployed it to our live apps (https://blue.kujira.app,  https://fin.kujira.app)

PR #1307, opened 2022-10-25
change import pattern of crypto, to allow import error to be caught a…

mvid:
crypto import:
change import pattern of crypto, to allow import error to be caught and acted on. the current model fails in react native, not exactly sure why, but using require instead avoids the issue.

Changed the lookup for `subtle`. Currently, if `subtle` doesn't exist on the crypto library, cosmjs will attempt to reimport node crypto, even if a proper crypto was already shimmed in. The changeset will pull subtle if it exists, otherwise return undefined. `getCryptoModule` will only be called if `crypto` is not already set on window/this

authz queries:
add more authz grant queries to catch up to current authz module

### DIFF
--- a/packages/crypto/src/pbkdf2.ts
+++ b/packages/crypto/src/pbkdf2.ts
@@ -11,9 +11,9 @@ import { sha512 as nobleSha512 } from "@noble/hashes/sha512";
  */
 export async function getCryptoModule(): Promise<any | undefined> {
   try {
-    const crypto = await import("crypto");
+    const crypto = await require("crypto");
     // We get `Object{default: Object{}}` as a fallback when using
-    // `crypto: false` in Webpack 5, which we interprete as unavailable.
+    // `crypto: false` in Webpack 5, which we interpret as unavailable.
     if (typeof crypto === "object" && Object.keys(crypto).length <= 1) {
       return undefined;
     }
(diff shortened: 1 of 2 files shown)


webmaster128:
Thank you. This looks both promising at first glance.

Could you split the additional queries and the crypto changes in 2 PRs please? This way we can keep the discussion specific and not block one change on the other change.

mvid:
> Thank you. This looks both promising at first glance.
> 
> Could you split the additional queries and the crypto changes in 2 PRs please? This way we can keep the discussion specific and not block one change on the other change.

Done, the authz queries PR is now in https://github.com/cosmos/cosmjs/pull/1308

packages/crypto/src/pbkdf2.ts
```diff
@@ -11,9 +11,9 @@ import { sha512 as nobleSha512 } from "@noble/hashes/sha512";
  */
 export async function getCryptoModule(): Promise<any | undefined> {
   try {
-    const crypto = await import("crypto");
+    const crypto = await require("crypto");
```
webmaster128, line 14: Could you try this instead?

```
    // HACK: Use a variable to get webpack to ignore this and cause a
    // runtime error instead of build system error or fallback implementation.
    const nodeCryptoPackageName = "crypto";
    const crypto = await import(nodeCryptoPackageName);
 ```
 
 We use this in a different place already. This should avoid build systems trying to bundle a crypto implementation when it is not available natively.

webmaster128:
The change LGTM. Good point regarding the subtle import.

I'd just like to get rid of the Node.js specific `require` call.

packages/crypto/src/pbkdf2.ts
mvid: I introduced this, however the problem is the same. It attempts to resolve `crypto` when bundling. In my current project, using Expo/React-Native/Metro, it seems to parse resolve any `import` call, hence why I switched it to a `require`. I believe this is specific to the Metro bundler. 

Any ideas on how to solve this?

webmaster128:
Sorry for letting you wait so long. I'm now trying to catch up with CosmJS.

It seems like metro allows us to use an "empty" crypto module. See
- https://facebook.github.io/metro/docs/configuration/
- https://facebook.github.io/metro/docs/resolution#resolverequest-customresolver

I never worked with metro though. Do you have a test project where I can reproduce the issues you are facing?

webmaster128:
I looked even more into this and have two changes in the pipeline:
1. getSubtle does not need to call getCryptoModule at all since I am not aware of any environment where subtle is only available in the crypto module
2. getCryptoModule can be removed entirely if we are willing to use a pure-JS pbkdf2 implementation for Node.js 14. Given that Node.js 14 EOLs  in 4 months and 3 weeks (30 Apr 2023), I think this is fine for CosmJS 0.30.0

webmaster128:
okay, this is far from solved. See https://github.com/cosmos/cosmjs/issues/1354 for a follow-up ticket.

PR #1314, opened 2022-10-29
chore: make integers errors more explicit

keupsonite:
For easy debugging.

### DIFF
--- a/packages/math/src/integers.ts
+++ b/packages/math/src/integers.ts
@@ -55,7 +55,7 @@ export class Uint32 implements Integer, WithByteConverters {
 
   public static fromString(str: string): Uint32 {
     if (!str.match(/^[0-9]+$/)) {
-      throw new Error("Invalid string format");
+      throw new Error("Invalid string format. Found: " + str);
     }
     return new Uint32(Number.parseInt(str, 10));
   }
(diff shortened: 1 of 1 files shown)


packages/math/src/integers.ts
```diff
@@ -55,7 +55,7 @@ export class Uint32 implements Integer, WithByteConverters {
 
   public static fromString(str: string): Uint32 {
     if (!str.match(/^[0-9]+$/)) {
-      throw new Error("Invalid string format");
+      throw new Error("Invalid string format. Found: " + str);
```
webmaster128, line 58: Can you put those in quotes? Probably easier to write with a template string.

```ts
throw new Error(`Invalid string format. Found: '${str}'`);
```

Thanks

packages/math/src/integers.ts
```diff
@@ -238,10 +238,10 @@ export class Uint64 implements Integer, WithByteConverters {
 
   private constructor(data: BN) {
     if (data.isNeg()) {
-      throw new Error("Input is negative");
+      throw new Error("Input is negative. Found: " + data);
```
webmaster128, line 241: Does that give you a proper debug representation of a BN?

webmaster128 (commented):
LGTM overall

packages/math/src/integers.ts
keupsonite: I had also thought about it, but I based it on https://github.com/cosmos/cosmjs/blob/2c3b27eeb3622a6108086267ba6faf3984251be3/packages/math/src/integers.ts#L75. I will adapt to reflect your suggestion.

packages/math/src/integers.ts
keupsonite: I don't think so, I will adapt this with the call toString()

keupsonite:
Should be good @webmaster128

keupsonite:
any news @webmaster128?

dynst:
The integer classes recently got rewritten to use `bigint` so there are merge conflicts now.

PR #1323, opened 2022-11-14
Improve HttpBatchClient

### LINKED ISSUE #1309: HttpBatchClient: Allow setting some options only
`HttpBatchClientOptions` currently requires all options to be set when it is used

### LINKED ISSUE #1310: HttpBatchClient: run tick() as soon as batchSizeLimit is reached
Once `batchSizeLimit` are queued, there is no need to wait any longer before sending the next request.

### LINKED ISSUE #1311: Export HttpBatchClient in tendermint-rpc's index.ts
- [ ] Export HttpBatchClient
- [ ] Export HttpBatchClientOptions

webmaster128:
Closes #1309
Closes #1310
Closes #1311

cc @codehans

### DIFF
--- a/packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
+++ b/packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
@@ -10,7 +10,9 @@ import { HttpEndpoint } from "./httpclient";
 import { hasProtocol, RpcClient } from "./rpcclient";
 
 export interface HttpBatchClientOptions {
+  /** Interval for dispatching batches (in milliseconds) */
   dispatchInterval: number;
+  /** Max number of items sent in one request */
   batchSizeLimit: number;
 }
 
@@ -58,6 +60,11 @@ export class HttpBatchClient implements RpcClient {
   public async execute(request: JsonRpcRequest): Promise<JsonRpcSuccessResponse> {
     return new Promise((resolve, reject) => {
       this.queue.push({ request, resolve, reject });
+
+      if (this.queue.length >= this.options.batchSizeLimit) {
+        // this train is full, let's go
+        this.tick();
+      }
     });
   }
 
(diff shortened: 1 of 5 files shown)


packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
```diff
@@ -68,27 +78,43 @@ export class HttpBatchClient implements RpcClient {
...
+          }
+        });
+      },
+      (error) => {
+        for (const requestId of requestIds) {
+          const req = batch.find((s) => s.request.id === requestId);
+          if (!req) return;
+          req.reject(error);
+        }
+      },
```
webmaster128, line 117: Those HTTP errors previosely got lost as unhandled promose rejections of the tick() call.

packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
```diff
@@ -55,6 +60,11 @@ export class HttpBatchClient implements RpcClient {
   public async execute(request: JsonRpcRequest): Promise<JsonRpcSuccessResponse> {
     return new Promise((resolve, reject) => {
       this.queue.push({ request, resolve, reject });
+
+      if (this.queue.length >= this.options.batchSizeLimit) {
+        // this train is full, let's go
+        this.tick();
```
codehans, line 66: Should the timer be reset in this instance?

codehans:
Looks great! Just the one comment

packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
webmaster128: I was thinking about this as well but ended up going with the simplest implementation. I guess it does not matter much.

Before this PR we had a tick after: 20ms, 40ms, 60ms, 80ms, …
With this, assuming 50 initial requests after 12ms of page load: 12ms, 12ms, 20ms, 40ms, 60ms, 80ms, …

So in case of high traffic, we get faster execution requests. The `dispatchInterval` remains the max time a request is delayed to wait for more requests but it is not a constant time.

packages/tendermint-rpc/src/rpcclients/httpbatchclient.ts
```diff
@@ -55,6 +60,11 @@ export class HttpBatchClient implements RpcClient {
   public async execute(request: JsonRpcRequest): Promise<JsonRpcSuccessResponse> {
     return new Promise((resolve, reject) => {
       this.queue.push({ request, resolve, reject });
+
+      if (this.queue.length >= this.options.batchSizeLimit) {
+        // this train is full, let's go
```
mikedotexe, line 65: choo choo! 🚄 💨

PR #1326, opened 2022-11-23
Fix protoDecimalToJson 0 fractional part

keupsonite:
0 value should be possible

### DIFF
--- a/packages/stargate/src/modules/staking/aminomessages.ts
+++ b/packages/stargate/src/modules/staking/aminomessages.ts
@@ -29,10 +29,10 @@ interface Description {
   readonly details: string;
 }
 
-function protoDecimalToJson(decimal: string): string {
+export function protoDecimalToJson(decimal: string): string {
   const parsed = Decimal.fromAtomics(decimal, 18);
   const [whole, fractional] = parsed.toString().split(".");
-  return `${whole}.${fractional.padEnd(18, "0")}`;
+  return `${whole}.${(typeof fractional !== "undefined" ? fractional : "").padEnd(18, "0")}`;
 }
 
 function jsonDecimalToProto(decimal: string): string {
--- a/packages/stargate/src/modules/staking/aminomessages.ts
+++ b/packages/stargate/src/modules/staking/aminomessages.ts
@@ -32,7 +32,7 @@ interface Description {
 export function protoDecimalToJson(decimal: string): string {
   const parsed = Decimal.fromAtomics(decimal, 18);
   const [whole, fractional] = parsed.toString().split(".");
-  return `${whole}.${(typeof fractional !== "undefined" ? fractional : "").padEnd(18, "0")}`;
+  return `${whole}.${(fractional ?? "").padEnd(18, "0")}`;
 }
 
 function jsonDecimalToProto(decimal: string): string {
--- a/packages/stargate/src/modules/index.ts
+++ b/packages/stargate/src/modules/index.ts
@@ -82,6 +82,7 @@ export {
   isAminoMsgDelegate,
   isAminoMsgEditValidator,
   isAminoMsgUndelegate,
+  protoDecimalToJson,
 } from "./staking/aminomessages";
 export {
   isMsgBeginRedelegateEncodeObject,
--- a/packages/stargate/src/modules/staking/aminomessages.spec.ts
+++ b/packages/stargate/src/modules/staking/aminomessages.spec.ts
@@ -18,10 +18,43 @@ import {
   AminoMsgEditValidator,
   AminoMsgUndelegate,
   createStakingAminoConverters,
+  protoDecimalToJson,
 } from "./aminomessages";
 
 describe("AminoTypes", () => {
   describe("toAmino", () => {
+    it("works for protoDecimalToJson", () => {
+      const testSet = [
+        {
+          decimal: "0",
+          expected: "0.000000000000000000",
+        },
+        {
+          decimal: "1",
+          expected: "0.000000000000000001",
+        },
+        {
+          decimal: "2497",
+          expected: "0.000000000000002497",
+        },
+        {
+          decimal: "987000000000000000",
+          expected: "0.987000000000000000",
+        },
+        {
+          decimal: "123987000000000000000",
+          expected: "123.987000000000000000",
+        },
+        {
+          decimal: "4872000000000000000000",
+          expected: "4872.000000000000000000",
+        },
+      ];
+      for (let i = 0; i < testSet.length; i++) {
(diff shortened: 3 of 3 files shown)


webmaster128 (commented):
What's the issue with returning "0.000000000000000000" instead of "0"? I think "0.000000000000000000" is the correct representation in JSON.

keupsonite:
@webmaster128 Decimal.fromAtomics method don't accept "0.0" and so the fractional variable is undefined and throw an error. The correct representation for JSON is "0".

webmaster128:
> Decimal.fromAtomics method don't accept "0.0"

Right, but protoDecimalToJson is not meant to parse "0.0". It only parses Cosmos SDK protobuf Dec fields, which is an integer. Where/how are you using that function?

webmaster128:
okay, there is indeed a bug here. But this is not only for 0 but all values with 0 fractional part. Decimal.toString does not always print a dot. I.e. we have to handle `fractional` being undefined. E.g. by doing `${(fractional ?? "").padEnd(18, "0")}`.

keupsonite:
> okay, there is indeed a bug here. But this is not only for 0 but all values with 0 fractional part. Decimal.toString does not always print a dot. I.e. we have to handle `fractional` being undefined. E.g. by doing `${(fractional ?? "").padEnd(18, "0")}`.

Oh okay, I'll adapt that.

webmaster128:
Thanks! Could you also make the function public (just inside of the folder) and add a few tests for `protoDecimalToJson` in `packages/stargate/src/modules/staking/aminomessages.spec.ts`?

keupsonite:
> okay, there is indeed a bug here. But this is not only for 0 but all values with 0 fractional part. Decimal.toString does not always print a dot. I.e. we have to handle `fractional` being undefined. E.g. by doing `${(fractional ?? "").padEnd(18, "0")}`.



> ```js
> 10000000000000000
> ```

I just did several tests and it seems that only 0 is a problem. Would you have a set of values to suggest to me so that I can do the test?

webmaster128:
okay, here are the relevant tests
```
Input                    Expected
"0"                      "0.000000000000000000"
"00"                     "0.000000000000000000"
"1"                      "0.000000000000000001"
"01"                     "0.000000000000000001"
"2497"                   "0.000000000000002497"
"02497"                  "0.000000000000002497"
"987000000000000000"     "0.987000000000000000"
"123987000000000000000"  "123.987000000000000000"
"4872000000000000000000" "4872.000000000000000000"
```

keupsonite:
> okay, here are the relevant tests
> 
> ```
> Input                    Expected
> "0"                      "0.000000000000000000"
> "00"                     "0.000000000000000000"
> "1"                      "0.000000000000000001"
> "01"                     "0.000000000000000001"
> "2497"                   "0.000000000000002497"
> "02497"                  "0.000000000000002497"
> "987000000000000000"     "0.987000000000000000"
> "123987000000000000000"  "123.987000000000000000"
> "4872000000000000000000" "4872.000000000000000000"
> ```

"00", "01" and "02497" values? is are these desired and acceptable values? I didn't think that these values were accepted, they shouldn't be preceded by a 0. Besides, the method `protoToDecimal` converts "01" to whole = "0" and fractional = "01" for an incorrect result of "0.010000000000000000" instead of "0.000000000000000001", because of the preceding 0.

keupsonite:
> okay, here are the relevant tests
> 
> ```
> Input                    Expected
> "0"                      "0.000000000000000000"
> "00"                     "0.000000000000000000"
> "1"                      "0.000000000000000001"
> "01"                     "0.000000000000000001"
> "2497"                   "0.000000000000002497"
> "02497"                  "0.000000000000002497"
> "987000000000000000"     "0.987000000000000000"
> "123987000000000000000"  "123.987000000000000000"
> "4872000000000000000000" "4872.000000000000000000"
> ```

I didn't think we wanted to accept values with a zero first unless there was only one, because I don't think those are correct values, but it seems that `protoDecimalToJson` is able to convert them correctly. So I will handle all cases with a zero and another digit in addition to the single zero. I will let you know when it is done. Thanks

webmaster128:
Strictly speaking the leading zeros as not expected inputs and maybe we should not even test them. I guess we get this behaviour for free when reading the string into an integer in `const parsed = Decimal.fromAtomics(decimal, 18);`. But you are right, it's not a requirement for the function, so we better don't test it.

keupsonite:
> Strictly speaking the leading zeros as not expected inputs and maybe we should not even test them. I guess we get this behaviour for free when reading the string into an integer in `const parsed = Decimal.fromAtomics(decimal, 18);`. But you are right, it's not a requirement for the function, so we better don't test it.

In this case, my initial patch is sufficient to handle all cases (except "00") and avoids further processing. Would you still prefer that I modify it to follow your recommendations? i.e:

> we have to handle `fractional` being undefined. E.g. by doing `${(fractional ?? "").padEnd(18, "0")}`.

webmaster128:
Yes, your patch fails for 4872000000000000000000 or any other number with all zero fractional part

keupsonite:
Oh yes indeed @webmaster128! I should have tested your entire test set before I spoke :D Thanks, I'll fix it asap.

packages/stargate/src/modules/staking/aminomessages.ts
```diff
@@ -29,10 +29,10 @@ interface Description {
   readonly details: string;
 }
 
-function protoDecimalToJson(decimal: string): string {
+export function protoDecimalToJson(decimal: string): string {
   const parsed = Decimal.fromAtomics(decimal, 18);
   const [whole, fractional] = parsed.toString().split(".");
-  return `${whole}.${fractional.padEnd(18, "0")}`;
+  return `${whole}.${(typeof fractional !== "undefined" ? fractional : "").padEnd(18, "0")}`;
```
webmaster128, line 35: This is the same as your code in modern notation:

```suggestion
  return `${whole}.${(fractional ?? "").padEnd(18, "0")}`;
```

keupsonite: Zero is considered false, hence the use of typeof.

keupsonite: Hum in fact "0" is considered true, it was the initial absence of parenthesis that generated an error. In your proposal the brackets are kept and it works.

webmaster128: The `??` operator only uses the fallback when the value is null or undefined. So even `0 ?? "fallback"` and `"" ?? "fallback"` preserve the original value. This is different from using `||`.

webmaster128:
I merged this as part of #1336. Thank you for bringing up this issue and your PR!

webmaster128:
Released as part of 0.29.5

PR #1328, opened 2022-11-24
Add QueryClient.queryAbci and deprecate QueryClient.queryUnverified

webmaster128:


kriskowal (approved):
Good for our needs. Thank you for illustrative tests.

PR #1340, opened 2022-12-07
Don't call getCryptoModule in getSubtle

webmaster128:


webmaster128:
@mvid this is step 1 from https://github.com/cosmos/cosmjs/pull/1307#issuecomment-1341373825

webmaster128:
Released as part of 0.29.5

packages/crypto/src/pbkdf2.ts
```diff
@@ -24,14 +24,18 @@ export async function getCryptoModule(): Promise<any | undefined> {
...
+  // In version 15 and 16 this was stored under the webcrypto key.
+  // With Node.js 17 it was moved to the same locations where browsers
+  // make it available.
+  // Loading `require("crypto")` here seems unnecessary since it only
+  // causes issues with bundlers and does not increase compatibility.
+
+  // Browsers and Node.js 17+
+  let subtle: any | undefined = (globalThis as any)?.crypto?.subtle;
+  // Node.js 15+
+  if (!subtle) subtle = (globalThis as any)?.crypto?.webcrypto?.subtle;
```
webmaster128, line 37: All those comments are wrong because they only apply to the Node.js REPL, not Node.js itself.

PR #1352, opened 2022-12-13
Test iterating over IBC clients

webmaster128:
This is a test case for https://github.com/cosmos/ibc-go/issues/2935


webmaster128:
Closing as bug was confirmed in ibc-go: https://github.com/cosmos/ibc-go/issues/2935

PR #1359, opened 2022-12-29
Add Unsafe Connect methods to TendermintClients

garrettparris:
A way to synchronously connect to the rpc without waiting for the resolution of `detectVersion`


webmaster128:
Let's go with https://github.com/cosmos/cosmjs/pull/1380, which should achieve the same without the need for new APIs. Thank you!

PR #1370, opened 2023-02-17
Fix app crashing error

paplco:
UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch().

This error will cause nodejs to crash when the connection times out. Firewalls will cause timeout.

Error could be gracefully handled.


webmaster128:
Thank you fro bringing this up. I think it is better to handle promise resolve and reject of client.connected in the place where this is called. Are you using `SocketWrapper` directly? Or through some higher level client in CosmJS?

PR #1372, opened 2023-02-18
implement instantiate permission in amino

### LINKED ISSUE #334: Implement and test MsgStoreCode.instantiate_permission
Right now `MsgStoreCode.instantiate_permission` is always undefined because

```ts
// TODO: implement
/**
 * @see https://github.com/CosmWasm/wasmd/blob/v0.10.0-alpha/x/wasm/internal/types/params.go#L68-L71
 */
export type AccessConfig = never;
```

loin3:
Closes #334

AccessType in cosmwasm proto set stringer false and implemented its own string method.


packages/cosmwasm-stargate/src/modules/wasm/aminomessages.ts
```diff
@@ -9,13 +9,51 @@ import {
...
+}
+
 /**
  * @see https://github.com/CosmWasm/wasmd/blob/v0.18.0-rc1/proto/cosmwasm/wasm/v1/types.proto#L36-L41
  */
-type AccessConfig = never;
+export interface AccessConfig {
+  /**
+   * Permission should be one kind of string 'Nobody', 'OnlyAddress', 'Everybody', 'Unspecified'
+   * @see https://github.com/CosmWasm/wasmd/blob/v0.28.0/x/wasm/types/params.go#L54
```
webmaster128, line 52:
```suggestion
   * Permission should be one kind of string 'Nobody', 'OnlyAddress', 'Everybody', 'AnyOfAddresses', 'Unspecified'
   * @see https://github.com/CosmWasm/wasmd/blob/v0.31.0/x/wasm/types/params.go#L54-L77
```

packages/cosmwasm-stargate/src/modules/wasm/aminomessages.ts
```diff
@@ -9,13 +9,51 @@ import {
...
+      return "Nobody";
+    case AccessType.ACCESS_TYPE_ONLY_ADDRESS:
+      return "OnlyAddress";
+    case AccessType.ACCESS_TYPE_EVERYBODY:
+      return "Everybody";
+    case AccessType.UNRECOGNIZED:
+    default:
+      return "UNRECOGNIZED";
+  }
+}
```
webmaster128, line 44: Could you add the AnyOfAddresses case here too?

packages/cosmwasm-stargate/src/modules/wasm/aminomessages.ts
```diff
@@ -9,13 +9,51 @@ import {
...
+      return AccessType.ACCESS_TYPE_UNSPECIFIED;
+    case "Nobody":
+      return AccessType.ACCESS_TYPE_NOBODY;
+    case "OnlyAddress":
+      return AccessType.ACCESS_TYPE_ONLY_ADDRESS;
+    case "Everybody":
+      return AccessType.ACCESS_TYPE_EVERYBODY;
+    default:
+      return AccessType.UNRECOGNIZED;
+  }
```
webmaster128, line 27: Could you add the AnyOfAddresses case here too?

webmaster128 (commented):
Very nice, thanks a lot.

Maybe you need to update the PR to the latest main branch in order to get the latest version of cosmjs-types.

Issue #1367, opened 2023-02-07
Amino encode not support some special char(<, >, $).


loin3:
When Tx memo contain special characters (<, >, $), Amino sign does not working

Step1. Write Tx with memo within special char. ex) "Tx & Send"
Step2. Sign with Secp256k1HdWallet

`signature verification failed; please verify account number (1), sequence (7) and chain-id (simd-testing): unauthorized`

loin3:
keplr-wallet already dealt with this problem, but did not contributed to cosmjs.
https://github.com/chainapsis/keplr-wallet/blob/master/packages/common/src/escape/index.ts#L8

webmaster128:
Very interesting, thank you for bringing this up. This is indeed not yet implemented.

The Keplr reference is great and helpful.

https://go.dev/play/p/p3RNTpm2LED shows the escaping bahaviour in Go.

PR #1373, opened 2023-02-20
fix fail amino sign verify with special chars

loin3:
this PR is for #1367 

I found that there is a problem with escaping special chars in keplr-wallet. If you follow this method, the value is still escaped when you check memo of tx in mintscan. Because cosmjs sends memo through protobuf, the value needs to be escaped only when signed with amino. So escaping is done only on values ​​encoded for amino signatures. This way, even if you look up the tx, the memo will not be escaped.

### DIFF
--- a/packages/amino/src/signdoc.spec.ts
+++ b/packages/amino/src/signdoc.spec.ts
@@ -1,8 +1,8 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import { Random } from "@cosmjs/crypto";
-import { toBech32 } from "@cosmjs/encoding";
+import { fromUtf8, toBech32, toUtf8 } from "@cosmjs/encoding";
 
-import { AminoMsg, makeSignDoc, sortedJsonStringify } from "./signdoc";
+import { AminoMsg, escapeCharacters, makeSignDoc, sortedJsonStringify } from "./signdoc";
 
 function makeRandomAddress(): string {
   return toBech32("cosmos", Random.getBytes(20));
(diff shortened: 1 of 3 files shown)


webmaster128 (commented):
Thank you for this PR. Since we only care about JSON strings and not the whole JSON document, I think it is cleaner and safer to use the [replacer argument in JSON.stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify?retiredLocale=de#syntax).

Here is some example how this can be used to manipulate the inner strings as part of the stringification: https://gist.github.com/Tanapruk/4ab1505ead2491229235c886b6fbfaac

NoahSaso:
hey @loin3, thanks for doing this :)

eagerly awaiting this over at DAO DAO because our users run into this bug all the time

webmaster128:
@NoahSaso if you need this quickly as part of the upcoming 0.30.0 release, have a look at https://github.com/cosmos/cosmjs/issues/1367. We should implement this differently. If you are in the mood, feel free to jump in.

webmaster128:
Sorry, I mixed up issue and PR here. I was referring to

> Thank you for this PR. Since we only care about JSON strings and not the whole JSON document, I think it is cleaner and safer to use the [replacer argument in JSON.stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify?retiredLocale=de#syntax).
>
> Here is some example how this can be used to manipulate the inner strings as part of the stringification: https://gist.github.com/Tanapruk/4ab1505ead2491229235c886b6fbfaac

loin3:
I agree that this implementation should be changed to something tidy way. because modifying the byte code directly is very ugly.

here is the reason why I should modify byte code directly.
utf8 byte code of character `&` is 38 and unicode of byte codes is `92, 117, 46, 46, 50, 54` (\u0026).
json marshaller in go automatically escape the character `&` to unicode. so when signature verifying, incoming messages's byte code should contain `92, 117, 46, 46, 50, 54` (\u0026) rather than `38`.
but the problem is JSON.stringfy authmatically change unicode (\u0026) to character &. so even though I used unicode, for example` {memo: "alice \u0026 bob"}`, unicode authmatically changed to "alice & bob".

webmaster128:
Oh, I see. This makes sense, thank you. In this case the current implementation makes sense. I'll have another look and throw a bunch of low level unit tests on the implementation to better understand it.

Thanks a lot!

webmaster128:
Oh damn, sorry, wrong PR. I did not aim to merge this yet. Reverted the merge. I'll look into this ASAP.

webmaster128:
I used this work in https://github.com/cosmos/cosmjs/pull/1388. There I migrated from binary to string based replacement and increased low level test coverage. Reviews appreciated.

PR #1378, opened 2023-03-06
Create PoC for metadata handling in ProtobufRpcClient

webmaster128:


Segfaultd:
Hey, any chance this get attention? Querying by height is possible on the chain CLI but no SDK is handling it so far.

PR #1388, opened 2023-03-09
Amino JSON string escaping (v2)

webmaster128:
Closes https://github.com/cosmos/cosmjs/issues/1367

Builds on top of #1373

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- @cosmjs/amino: Fix escaping of "&", "<" and ">" characters in Amino JSON
+  encoding to match the Go implementation ([#1373], [#1388]).
+
+[#1373]: https://github.com/cosmos/cosmjs/pull/1373
+[#1388]: https://github.com/cosmos/cosmjs/pull/1388
+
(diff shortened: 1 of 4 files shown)


loin3:
oh wow. this works perfectly and the memo seems fine when I queried the tx.
스크린샷 2023-03-10 오전 4 26 37

but I can't understand why this works and [keplr](https://github.com/chainapsis/keplr-wallet/blob/master/packages/common/src/escape/index.ts#L8) fails while escape logic is same. I sent tx with &, <, > characters through keplr and the memo was escaped. https://www.mintscan.io/cosmos/txs/1858CBE6DD792D915E0CE56174323D983FC1DBD3A429B755FBCEDF74EDFA7709

webmaster128:
Seems like Keplr does not do the escaping at the correct level
- https://github.com/chainapsis/keplr-wallet/blob/671429d8037aaa6133708ad93c1eba526f37009e/packages/background/src/keyring/service.ts#L363-L366
- https://github.com/chainapsis/keplr-wallet/blob/671429d8037aaa6133708ad93c1eba526f37009e/packages/background/src/keyring/service.ts#L429

So the content of the memo is changed before the encoding starts. So one effect this could have is the memo of the sign doc is not the memo of the protobuf transaction.

Anyways, I'm not familiar with the Keplr codebase.

PR #1394, opened 2023-03-22
Tendermint34Client maxAge

codehans:
On the Kujira dapps we use a community sourced set of RPC clients to provide chain connections for users, in order to promote redundancy and decentralisation. We use a `Promise.any` approach to selecting an RPC provider when a UI initialises, in order to pick one with low latency to the user.  Sometimes some of these clients fall behind yet still have a fast response time, particularly when you're accessing from a geographically underserved location.  This results in stale data, account sequence mismatches, etc. 

This PR adds an optional `maxAge` parameter to the `connect` function of `Tendermint34Client`, and throws the connection if the reported `latest_block_time` is stale. 

@webmaster128  - will sort out tests etc for this if you feel it's a useful addition. As it stands, we're pushing for decentralisation of infrastructure, but this is making the dApps unusable for some connections.

### DIFF
--- a/packages/tendermint-rpc/src/tendermint34/tendermint34client.ts
+++ b/packages/tendermint-rpc/src/tendermint34/tendermint34client.ts
@@ -13,13 +13,23 @@ import { adaptor34, Decoder, Encoder, Params, Responses } from "./adaptor";
 import * as requests from "./requests";
 import * as responses from "./responses";
 
+export interface Tendermint34ClientOptions {
+  /**
+   * Max permitted diff in ms between node status.sync_info.latest_block_time and new Date()
+   */
+  maxAge?: number;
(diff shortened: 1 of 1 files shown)


webmaster128 (commented):
Very interesting idea and sane implementation. I like not making a single Tendermint client responsible for the balancing but having that at a higher level.

webmaster128:
Is this still relevant? I like the idea in general. but in the meantime we just removed the `detectVersion` PRC call on connect as a bunch of folks complained about the extra request in production. So if we want to implement, we need a slightly different solution.

I am closing this now as the current implementation cannot be merged after https://github.com/cosmos/cosmjs/pull/1772. But we can continue the discussion in an issue.

Issue #1316, opened 2022-10-31
Can we skip polling when using `broadcastTx` ?


williamchong:
Currently `broadcastTx` always polls tx for block inclusion. While `timeoutMs` and `pollIntervalMs` is configurable, setting them to 0 would make the function instantly throws a timeout error as a side effect.
Allowing polling to be skipped in `broadcastTx()` without throwing an error would allow batch sending of many transactions to be easier.

(I understand many messages can be included in one transaction, however in some cases we would want to send batch transactions with different memos)

webmaster128:
Tracking the block inclusion is an essential feature of `broadcastTx`. If you don't need or want that feature, it buils down to just this line:

```ts
const broadcasted = await this.forceGetTmClient().broadcastTxSync({ tx });
```

which you can call directly if you have a `Tendermint34Client` at hand. You can do that manually.

An alternative approach to using Tendermint34Client is keeping the polling but not waiting for the promise resolutions. In https://github.com/noislabs/nois-bot/blob/74a5f9c8fc64c325f0d078420169a8514e3c3dbf/index.js#L205-L222 you find a multi-cast implementation where 3 `broadcastTx` to different RPCs are made in parallel. Once everything is sent, the code checks all the results.

williamchong:
it's actually hard to get `Tendermint34Client` from `stargateSigningClient` since it's a private getter
To make usage easier, do you think skipping polling if `timeoutMs` is set to special value 0 or -1 makes sense? If yes I can try to make a PR

webmaster128:
What if we add a `broadcastTxSync` next to `broadcastTx` which does not track the block inclusion? It would only contain those lines and then return after checkTx:
https://github.com/cosmos/cosmjs/blob/e89a91401f50326b528cce5fc189f8aed3ee400b/packages/stargate/src/stargateclient.ts#L473-L479

williamchong:
Yes that would work well 👍

webmaster128:
@williamchong Can you have a look at #1396 to see if it solves your problem?

PR #1396, opened 2023-03-27
Improve DX - Stargate sign and broadcast

DavideSegullo:
Hi, as stated in the title of the PR I would like to suggest the inclusion of a new method to sign a tx and then not have to wait for it to be inserted into a block.

Currently it is possible to do this, but by putting workarounds with ts-ignore, I think it is more appropriate to add a method directly.

### Why?
In nabla, we are working on a [utility](https://github.com/nabla-studio/cosmos-txs-tracer) to track transactions using websockets, so it is useful to get the hash of the tx right away and then pass it to the utility, as polling is limited.


webmaster128 (commented):
This is cool, thank you for your contribution! It closes this ticket and uses the solution I suggested over there: https://github.com/cosmos/cosmjs/issues/1316.

I just have two change requests:
- Could you rename `*WithoutPolling` -> `*Sync` to follow the naming given by the backend? Also polling is an implementation detail that can change over time.
- Can you deduplicate the code for all cases such that e.g. `broadcastTx` uses `broadcastTxSync` for the first steps of the operation (accepted by the node)?

DavideSegullo:
> This is cool, thank you for your contribution! It closes this ticket and uses the solution I suggested over there: #1316.
> 
> I just have two change requests:
> 
> * Could you rename `*WithoutPolling` -> `*Sync` to follow the naming given by the backend? Also polling is an implementation detail that can change over time.
> * Can you deduplicate the code for all cases such that e.g. `broadcastTx` uses `broadcastTxSync` for the first steps of the operation (accepted by the node)?

Yes, it seems fine to me, I did some changes, is it okay now?

webmaster128 (commented):
Diff LGTM, thanks!

Could you add one test for each new function (one `describe` block per function)? No need to copy all the cases we have for signAndBroadcast but at least one happy path to ensure this does not break over time.

And then we need a CHANGELOG entry in the Unreleased section for this work.

Once we have the two things, this can get shipped in 0.31.0. Thank you!

DavideSegullo:
> signAndBroadcast

Ofc,  I have added some tests, please tell me if they are enought!

packages/stargate/src/signingstargateclient.spec.ts
```diff
@@ -690,6 +690,313 @@ describe("SigningStargateClient", () => {
...
+        };
+        const fee = {
+          amount: coins(2000, "ucosm"),
+          gas: "222000", // 222k
+        };
+        const memo = "Use your power wisely";
+        const transactionHash = await client.signAndBroadcastSync(faucet.address0, [msgAny], fee, memo);
+       
+        expect(transactionHash).toMatch(/^[0-9A-F]{64}$/);
+      });
```
webmaster128, line 721: Could you end all the new tests with

```ts
await sleep(simapp.blockTime * 1.5); // wait for block inclusion to update account sequence for next test
```

Otherwise the tests ends with a tx in the mempool and the next test gets the old sequence.

packages/stargate/src/signingstargateclient.spec.ts
```diff
@@ -690,6 +690,329 @@ describe("SigningStargateClient", () => {
...
+
+        const msg = MsgDelegate.fromPartial({
+          delegatorAddress: faucet.address0,
+          validatorAddress: validator.validatorAddress,
+          amount: coin(1234, "ustake"),
+        });
+        const msgAny: MsgDelegateEncodeObject = {
+          typeUrl: "/cosmos.staking.v1beta1.MsgDelegate",
+          value: msg,
+        };
```
webmaster128, line 712: Okay, looking at the CI failure we need a different message type here. The additional MsgDelegate makes an unrelated staking query fail. When we change to some simple MsgSend the problem should go away.

packages/stargate/src/signingstargateclient.spec.ts
webmaster128: Seems like this worked!

packages/stargate/src/signingstargateclient.spec.ts
DavideSegullo: Done

packages/stargate/src/signingstargateclient.spec.ts
```diff
@@ -690,6 +690,333 @@ describe("SigningStargateClient", () => {
...
+          amount: coins(2000, "ucosm"),
+          gas: "200000",
+        };
+        const memo = "Use your power wisely";
+        const transactionHash = await client.signAndBroadcastSync(faucet.address0, [msgAny], fee, memo);
+        
+        expect(transactionHash).toMatch(/^[0-9A-F]{64}$/);
+
+        await sleep(simapp.blockTime * 1.5); 
+      });
```
webmaster128, line 1016:
```suggestion
```

webmaster128 (commented):
Let's remove those delegation tests. Otherwise the staking about in some other test fails. This is not great right now as the staking balance tests depends on the MsgDelegate tests. But let's move on here and get this in.

packages/stargate/src/signingstargateclient.spec.ts
webmaster128: It seems like I don't have permission to commit this change

DavideSegullo:
> Let's remove those delegation tests. Otherwise the staking about in some other test fails. This is not great right now as the staking balance tests depends on the MsgDelegate tests. But let's move on here and get this in.

Done, I approved it!

webmaster128:
Okay it seems like only a linting issue left. Could you fix them (with `yarn && yarn lint-fix` in the repo root)?

webmaster128:
Seems like I can't push here. Could you allow changes by maintaners? Then I can write a CHANGELOG entry, push here and merge to get it in the 0.31 release.

DavideSegullo:
> Seems like I can't push here. Could you allow changes by maintaners? Then I can write a CHANGELOG entry, push here and merge to get it in the 0.31 release.

Do you know how I can enable it?

> Okay it seems like only a linting issue left. Could you fix them (with `yarn && yarn lint-fix` in the repo root)?

Done

webmaster128:
Thanks, see here: https://github.com/bisq-network/style/issues/4

DavideSegullo:
> Thanks, see here: [bisq-network/style#4](https://github.com/bisq-network/style/issues/4)

I tried it, but I don't see "Allow edits from maintainers" button

DavideSegullo:
Well, I found out that the problem is related to cross-org PRs: https://github.com/orgs/community/discussions/5634

I'll try to give you permissions on that repo

Issue #1448, opened 2023-06-29
Missing earliest block in RPC sync info


livthomas:
Cosmjs only parses the following fields from `sync_info` record returned by `/status` RPC method:

- catching_up
- latest_app_hash
- latest_block_hash
- latest_block_height
- latest_block_time

But there are also other fields present in `sync_info` which are not processed by cosmjs:

- earliest_app_hash
- earliest_block_hash
- earliest_block_height
- earliest_block_time

# Steps to Reproduce

1. Try to hit `/status` endpoint on any public RPC node (e.g. https://cosmos-rpc.rockrpc.net/status).
2. See what fields are returned within `result.sync_info`.
3. Look at `RpcSyncInfo` interface in both [tendermint34](https://github.com/cosmos/cosmjs/blob/21d150e7a101db5ae2e4542acc94ff5382f7f7c1/packages/tendermint-rpc/src/tendermint34/adaptor/responses.ts#L621-L629) and [tendermint37](https://github.com/cosmos/cosmjs/blob/21d150e7a101db5ae2e4542acc94ff5382f7f7c1/packages/tendermint-rpc/src/tendermint37/adaptor/responses.ts#L622-L630) adaptors.

livthomas:
Yes, I'm already working on it.

webmaster128:
Done in #1449

PR #1449, opened 2023-06-29
fix(tendermint-rpc): add earliest block to sync info

livthomas:
This should fix issue #1448. Those fields are returned from `/status` endpoint but I haven't tested this code. I'm open to any suggestions how I can test it easily.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- @cosmjs/tendermint-rpc: Add missing `earliest_*` fields to `SyncInfo` record
+  returned from the `/status` RPC endpoint ([#1448]).
+
 ## [0.31.0] - 2023-06-22
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


webmaster128:
See
- https://github.com/cosmos/cosmjs/blob/main/packages/tendermint-rpc/src/tendermint34/tendermint34client.spec.ts#L218-L220
- https://github.com/cosmos/cosmjs/blob/main/packages/tendermint-rpc/src/tendermint37/tendermint37client.spec.ts#L218-L220

for the place to put the tests. This is tested against a real Tendermint instance in CI

webmaster128:
Thew CI runs Tendermint version 0.34 and 0.37.

Looking at the code you linked there it seems like the fields may not be set if the metadata about the earliest block is missing
Bildschirmfoto 2023-06-29 um 23 30 45

So all 4 fields are probably optional and some sort of unset/default/zero value. Could you just dump `status.syncInfo` and check what you actually get from the node?

webmaster128 (approved):
Thank you!

Issue #1413, opened 2023-04-11
Add FungibleTokenPacketData for sendIbcTokens


Reecepbcups:
Can we get FungibleTokenPacketData memo so we can create Middleware UIs ([packet forward middleware](https://github.com/strangelove-ventures/packet-forward-middleware) specifically)? 
It requires JSON encoded values in this to properly work. This works for the CLI but not on cosmjs

Working CLI example
```
osmosisd tx ibc-transfer transfer transfer channel-0 cosmos10r39fueph9fq7a6lgswu4zdsg8t3gxlqvvvyvn 2uosmo --from reece --gas=250000 --gas-prices=0.003uosmo --node https://rpc.osmosis.zone:443 --chain-id=osmosis-1 --memo "{ \"forward\": { \"receiver\": \"juno10r39fueph9fq7a6lgswu4zdsg8t3gxlq670lt0\", \"port\": \"transfer\", \"channel\": \"channel-207\" }}"
```

Even though this is --memo, it's not the 'correct' memo for Keplr / cosmjs

https://github.com/cosmos/ibc-go/blob/v5.1.0/proto/ibc/applications/transfer/v2/packet.proto

webmaster128:
Can you provide a dump of the unsigned transaction from the `osmosisd tx ibc-transfer transfer transfer ...` command? Is this the transaction memo or the MsgTransfer memo? The later is currently not supported but could be.

Reecepbcups:
@webmaster128 
```json
[...]

PR #1456, opened 2023-08-01
Add memo field to AminoMsgTransfer

shanev:
This enables Packet Forwarding Middleware to work with IBC Hooks. As discussed in #1413 and based on work by @jhernandezb and @luke-ae.

### DIFF
--- a/packages/stargate/src/modules/ibc/aminomessages.spec.ts
+++ b/packages/stargate/src/modules/ibc/aminomessages.spec.ts
@@ -40,6 +40,7 @@ describe("AminoTypes", () => {
             revision_number: "456",
           },
           timeout_timestamp: "789",
+          memo: undefined,
         },
       };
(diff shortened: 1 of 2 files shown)


webmaster128:
Diff LGTM. Unfortunately we cannot easily test it against a chain because simapp does not have IBC anymore.

Could you please allow modification from maintainers? Then I can add a CHANGELOG entry and merge.

webmaster128:
Ping

shanev:
> Diff LGTM. Unfortunately we cannot easily test it against a chain because simapp does not have IBC anymore.
> 
> Could you please allow modification from maintainers? Then I can add a CHANGELOG entry and merge.

Can you check now?

webmaster128:
Amino support for the MsgTransfer memo should work as of 0.31.3, released today. See https://github.com/cosmos/cosmjs/issues/1493 for an overview of the state of affairs.

PR #1486, opened 2023-10-12
Add Client class to serve as a general purpose base without message types

webmaster128:


jgimeno:
Hi Simon, I wonder if this can also solve the oposite scenario. Imagine that I have some data in proto format coming from a query, in the SDK I can use reflection and the interfaceRegistry to get back at least the json representation of that code.

Do you know if this is something that can be done with cosmjs?

webmaster128:
> Do you know if this is something that can be done with cosmjs?

We designed CosmJS with a lot of type safety in mind. A fully dynamic approach would give you no guarantees at compile time. I think both approaches are valid but a client-library focussed on reflections would be a different projects with different trade-offs. I'm happy if someone builds that but it's out of scope for CosmJS.

PR #1484, opened 2023-10-04
Upgrade to new cosmjs-types (Telescope 1.0, big int 🔥)

webmaster128:
Closes #1425

PR #1489, opened 2023-10-16
feat(timeout-height): allow define optional timeout height parameter

janfabian:
Introduce timeout height option for signing transactions (amino and direct) to support new POB/block-sdk modules. 

Skip's documentation for details: https://docs.skip.money/searcher-docs


packages/cosmwasm-stargate/src/signingcosmwasmclient.spec.ts
```diff
@@ -1195,6 +1195,44 @@ describe("SigningCosmWasmClient", () => {
...
+          [msgAny],
+          fee,
+          memo,
+          undefined,
+          Long.fromNumber(height + 1),
+        );
+
+        // ensure signature is valid
+        const result = await client.broadcastTx(Uint8Array.from(TxRaw.encode(signed).finish()));
+        assertIsDeliverTxSuccess(result);
```
webmaster128, line 1232: Could you add a test (in both cases) where the timeout height is exceeded? You can use `client.getHeight()-1` or something like that for the timeout height.

packages/cosmwasm-stargate/src/signingcosmwasmclient.spec.ts
```diff
@@ -1195,6 +1195,44 @@ describe("SigningCosmWasmClient", () => {
...
+        };
+        const memo = "Use your power wisely";
+        const height = await client.getHeight();
+        const signed = await client.sign(
+          alice.address0,
+          [msgAny],
+          fee,
+          memo,
+          undefined,
+          Long.fromNumber(height + 1),
```
webmaster128, line 1227: This test is likely a bit flawky because block times are short in CI. Let's stabilize it a bit.

```suggestion
          Long.fromNumber(height + 3),
```

packages/cosmwasm-stargate/src/signingcosmwasmclient.spec.ts
```diff
@@ -1413,6 +1451,44 @@ describe("SigningCosmWasmClient", () => {
...
+        };
+        const memo = "Use your tokens wisely";
+        const height = await client.getHeight();
+        const signed = await client.sign(
+          alice.address0,
+          [msgAny],
+          fee,
+          memo,
+          undefined,
+          Long.fromNumber(height + 1),
```
webmaster128, line 1483:
```suggestion
          Long.fromNumber(height + 3),
```

packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
```diff
@@ -644,6 +644,7 @@ export class SigningCosmWasmClient extends CosmWasmClient {
     fee: StdFee,
     memo: string,
     explicitSignerData?: SignerData,
+    timeoutHeight?: Long,
```
webmaster128, line 647: Can you make that a bigint? CosmJS 0.32 will pretty much get rid of long (see #1484)

packages/amino/src/signdoc.ts
```diff
@@ -61,6 +62,7 @@ export function makeSignDoc(
   memo: string | undefined,
   accountNumber: number | string,
   sequence: number | string,
+  timeout_height?: string,
```
webmaster128, line 65: let's make this bigint for some some type safety

webmaster128 (commented):
Great stuff, love it. High quality contribution ❤️

Some smaller changes. Then this can go in 0.32.

packages/cosmwasm-stargate/src/signingcosmwasmclient.spec.ts
webmaster128: Nice updates. As far as I can see only those thests are missing: timeout height exceeded.

janfabian: exactly, just added those

packages/cosmwasm-stargate/src/signingcosmwasmclient.spec.ts
```diff
@@ -1191,6 +1191,73 @@ describe("SigningCosmWasmClient", () => {
...
+          gas: "222000", // 222k
+        };
+        const memo = "Use your power wisely";
+        const height = await client.getHeight();
+        const signed = await client.sign(alice.address0, [msgAny], fee, memo, undefined, BigInt(height - 1));
+
+        try {
+          await client.broadcastTx(Uint8Array.from(TxRaw.encode(signed).finish()));
+        } catch (e: any) {
+          assert(e.code === 30);
```
webmaster128, line 1253:
```suggestion
          expect(e.code).toEqual(30);
```

packages/cosmwasm-stargate/src/signingcosmwasmclient.spec.ts
```diff
@@ -1191,6 +1191,73 @@ describe("SigningCosmWasmClient", () => {
...
+          return;
```
webmaster128, line 1254: Doesn't this return the function without disconnecting?

You can also write expected exceptions like this. Also check out other `expectAsync` usages in the codebase

```
      await expectAsync(
        client.queryContractSmart(nonExistentAddress, { verifier: {} }),
      ).toBeRejectedWithError(/no such contract/i);
```

janfabian: It should run finally block, but anyway `expectAsync` is the function I was looking for. Thanks :) 

btw I like when things go full circle https://github.com/jasmine/jasmine/issues/1625#issuecomment-444810001

webmaster128: Thanks, looks nicer!

webmaster128 (commented):
Could you also add the parameter to `signAndBroadcastSync` and `signAndBroadcast`? I think signAndBroadcast is the highest level where people might need this. No need to write extra tests for those but just propagate the parameter.

webmaster128 (approved):
Beautiful, thank you!

PR #1496, opened 2023-10-26
Fix typo in comment

twhy:


packages/stargate/src/stargateclient.ts
```diff
@@ -108,7 +108,7 @@ export interface DeliverTxResponse {
   readonly height: number;
   /** The position of the transaction within the block. This is a 0-based index. */
   readonly txIndex: number;
-  /** Error code. The transaction suceeded iff code is 0. */
+  /** Error code. The transaction suceeded if code is 0. */
```
webmaster128, line 111: Oh, this is not a typo. It's short for "if and only if", an abbreviation used by math nerds.

webmaster128, line 111:
```suggestion
  /** Error code. The transaction suceeded if and only if code is 0. */
```

PR #1506, opened 2023-11-09
[amino]: fix remove unnecessary any type assertion

hyperbolic-versor:
remove unnecessary `any` type assertion


webmaster128:
As you can see in the CI build, they seem to be needed.

You can check `yarn && yarn build` locally. Closing for now. Feel free to open another PR or issue if I missed something. Thanks!

PR #1508, opened 2023-11-12
Add pagination option to bank query

AmitPr:
We're running into cases on Kujira where `allBalances` needs to be paginated to retrieve all of the balances. Exposing this to make it easier for us to do so.

### DIFF
--- a/packages/stargate/src/modules/bank/queries.ts
+++ b/packages/stargate/src/modules/bank/queries.ts
@@ -8,13 +8,14 @@ import {
   QueryTotalSupplyResponse,
 } from "cosmjs-types/cosmos/bank/v1beta1/query";
 import { Coin } from "cosmjs-types/cosmos/base/v1beta1/coin";
+import type { PageRequest } from "cosmjs-types/helpers";
 
 import { createPagination, createProtobufRpcClient, QueryClient } from "../../queryclient";
 
 export interface BankExtension {
   readonly bank: {
     readonly balance: (address: string, denom: string) => Promise<Coin>;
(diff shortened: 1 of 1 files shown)


packages/stargate/src/modules/bank/queries.ts
```diff
@@ -8,13 +8,14 @@ import {
...
 import { Coin } from "cosmjs-types/cosmos/base/v1beta1/coin";
+import type { PageRequest } from "cosmjs-types/helpers";
 
 import { createPagination, createProtobufRpcClient, QueryClient } from "../../queryclient";
 
 export interface BankExtension {
   readonly bank: {
     readonly balance: (address: string, denom: string) => Promise<Coin>;
-    readonly allBalances: (address: string) => Promise<Coin[]>;
+    readonly allBalances: (address: string, pagination?: PageRequest) => Promise<Coin[]>;
```
webmaster128, line 18: I see. But this is inconsistent with the other APIs that have pagination exposed like `totalSupply`.

Unfortunately this PR falls into point number 3 of things that don't scale:
https://github.com/cosmos/cosmjs/issues/1440
I don't know a great way to handle this stuff long term and patching query extensions left and right is not sustainable.

AmitPr: Yeah agreed, this really is a Monkey patch; Hopefully more ergonomic way to pass through advanced options comes along

webmaster128: Do you use any Telescope tooling to generate queries for your chain? There I think you get the full proto API exposed by default.

PR #1516, opened 2023-11-30
fix: Ensure RFC dates between years 0001 and 0099 are parsed correctly

clockworkgr:
This fixes an issue where an RFC3339 Date like `0001-01-01T00:00:00Z` (which is used as the default/empty/ value in `CommitSig.Timestamp` when `BlockIdFlag` is `ABSENT`) was parsed as `1901-01-01T00:00:00Z` due to the way `Date.UTC()` treats years 0 to 99.

This led to incorrect timestamps being generated.

Unit test has been added.

### DIFF
--- a/packages/encoding/src/rfc3339.ts
+++ b/packages/encoding/src/rfc3339.ts
@@ -40,13 +40,13 @@ export function fromRfc3339(str: string): Date {
 
   const tzOffset = tzOffsetSign * (tzOffsetHours * 60 + tzOffsetMinutes) * 60; // seconds
 
-  let timestamp = Date.UTC(year, month - 1, day, hour, minute, second, milliSeconds);
+  const date = new Date();
+  date.setUTCFullYear(year);
+  date.setUTCMonth(month - 1);
+  date.setUTCDate(day);
+  date.setUTCHours(hour, minute, second, milliSeconds);
 
-  // Date.UTC maps year 0-99 to 1900-1999. Ensure the correct year is set and THEN apply the offset
-  const date = new Date(timestamp);
-  timestamp = date.setUTCFullYear(year) - tzOffset * 1000;
-
-  return new Date(timestamp);
+  return new Date(date.getTime() - tzOffset * 1000);
 }
 
 export function toRfc3339(date: Date | ReadonlyDate): string {
--- a/packages/encoding/src/rfc3339.ts
+++ b/packages/encoding/src/rfc3339.ts
@@ -41,9 +41,7 @@ export function fromRfc3339(str: string): Date {
   const tzOffset = tzOffsetSign * (tzOffsetHours * 60 + tzOffsetMinutes) * 60; // seconds
 
   const date = new Date();
-  date.setUTCFullYear(year);
-  date.setUTCMonth(month - 1);
-  date.setUTCDate(day);
+  date.setUTCFullYear(year, month - 1, day);
   date.setUTCHours(hour, minute, second, milliSeconds);
 
   return new Date(date.getTime() - tzOffset * 1000);
--- a/packages/encoding/src/rfc3339.ts
+++ b/packages/encoding/src/rfc3339.ts
@@ -40,7 +40,12 @@ export function fromRfc3339(str: string): Date {
 
   const tzOffset = tzOffsetSign * (tzOffsetHours * 60 + tzOffsetMinutes) * 60; // seconds
 
-  const timestamp = Date.UTC(year, month - 1, day, hour, minute, second, milliSeconds) - tzOffset * 1000;
+  let timestamp = Date.UTC(year, month - 1, day, hour, minute, second, milliSeconds);
+
+  // Date.UTC maps year 0-99 to 1900-1999. Ensure the correct year is set and THEN apply the offset
+  const date = new Date(timestamp);
+  timestamp = date.setUTCFullYear(year) - tzOffset * 1000;
+
   return new Date(timestamp);
 }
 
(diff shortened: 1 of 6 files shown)


clockworkgr:
Related: https://github.com/confio/cosmjs-types/issues/89

clockworkgr:
This change, combined with the fact we no longer allow empty/unset fields in cosmjs-types means that this:

https://github.com/cosmos/cosmjs/issues/704#issuecomment-797122415

must be addressed by no longer treating the timestamp as optional.

packages/encoding/src/rfc3339.ts
```diff
@@ -40,7 +40,12 @@ export function fromRfc3339(str: string): Date {
 
   const tzOffset = tzOffsetSign * (tzOffsetHours * 60 + tzOffsetMinutes) * 60; // seconds
 
-  const timestamp = Date.UTC(year, month - 1, day, hour, minute, second, milliSeconds) - tzOffset * 1000;
+  let timestamp = Date.UTC(year, month - 1, day, hour, minute, second, milliSeconds);
+
+  // Date.UTC maps year 0-99 to 1900-1999. Ensure the correct year is set and THEN apply the offset
+  const date = new Date(timestamp);
+  timestamp = date.setUTCFullYear(year) - tzOffset * 1000;
+
```
webmaster128, line 48: Alright, let's see. Few observations here:

- A time before epoch seems well supported:
  > The maximum timestamp representable by a Date object is slightly smaller than the maximum safe integer ([Number.MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER), which is 9,007,199,254,740,991). A Date object can represent a maximum of ±8,640,000,000,000,000 milliseconds, or ±100,000,000 (one hundred million) days, relative to the epoch. This is the range from April 20, 271821 BC to September 13, 275760 AD. Any attempt to represent a time outside this range results in the Date object holding a timestamp value of [NaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN), which is an "Invalid Date".
- The two year issue seems to be a known problem: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date?retiredLocale=de#interpretation_of_two-digit_years and
  > Note: Some methods, including the Date() constructor, Date.UTC(), and the deprecated [getYear()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getYear)/[setYear()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setYear) methods, interpret a two-digit year as a year in the 1900s. For example, new Date(99, 5, 24) is interpreted as June 24, 1999, not June 24, 99. See [Interpretation of two-digit years](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date?retiredLocale=de#interpretation_of_two-digit_years) for more information.

So this all makes sense. Now I think the code it a bit hard to follow with using `Date.UTC` first and then patching its broken behaviour. Can we instead create an empty Date (e.g. `new Date()`) and then call `setUTCFullYear(yearValue, monthValue, dateValue)`, `setUTCHours(hoursValue, minutesValue, secondsValue, msValue)`. Then apply the offset. I think the current code is correct, but confusing to audit.

packages/tendermint-rpc/src/comet38/adaptor/responses.ts
```diff
@@ -467,22 +467,11 @@ type RpcSignature = {
...
-  const nonZeroTime = timestamp && !timestamp.startsWith("0001-01-01");
-  return nonZeroTime ? fromRfc3339WithNanoseconds(timestamp) : undefined;
-}
-
 function decodeCommitSignature(data: RpcSignature): CommitSignature {
   return {
     blockIdFlag: decodeBlockIdFlag(data.block_id_flag),
     validatorAddress: data.validator_address ? fromHex(data.validator_address) : undefined,
-    timestamp: decodeOptionalTime(data.timestamp),
+    timestamp: data.timestamp ? fromRfc3339WithNanoseconds(data.timestamp) : undefined,
```
webmaster128, line 474: sweet

packages/encoding/src/rfc3339.ts
clockworkgr: Why the explicit implementation btw instead of simply using `Date.parse()` ?

packages/tendermint-rpc/src/comet38/adaptor/responses.ts
clockworkgr: wanna run this past @ethanfrey who brought up the original issue that that hacky `decodeOptionalTime()` was supposed to solve?

packages/encoding/src/rfc3339.ts
webmaster128: Good question. I think the main reason was nanosecond support in the fractional part. But it might also be because of implementation inconsistencies.

packages/encoding/src/rfc3339.ts
```diff
@@ -40,8 +40,13 @@ export function fromRfc3339(str: string): Date {
 
   const tzOffset = tzOffsetSign * (tzOffsetHours * 60 + tzOffsetMinutes) * 60; // seconds
 
-  const timestamp = Date.UTC(year, month - 1, day, hour, minute, second, milliSeconds) - tzOffset * 1000;
-  return new Date(timestamp);
+  const date = new Date();
+  date.setUTCFullYear(year);
+  date.setUTCMonth(month - 1);
+  date.setUTCDate(day);
```
webmaster128, line 46:
```suggestion
  date.setUTCFullYear(year, month - 1, day);
```

This should work too according to docs

webmaster128 (approved):
nice. Could you add a CHANGELOG entry for this?

webmaster128:
Released in 0.32.1 – thanks a lot!

PR #1522, opened 2023-12-11
Update validation of GasPrice.fromString to allow using ibc denoms as gas denom

madrezaz:
Currently the `GasPrice.fromString` method claims that the denom validation matches the "[Cosmos SDK 0.42 pattern](https://github.com/cosmos/cosmos-sdk/blob/v0.42.4/types/coin.go#L599-L601)"; But the regex being used for validation is not the same as the one used in the sdk. Because of that, creating a gas price using IBC denoms which contain "/" is not possible and throws an error. This PR intends to update the validation regex to fix this issue.

### DIFF
--- a/packages/stargate/src/fee.ts
+++ b/packages/stargate/src/fee.ts
@@ -37,7 +37,7 @@ export class GasPrice {
    */
   public static fromString(gasPrice: string): GasPrice {
     // Use Decimal.fromUserInput and checkDenom for detailed checks and helpful error messages
-    const matchResult = gasPrice.match(/^([0-9.]+)([a-zA-Z][a-zA-Z0-9/:._-]*)$/i);
+    const matchResult = gasPrice.match(/^([0-9.]+)([a-zA-Z][a-zA-Z0-9/:._-]*)$/);
     if (!matchResult) {
       throw new Error("Invalid gas price string");
     }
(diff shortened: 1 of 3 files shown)


webmaster128 (commented):
Thank for the fix. Could you keep the structure of the code as it was before? This provides much more detailed error messages to the user in case something is wrong.

madrezaz:
@webmaster128 sure. although the reason I changed the structure was the consistency of the regex with the one in the sdk.

packages/stargate/src/fee.ts
```diff
@@ -37,7 +37,7 @@ export class GasPrice {
    */
   public static fromString(gasPrice: string): GasPrice {
     // Use Decimal.fromUserInput and checkDenom for detailed checks and helpful error messages
-    const matchResult = gasPrice.match(/^([0-9.]+)([a-z][a-z0-9]*)$/i);
+    const matchResult = gasPrice.match(/^([0-9.]+)([a-zA-Z][a-zA-Z0-9/:._-]*)$/i);
```
webmaster128, line 40: If you add A-Z here then we should remove the `i` (case insensitive)

```suggestion
    const matchResult = gasPrice.match(/^([0-9.]+)([a-zA-Z][a-zA-Z0-9/:._-]*)$/);
```

webmaster128 (commented):
Thank you! Could you rebase that to latest main? I made a bis of an mistake before where the 0.32.1 release was published to npm but not yet pushed to GitHub. I.e. the new CHANGELOG entry should move into the Unreleased section above 0.32.1.

webmaster128 (approved):
Thanks

webmaster128:
okay, just the linter: `yarn && yarn lint-fix` should do

webmaster128:
released today as part of 0.32.2

PR #1530, opened 2023-12-22
feat: make account parser implementation extendable 

0xpatrickdev:
- adds `AccountParserManager` class that enables registering `AccountParser`s
- adds `createAccountParserRegistry`, a map of `typeUrl`s and `AccountParser`'s
- preserves existing `accountFromAny` functionality

### DIFF
--- a/packages/stargate/src/accounts.ts
+++ b/packages/stargate/src/accounts.ts
@@ -40,49 +40,71 @@ function accountFromBaseAccount(input: BaseAccount): Account {
  */
 export type AccountParser = (any: Any) => Account;
 
+export type AccountParserRegistry = Map<string, AccountParser>;
+
+export class AccountParserManager {
+  private readonly registry = new Map<string, AccountParser>();
+
+  public constructor(initialRegistry: AccountParserRegistry = new Map()) {
+    this.registry = initialRegistry;
+  }
+
+  public register(typeUrl: string, parser: AccountParser): void {
+    this.registry.set(typeUrl, parser);
+  }
+
+  public parseAccount(input: Any): Account {
+    const parser = this.registry.get(input.typeUrl);
+    if (!parser) {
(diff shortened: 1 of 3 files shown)


packages/stargate/src/accounts.ts
```diff
@@ -40,49 +40,103 @@ function accountFromBaseAccount(input: BaseAccount): Account {
...
+/**
+ * Basic implementation of AccountParser. This is supposed to support the most relevant
+ * common Cosmos SDK account types. If you need support for additional account types,
+ * you'll need to use `AccountParserManager` and `createAccountParserRegistry` directly.
+ *
+ * @example
+ * ```
+ * const myAccountParserManager = new AccountParserManager(createAccountParserRegistry());
+ * myAccountParserManager.register('/custom.type.v1.CustomAccount', customAccountParser);
+ * const account = customParserManager.parseAccount(someInput);
```
0xpatrickdev, line 136:
```suggestion
 * const { register, parseAccount } = new AccountParserManager(createAccountParserRegistry());
 * register('/custom.type.v1.CustomAccount', customAccountParser);
 * const options: StargateClientOptions = { accountParser: parseAccount };
```

This probably aligns closer to expected usage

packages/stargate/src/accounts.ts
```diff
@@ -40,49 +40,103 @@ function accountFromBaseAccount(input: BaseAccount): Account {
  */
 export type AccountParser = (any: Any) => Account;
 
+export type AccountParserRegistry = Map<Any["typeUrl"], AccountParser>;
```
0xpatrickdev, line 43: Realizing we may also want to export `accountFromBaseAccount`, as this will likely be needed by those supplying a custom account parser.  Alternatively, this could be baked into every `parseAccount` call as every AccountParser function likely needs this as the last step. 

Will wait for feedback on this and the overall design before making any updates.

PR #1536, opened 2024-01-02
fix(amino/coin): ibc denoms should be parsable

keupsonite:
Applied same regex as https://github.com/cosmos/cosmjs/commit/643bf25a8a5449a00ad8c11bce01b5da6db4f6f4#diff-9136eafb27b821610a184d11b99ace152206c81924606263f9443a75ff2865aeR28.

ps: the commit specifies that it has been merged into main and yet I don't see any changes in version v0.32.2 either via npm or this github repository.

### DIFF
--- a/packages/amino/src/coins.ts
+++ b/packages/amino/src/coins.ts
@@ -60,7 +60,7 @@ export function parseCoins(input: string): Coin[] {
     .split(",")
     .filter(Boolean)
     .map((part) => {
-      const match = part.match(/^([0-9]+)([a-zA-Z]+)/);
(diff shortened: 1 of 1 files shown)


webmaster128:
The [change you are referring to](https://github.com/cosmos/cosmjs/pull/1522) is for `GasPrice.fromString`, i.e. parses a single amount+denom pair. parseCoins is different as it is for a list of coins. Here we have the ibc issue as well.

Unfortunately the diff is hard to review and introduces decimal amounts without any additional test. Will close this and add IBC support to parseCoins in a deparate PR.

webmaster128:
Done in https://github.com/cosmos/cosmjs/pull/1554. Turns out, `import { parseCoins } from "@cosmjs/proto-signing";` had the IBC support already but this used a more advanced version than the `@cosmjs/amino` one (for historic heasons).

PR #1538, opened 2024-01-06
fix(stargate/allBalances): use pagination if there are more than 100 balances

phips28:
I have a Kujira wallet that has more than 100 balances/tokens (~116 at the time writing), therefore the allBalances function does not return **all** balances, only 100.

This PR changes check if there is a pagination key and gets the next page of balances.


phips28:
@webmaster128 can you review/merge that pls?

NotJeremyLiu:
Hey folks, I'm from the Skip Go (formerly known as Skip API) team. This is a high priority improvement from our use case and is a big cause of customer complaints. Is there anything we can do to up-prioritize including this?

BlockJayn:
I've tried to find a way to reach out to @webmaster128, no luck unfortunately...

PR #1541, opened 2024-01-13
Feat:upgrade yarn to 4.0.2

hoangdv2429:
this PR will upgrade cosmjs to use latest yarn version at the moment 4.0.2.


webmaster128:
Thank you! But you break down all the upgrades one by one? I now have yarn 3 and yarn 4 upgrades here: #1553 #1552

PR #1542, opened 2024-01-13
Feat: Add authz amino converter

hoangdv2429:
This PR will add Authz amino converter back to Cosmjs as it's working in Cosmos SDK again. Some review required on the usage of Bigint and Any object wrapped in AminoMsg.


hoangdv2429:
@webmaster128 if you're not too busy can you spend some time review this PR sir ? Thank you 🙏

PR #1561, opened 2024-02-27
fix: set date in advance during faucet to prevent multiple requests from the same address

### LINKED ISSUE #1556: Rapidly sending multiple requests to the faucet can obtain multiple test coins
Hello, I am trying to use cosmjs/faucet. 

However, I have noticed that if I send the same request multiple times in a short period, either through CLI or the frontend, I can obtain multiple test coins, thereby bypassing the rule (https://github.com/cosmos/cosmjs/blob/main/packages/faucet/src/api/webserver.ts#L69-L77) set that one can only receive coins once within a certain period. 

Is it necessary to implement a exclusive lock to prevent the same address from making multiple requests within a short period?

170210:
> Since the await faucet.credit(address, matchingDenom) call only returns when the transaction is in a block, the date is set too late. We should swap the two.

close #1556

### DIFF
--- a/packages/faucet/src/api/webserver.ts
+++ b/packages/faucet/src/api/webserver.ts
@@ -83,9 +83,9 @@ export class Webserver {
           }
 
           try {
-            await faucet.credit(address, matchingDenom);
             // Count addresses to prevent draining
             this.addressCounter.set(address, new Date());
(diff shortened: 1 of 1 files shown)


webmaster128 (approved):
Thanks

PR #1563, opened 2024-02-28
fix for sdk >= v50

pynixwang:
+ remove parsedLogs
+ fix findAttribute
+ fill logs with []


packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
```diff
@@ -327,11 +326,10 @@ export class SigningCosmWasmClient extends CosmWasmClient {
     if (isDeliverTxFailure(result)) {
       throw new Error(createDeliverTxResponseErrorMessage(result));
     }
-    const parsedLogs = logs.parseRawLog(result.rawLog);
-    const contractAddressAttr = logs.findAttribute(parsedLogs, "instantiate", "_contract_address");
+    const contractAddressAttr = logs.findAttribute(result.events, "instantiate", "_contract_address");
     return {
       contractAddress: contractAddressAttr.value,
-      logs: parsedLogs,
+      logs: [],
```
webmaster128, line 332: Nice, let's deperacte this field

packages/stargate/src/logs.ts
```diff
@@ -69,9 +69,8 @@ export function parseRawLog(input = "[]"): readonly Log[] {
  *
  * Throws if the attribute was not found.
  */
-export function findAttribute(logs: readonly Log[], eventType: string, attrKey: string): Attribute {
-  const firstLogs = logs.find(() => true);
-  const out = firstLogs?.events
+export function findAttribute(events: readonly Event[], eventType: string, attrKey: string): Attribute {
+  const out = events
```
webmaster128, line 73: Let's keep this function unchanged since it is public. Instead we can make a new private one in cosmwasm-stargate to do the job we need.

scripts/wasmd/send_first.js
```diff
@@ -15,15 +15,15 @@ const rpcUrl = "http://localhost:26659";
 const prefix = "wasm";
 const faucet = {
   mnemonic:
-    "economy stock theory fatal elder harbor betray wasp final emotion task crumble siren bottom lizard educate guess current outdoor pair theory focus wife stone",
-  address0: "wasm1pkptre7fdkl6gfrzlesjjvhxhlc3r4gm32kke3",
+    "flat ski gap size ankle rifle congress month nice state follow mechanic produce cube moment boss enlist disorder during logic brother ride erosion member",
```
webmaster128, line 18: Why this change?

scripts/wasmd/setup.sh
```diff
@@ -0,0 +1,11 @@
...
+set -o errexit -o nounset -o pipefail
+
+SCRIPT_DIR="$(realpath "$(dirname "$0")")"
+source "$SCRIPT_DIR"/env
+
+
+docker run --rm \
+  --mount type=bind,source="$SCRIPT_DIR/template_v50",target=/root \
+  "$REPOSITORY:$VERSION" \
+  ./setup_wasmd.sh
```
webmaster128, line 11: I think we can use `generate_template.sh` instead of this new script

scripts/wasmd/start.sh
```diff
@@ -26,7 +26,7 @@ docker run --rm \
   --name "$CONTAINER_NAME" \
   -p "$TENDERMINT_PORT_HOST":"$TENDERMINT_PORT_GUEST" \
   -p "$LCD_API_PORT_HOST":"$LCD_API_PORT_GUEST" \
-  --mount type=bind,source="$SCRIPT_DIR/template",target=/template \
+  --mount type=bind,source="$SCRIPT_DIR/template_v50",target=/template \
```
webmaster128, line 29: We can overrise template since we have no way to switch between different wasmd version anyways

webmaster128 (commented):
Great stuff, thank you. Let me finish this up since I need the upgrade as well quickly.

webmaster128:
I'll follow up on your work in https://github.com/cosmos/cosmjs/pull/1564. Thanks again!

scripts/wasmd/send_first.js
pynixwang: this key has balance

PR #1566, opened 2024-03-01
Simpify queries code with object property shorthand

twhy:


twhy:
@webmaster128  Hi Simon, could you please review this when you have time? thanks.

webmaster128:
Please avoid opening issues that add no value to the project. If we decide the object property shorthand is better (which I don't agree with), such things should only be done through auto-fixing lint tooling.

PR #1584, opened 2024-04-29
🐛 Fix inconsistent gas multiplier value between `signAndBroadcast` and `signAndBroadcastSync` methods

AlexanderFSP:
Synchronizes the gas multiplier value between the `#signAndBroadcast` and `#signAndBroadcastSync` methods so that it is equal to `1.4` everywhere.

### DIFF
--- a/packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
+++ b/packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
@@ -202,6 +202,9 @@ export class SigningCosmWasmClient extends CosmWasmClient {
   private readonly signer: OfflineSigner;
   private readonly aminoTypes: AminoTypes;
   private readonly gasPrice: GasPrice | undefined;
+  // Starting with Cosmos SDK 0.47, we see many cases in which 1.3 is not enough anymore
+  // E.g. https://github.com/cosmos/cosmos-sdk/issues/16020
+  private readonly gasMultiplier = 1.4;
 
   /**
    * Creates an instance by connecting to the given CometBFT RPC endpoint.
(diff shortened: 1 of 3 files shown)


packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
```diff
@@ -202,6 +202,9 @@ export class SigningCosmWasmClient extends CosmWasmClient {
   private readonly signer: OfflineSigner;
   private readonly aminoTypes: AminoTypes;
   private readonly gasPrice: GasPrice | undefined;
+  // Starting with Cosmos SDK 0.47, we see many cases in which 1.3 is not enough anymore
+  // E.g. https://github.com/cosmos/cosmos-sdk/issues/16020
+  private readonly gasMultiplier = 1.4;
```
webmaster128, line 207: Nice, can you call this `defaultGasMultiplier` since it is only used when the user does not specify anything

packages/stargate/src/signingstargateclient.ts
```diff
@@ -110,6 +110,9 @@ export class SigningStargateClient extends StargateClient {
   private readonly signer: OfflineSigner;
   private readonly aminoTypes: AminoTypes;
   private readonly gasPrice: GasPrice | undefined;
+  // Starting with Cosmos SDK 0.47, we see many cases in which 1.3 is not enough anymore
+  // E.g. https://github.com/cosmos/cosmos-sdk/issues/16020
+  private readonly gasMultiplier = 1.4;
```
webmaster128, line 115: here as well

webmaster128 (commented):
LGTM with the renaming. Could you add a CHANGELOG entry to the Unreleased section?

PR #1592, opened 2024-06-06
feat(faucet) add google recaptcha v2 validation

wwwehr:
when the secret key is set in environment variable, it activates the check for each `/credit` call.


webmaster128:
This is great work, thank you. However, I would prefer to not invest in the faucet app any further in this library. Ideally it remains a minimal demo.

Feel free to fork the faucet code and publish it as a separate project with all the great additions

PR #1593, opened 2024-06-07
fix(faucet): `isValidAddress` should accept all bech32 addresses

0xpatrickdev:
This PR updates the `isValidAddress` function to accept a wider range of valid addresses, including account addresses from Interchain Accounts (ICS-27).  The updated implementation removes the fixed length check and replaces it with a more flexible validation based on the Bech32 specification from [BIP173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#bech32).


packages/faucet/src/addresses.ts
```diff
@@ -1,12 +1,14 @@
...
+/**
+ * Check length based on Bech32 from {@link https://github.com/bitcoin/bips/blob/e1e7b77c027b3d40d07d306cc75c2b5859c91db2/bip-0173.mediawiki#bech32 | BIP173}
+ */
 export function isValidAddress(input: string, requiredPrefix: string): boolean {
   try {
     const { prefix, data } = fromBech32(input);
-    if (prefix !== requiredPrefix) {
-      return false;
-    }
-    return data.length === 20;
```
webmaster128, line 9: I think the input length should be checked in fromBech32 already. Let's just fix this by checking the data range

```ts
return data.length >= 20 && data.length <= 32;
```

webmaster128 (approved):
Thanks!

PR #1602, opened 2024-08-16
backport 0.31x: add http timeout & more data decode tx resp

ducphamle2:
- Add a timeout value for the `http()` function, preventing long requests from hanging.
- Return additional data fields in `decodeTxResponse` in case of custom responses like timestamp, block info...
- Return additional tx fields in `txsQuery` of StargateClient for custom query responses.

We have been patching this in different projects for quite some time. Eg: [oraidex-sdk](https://github.com/oraichain/oraidex-sdk/blob/main/patches/%40cosmjs%2Btendermint-rpc%2B0.31.3.patch)

It would be inconvenient to just include patch logic in every repository we need. Also, contributing to cosmjs is also a good way to help improve the tool for the community

### DIFF
--- a/packages/stargate/src/stargateclient.ts
+++ b/packages/stargate/src/stargateclient.ts
@@ -510,6 +510,7 @@ export class StargateClient {
     return results.txs.map((tx): IndexedTx => {
       const txMsgData = TxMsgData.decode(tx.result.data ?? new Uint8Array());
       return {
+        ...tx,
         height: tx.height,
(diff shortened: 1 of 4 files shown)


webmaster128:
Let's break this down into separate PRs please such that we discuss and merge the changes one by one. Good things are in here, others are not ready for merge yet. For the timeout you can use #1660 once 0.35.0 is released.

PR #1606, opened 2024-09-06
Expose explicitSignerData for simulation method and reduce rpc node requests

DavideSegullo:
At the moment when we want to perform a simulation, using the `simulate` method exposed in the `SigningCosmWasmClient` and `SigningCosmWasmClient` classes, it is not possible to specify the `sequence`, but a request is made each time to get it.

However, this in practical use results in sending multiple requests to the RPC nodes, for example when we go to use the `signAndBroadcast` method at the moment three requests are made to the nodes:
1) getAccount (fetched by the `simulation` method)
2) getAccount (fetched by the `sign` method)
3) getStatus

So we make the same request twice in a row to get the account details, specifically `sequence` and `accountNumber`.

The goal of this PR is to remove this redundancy and reduce the number of requests made to nodes in the chain.

### DIFF
--- a/packages/cosmwasm-stargate/src/signingcosmwasmclient.spec.ts
+++ b/packages/cosmwasm-stargate/src/signingcosmwasmclient.spec.ts
@@ -102,6 +102,32 @@ describe("SigningCosmWasmClient", () => {
       expect(gasUsed).toBeLessThanOrEqual(140_000);
       client.disconnect();
     });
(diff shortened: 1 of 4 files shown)


ValarDragon (approved):
Much needed fix, tysm :pray:

angelorc:
Is there a chance we’ll see this merged? 😄

PR #1607, opened 2024-09-06
Remove an unnecessary RPC request due to detectVersion method

DavideSegullo:
Right now, due to CI/CD, requirements tendermint clients send an unnecessary request to the blockchain nodes, this PR address this issue.

### DIFF
--- a/jasmine-global-variables.js
+++ b/jasmine-global-variables.js
@@ -1,0 +1,13 @@
+const process = {
+  ...process,
+  env: {
+    ...process.env,
+    HTTPSERVER_ENABLED: 1,
+    SIMAPP44_ENABLED: 1,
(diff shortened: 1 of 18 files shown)


ValarDragon (approved):
LGTM, thanks for adding this

webmaster128:
Thank you. This is a nice approach for sure. I tested the full removal again and it turns out we don't need the extra call anymore. See #1772 which will be released as part of 0.35.0.

DavideSegullo:
> Thank you. This is a nice approach for sure. I tested the full removal again and it turns out we don't need the extra call anymore. See #1772 which will be released as part of 0.35.0.

Amazing!!! Thank you!

PR #1614, opened 2024-11-08
fix: update `signAmino` method to use `messages` directly

fmorency:
This change simplifies the code by removing unnecessary mapping of messages.

This PR also acts as a workaround for https://github.com/cosmology-tech/telescope/issues/660 where `toAmino` and `fromAmino` are not symmetrical.

The [ManifestJS](https://github.com/liftedinit/manifestjs) and [Manifest WebApp](https://github.com/liftedinit/manifest-app) projects use this fix. Both project are now able to properly sign AMINO transactions, including using Keplr and Ledger in-browser.

[ManifestJS](https://github.com/liftedinit/manifestjs) has a Starship test suite covering some of the use-cases that are failing without this fix.


fmorency:
The CI failure are not related to the changes in this PR. The CI is misconfigured and needs to be updated.

PR #1625, opened 2025-02-04
Support CometBFT v1.0

cor:
v1.0 should use the Comet38Client


webmaster128:
Looks good to me. In https://github.com/cosmos/cosmjs/pull/1679 I started testing the Comet38 client against 1.0 backends. Seems to work except for [one regression bug found](https://github.com/cometbft/cometbft/issues/5219).

PR #1638, opened 2025-04-14
Fix duplicate condition check in CustomMsgDelegate encode method

leopardracer:
This PR fixes a redundant condition check in the `encode` method of `CustomMsgDelegate`. The current implementation checks `message.customAmount !== undefined` twice in the same if statement, which is unnecessary. This change simplifies the code by removing the duplicate check while maintaining the same functionality.

The fix makes the code cleaner and more maintainable without changing the behavior of the application.


leopardracer:
@Zygimantass

leopardracer:
@Zygimantass

leopardracer:
@Zygimantass

webmaster128 (commented):
Thanks. Could you do the same in `packages/stargate/src/signingstargateclient.spec.ts`?

webmaster128:
Also could you please rebase to latest main? Then the CI should work again

leopardracer:
> Also could you please rebase to latest main? Then the CI should work again

rebased but have conflicts

webmaster128:
oO. Then it is probably easier to start a fresh PR than trying to fix this

PR #1645, opened 2025-05-04
replace axios with cross-fetch

### LINKED ISSUE #1627: replace Axios with Fetch API
# Problem

Axios adds weight and is not compatible with frozen prototypes,
- https://github.com/axios/axios/pull/6265

Browsers and now NodeJs support [WHATWG Fetch](https://fetch.spec.whatwg.org/). 

It's available in all LTS versions of Node.js and is no longer experimental.

# Suggested solution

Revert https://github.com/cosmos/cosmjs/commit/182b1e778cbfdcbc73f95f2de8345e5c0613fb00

Remove "axios" dependency

dynst:
Closes #1627


webmaster128:
Finished in #1655, thank you!

PR #1666, opened 2025-06-12
migrate to @scure/base for base64 & bech32 encoding

dynst:
TypeScript return types are narrowed in a way that's still backwards-compatible.

### DIFF
--- a/packages/encoding/src/bech32.ts
+++ b/packages/encoding/src/bech32.ts
@@ -1,7 +1,7 @@
-import * as bech32 from "bech32";
+import { bech32 } from "@scure/base";
 
-export function toBech32(prefix: string, data: Uint8Array, limit?: number): string {
-  const address = bech32.encode(prefix, bech32.toWords(data), limit);
+export function toBech32(prefix: string, data: Uint8Array, limit?: number): `${string}1${string}` {
+  const address: `${typeof prefix}1${string}` = bech32.encode(prefix, bech32.toWords(data), limit);
   return address;
 }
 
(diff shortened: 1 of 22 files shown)


webmaster128:
Thanks. LFS budget was fixed now.

I am happy for the new implementation. But can we do that without changing the types of CosmJS interfaces? Is it possible to use bech32 decoding without knowing the prefix?

dynst:
> But can we do that without changing the types of CosmJS interfaces? Is it possible to use bech32 decoding without knowing the prefix?

For a backwards compatible 0.33.2 release, existing TypeScript code can't break if it's expecting a string to be returned, and a more-specific-kind-of-string gets returned now, right?

And yeah, decoding doesn't require knowing the prefix in advance, only the return type is changed (in a way that should be compatible), the arguments passed into the function are unchanged.

packages/encoding/src/bech32.ts
```diff
@@ -1,15 +1,15 @@
...
+  const address: `${typeof prefix}1${string}` = bech32.encode(prefix, bech32.toWords(data), limit);
   return address;
 }
 
 export function fromBech32(
   address: string,
   limit = Infinity,
 ): { readonly prefix: string; readonly data: Uint8Array } {
-  const decodedAddress = bech32.decode(address, limit);
+  const decodedAddress = bech32.decode(address as `${string}1${string}`, limit);
```
webmaster128, line 12: This is an unsafe cast, isn't it? `address` might be any string but we tell `bech32.decode` it is guaranteed to have a certain structure. I guess if we add an

```ts
function isBech32Structure(input: string): input is ${string}1${string} {
    // return false if malformatted
}
```

we get a safe type narrowing

webmaster128:
I split this work into https://github.com/cosmos/cosmjs/pull/1825 and https://github.com/cosmos/cosmjs/issues/1826. Thank you for pushing this foreward

PR #1676, opened 2025-06-17
Remove SES from codebase

webmaster128:
This feature was unmaintainerd/unused for way too long. Better remove it. We can add it back later, no problem

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,8 +53,6 @@ jobs:
           SOCKETSERVER_ENABLED: 1
           SKIP_BUILD: 1
           WASMD_ENABLED: 1
-          # Stopped working for Node.js 20. Maybe we need to upgrade the dependency or something.
-          # SES_ENABLED: 1
         run: yarn test --stream
       - name: Run CLI selftest
         working-directory: packages/cli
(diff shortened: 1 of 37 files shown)


kriskowal:
Can I get a handle to what composed poorly between ses, cosmjs, and Node.js 20? If we release a ses that addresses the composition problem, are you open to a PR that essentially reverts this change?

webmaster128:
Okay cool. Just submit a PR. We have node 18/20/22/24 in GitHub Actions testing now, which should run for all PRs.

dynst:
> Okay cool. Just submit a PR. We have node 18/20/22/24 in GitHub Actions testing now, which should run for all PRs.

They're actually configured not to run without manual approval.

https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks

> Depending on the "Approval for running fork pull request workflows from contributors" setting, workflows on pull requests to public repositories will not run automatically and may need approval if:

kriskowal:
> > Okay cool. Just submit a PR. We have node 18/20/22/24 in GitHub Actions testing now, which should run for all PRs.
> 
> They're actually configured not to run without manual approval.

Handy, much obliged. I’m investigating whether upgrading SES is sufficient. https://github.com/cosmos/cosmjs/pull/1689

PR #1686, opened 2025-06-18
yarn dedupe

dynst:
It might make sense to have a script run `yarn dedupe` automatically at some point.

Edit: The new Github Actions don't run on a pull request without manual approval, is that intentional?

https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks

> Depending on the "Approval for running fork pull request workflows from contributors" setting, workflows on pull requests to public repositories will not run automatically and may need approval if:

### DIFF
--- a/.pnp.cjs
+++ b/.pnp.cjs
@@ -308,7 +308,7 @@ const RAW_RUNTIME_STATE =
     ],\
     [\
       "@jridgewell/gen-mapping",\
-      "npm:0.3.2"\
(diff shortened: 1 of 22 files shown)


webmaster128:
> is that intentional

Not intentional. It is a side effect of gradually migrating from CircleCI to GH Actions. I will check this out the next couple of weeks.

PR #1692, opened 2025-06-20
add files generated by yarn install to .gitignore

dynst:
Fix on #788

https://yarnpkg.com/features/pnp

`.pnp.loader.mjs` never changes, and `.pnp.cjs` (pnp's replacement for a `node_modules/` directory) is just a <del>22k+</del> 23k line file derived from from the 8k line `yarn.lock`, except it's much more prone to git merge conflicts.

The best way to solve merge conflicts is to delete it and regenerate it (in cases where there were no conflicts in yarn.lock), so best practice is to just delete it ahead of time.


webmaster128:
Thank you for bringing this up. Turns our that this depends on Zero-Installs vs. non-Zero-Installs. See https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored and https://yarnpkg.com/features/caching#zero-installs.

I'm not necessarily convinced we need Zero-Installs but for now we use it. This is also why the whole cache is checked-in. If we give up Zero-Installs this should be discussed in a separate issue.

So I did #1696 instead. Hope it helps.

PR #1698, opened 2025-07-08
Migrate to @noble/curves

webmaster128:


stalniy:
it's awesome! Thank you @webmaster128 !

PR #1699, opened 2025-07-08
use just one eslint config file in the monorepo

dynst:
https://github.com/cosmos/cosmjs/blob/46f0c9b451a0bd1ced0f153b51fce1cbe50a4009/.eslintrc.js#L1-L6

This single errant line was preventing the use of 1 universal file for all workspaces this whole time:

https://github.com/cosmos/cosmjs/blob/cee190956bb3bbc0671a75ca5866228031f335c6/.eslintrc.js#L20

eslint searches upwards in the directory hierarchy looking for a config file.

https://eslint.org/docs/v8.x/use/configure/configuration-files#cascading-and-hierarchy

With tsconfigRootDir left undefined, it'll look for `"./tsconfig.eslint.json"` relative to the working directory where `eslint` was run from (each individual workspace), not the same directory the .eslintrc.js is in.

https://typescript-eslint.io/packages/parser/#tsconfigrootdir

Fix on #33 #788 https://github.com/cosmos/cosmjs/pull/788/commits/e2caf9bde012adb7e8d9bba8304b661c13d65122 #1375


webmaster128:
I am still struggeling to get elint to work in VS code. Not sure what's wrong here. Too many moving parts. But it did not work before and after this PR. Do you have any luck with that?

PR #1706, opened 2025-07-09
eslint config updates

dynst:
This deletes the tsconfig.eslint.json files from #788. JavaScript files shouldn't be in a tsconfig.json and typescript-eslint v8 will pretend they aren't there (leading to errors).

.js files should use  the default `"espree"` [parser](https://eslint.org/docs/v8.x/use/configure/parser
) instead. Since it's not a TypeScript parser, a modern `ecmaVersion` is needed to enable parsing this syntax.

https://github.com/cosmos/cosmjs/blob/7cc67526e9b76c0d49131d24a5bf34bc1bc7f023/packages/ledger-amino/demo/node.js#L9

And remove references to these:

https://typescript-eslint.io/rules/no-dupe-class-members
https://typescript-eslint.io/rules/ban-types/


packages/tendermint-rpc/src/jsonrpc.ts
```diff
@@ -20,8 +20,7 @@ function randomId(): number {
 }
 
 /** Creates a JSON-RPC request with random ID */
-// eslint-disable-next-line @typescript-eslint/ban-types
-export function createJsonRpcRequest(method: string, params?: {}): JsonRpcRequest {
+export function createJsonRpcRequest(method: string, params?: object): JsonRpcRequest {
```
webmaster128, line 23: I think this is the better type here. The original code was written before Record was a thing.

```suggestion
export function createJsonRpcRequest(method: string, params?: params?: Record<string, any>): JsonRpcRequest {
```

packages/tendermint-rpc/src/jsonrpc.ts
```diff
@@ -20,8 +20,7 @@ function randomId(): number {
 }
 
 /** Creates a JSON-RPC request with random ID */
-// eslint-disable-next-line @typescript-eslint/ban-types
-export function createJsonRpcRequest(method: string, params?: {}): JsonRpcRequest {
+export function createJsonRpcRequest(method: string, params?: params?: Record<string, any>): JsonRpcRequest {
```
webmaster128, line 23: Fix my own messup. Sorry
```suggestion
export function createJsonRpcRequest(method: string, params?: Record<string, any>): JsonRpcRequest {
```

PR #1712, opened 2025-07-15
feat(encoding): improve hex conversion performance

reallesee:
Replace string concatenation in toHex() with Array.from().map().join() pattern for better performance with large arrays. This approach avoids creating intermediate string objects during concatenation, resulting in better memory usage and execution speed, especially when processing large byte arrays

### DIFF
--- a/packages/encoding/src/hex.ts
+++ b/packages/encoding/src/hex.ts
@@ -1,3 +1,5 @@
+// Helper function needed to support environments that do not yet have
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/map
 function* map<T, U>(iterable: Iterable<T>, fn: (value: T) => U): IterableIterator<U> {
   for (const value of iterable) {
     yield fn(value);


webmaster128 (commented):
Sounds good. I just wonder if we can avoid creating the `Array<number>` in the first line which would be quite some overhead if not optimized away.

Would it be an option to do something like `data.values().map((byte) => ...)`? Do you have tests to measure the performance?

packages/encoding/src/hex.ts
```diff
@@ -1,9 +1,7 @@
 export function toHex(data: Uint8Array): string {
-  let out = "";
-  for (const byte of data) {
-    out += ("0" + byte.toString(16)).slice(-2);
-  }
-  return out;
+  return Array.from(data)
```
dynst, line 2:
```suggestion
  return data[Symbol.iterator]()
```

But this only works natively in Node.js 22 and ES2025, so a simple polyfill would be needed.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/map

dynst:
Sorry, I didn't think that one through. The global `class Iterator` was standardized as the same time `Iterator.prototype.map` was so of course it doesn't exist to be assigned to.

You just need a helper function for `map(data.values(), byte => ...)`. Something like...

```ts
function* map<T, U>(iterable: Iterable<T>, fn: (value: T) => U): IterableIterator<U> {
  for (const value of iterable) {
    yield fn(value);
  }
}
```

packages/encoding/src/hex.ts
```diff
@@ -1,9 +1,11 @@
-export function toHex(data: Uint8Array): string {
-  let out = "";
-  for (const byte of data) {
-    out += ("0" + byte.toString(16)).slice(-2);
+function* map<T, U>(iterable: Iterable<T>, fn: (value: T) => U): IterableIterator<U> {
```
webmaster128, line 1:
```suggestion
// Helper function needed to support environments that do not yet have
// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/map
function* map<T, U>(iterable: Iterable<T>, fn: (value: T) => U): IterableIterator<U> {
```

@dynst is this correct?

dynst (approved):

webmaster128:
Closing as long as there are no benchmarks available to back up the code changes. Feel free to re-open if you consider this change really important.

PR #1719, opened 2025-07-17
run eslint from monorepo root

dynst:
Unless there's a good reason not to run eslint on the whole repo?


webmaster128:
In https://github.com/eslint/eslint/discussions/16960 there is a good amount of users complaining about the suggestion to centralize eslint config in a repo root file. It seems like this feedback was not expected when the flat config system was developed. But starting with eslint 10 there is a new resolution system that looks up the config in the current directory and then moves up.
https://eslint.org/docs/latest/use/configure/configuration-files#experimental-configuration-file-resolution
Bildschirmfoto 2025-07-29 um 15 13 50

For us I think it is not a big deal to move everything to repo root. Linting per package is not really needed and possible by calls as you showed above.

So from my point of view we can proceed here and get this work ready for merge.

Thank you!

dynst:
I think this has to wait on the `.eslintrc.js` changes in #1719 (so typescript lints aren't run on the .js files in `scripts/`)

PR #1720, opened 2025-07-19
remove bn.js

### LINKED ISSUE #1711: replacing bn.js
https://github.com/cosmos/cosmjs/blob/318c8d5d1c5fed6c09ba5833a23ecaf81fb2eb05/packages/math/src/decimal.ts#L108-L111

This could just use `BigInt` now, right? It's been available since Node.js 10.4.

dynst:
Closes #1711

`bigint` doesn't seem to actually be missing any math?

The new byte array convesion helper function from #1698 `bytesToUnsignedBigInt()` really helps though.

### DIFF
--- a/packages/crypto/src/slip10.ts
+++ b/packages/crypto/src/slip10.ts
@@ -1,8 +1,6 @@
-import { toAscii, toHex } from "@cosmjs/encoding";
+import { fromHex, toAscii, toHex } from "@cosmjs/encoding";
 import { Uint32, Uint53 } from "@cosmjs/math";
 import { secp256k1 } from "@noble/curves/secp256k1";
-// eslint-disable-next-line @typescript-eslint/naming-convention
-import BN from "bn.js";
 
 import { Hmac } from "./hmac";
 import { Sha512 } from "./sha";
@@ -24,6 +22,12 @@ export enum Slip10Curve {
 
 function bytesToUnsignedBigInt(a: Uint8Array): bigint {
   return BigInt("0x" + toHex(a));
+}
+
+function intTo32be(n: bigint): Uint8Array {
+  // 32 bytes is 64 hexadecimal characters
+  const hex = n.toString(16).padStart(64, "0");
+  return fromHex(hex);
 }
 
 /**
@@ -174,9 +178,9 @@ export class Slip10 {
     }
 
     // step 5
-    const n = this.n(curve);
-    const returnChildKeyAsNumber = new BN(il).add(new BN(parentPrivkey)).mod(n);
-    const returnChildKey = Uint8Array.from(returnChildKeyAsNumber.toArray("be", 32));
+    const n: bigint = this.n(curve);
+    const returnChildKeyAsNumber = (bytesToUnsignedBigInt(il) + bytesToUnsignedBigInt(parentPrivkey)) % n;
+    const returnChildKey = intTo32be(returnChildKeyAsNumber);
 
     // step 6
     if (this.isGteN(curve, il) || this.isZero(returnChildKey)) {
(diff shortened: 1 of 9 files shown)


packages/crypto/src/slip10.ts
```diff
@@ -26,6 +24,12 @@ function bytesToUnsignedBigInt(a: Uint8Array): bigint {
   return BigInt("0x" + toHex(a));
 }
 
+function intTo32be(n: bigint): Uint8Array {
+  // 32 bytes is 64 hexadecimal characters
+  const hex = n.toString(16).padStart(64, "0");
+  return fromHex(hex);
+}
```
webmaster128, line 31:
```suggestion
function uintTo32be(n: bigint): Uint8Array {
  assert(n >= 0n);
  assert(n < BigInt(2**32));
  // 32 bytes is 64 hexadecimal characters
  const hex = n.toString(16).padStart(64, "0");
  return fromHex(hex);
}
```

The implementation is not correct for values outside of the uint32 range. So let's just assert that. (we need to import asser from @cosmjs/utils)

packages/crypto/src/slip10.ts
```diff
@@ -198,14 +202,14 @@ export class Slip10 {
   }
 
   private static isGteN(curve: Slip10Curve, privkey: Uint8Array): boolean {
-    const keyAsNumber = new BN(privkey);
-    return keyAsNumber.gte(this.n(curve));
+    const keyAsNumber: bigint = bytesToUnsignedBigInt(privkey);
```
webmaster128, line 205: Type can be inferred, right?
```suggestion
    const keyAsNumber = bytesToUnsignedBigInt(privkey);
```

packages/crypto/src/slip10.ts
```diff
@@ -174,9 +178,9 @@ export class Slip10 {
     }
 
     // step 5
-    const n = this.n(curve);
-    const returnChildKeyAsNumber = new BN(il).add(new BN(parentPrivkey)).mod(n);
-    const returnChildKey = Uint8Array.from(returnChildKeyAsNumber.toArray("be", 32));
+    const n: bigint = this.n(curve);
```
webmaster128, line 181:
```suggestion
    const n = this.n(curve);
```

packages/math/src/decimal.ts
```diff
@@ -94,7 +91,7 @@ export class Decimal {
 
   public static compare(a: Decimal, b: Decimal): number {
     if (a.fractionalDigits !== b.fractionalDigits) throw new Error("Fractional digits do not match");
-    return a.data.atomics.cmp(new BN(b.atomics));
+    return Math.sign(Number(a.data.atomics - b.data.atomics));
```
webmaster128, line 94: Can we implement this manually instead of using `Math.sign`, please? This function can cause all sorts of issues with -0 or NaN as return values. We don't need those cases as we have 2 ints as inputs.

packages/math/src/decimal.ts
```diff
@@ -118,7 +115,7 @@ export class Decimal {
     }
 
     this.data = {
-      atomics: new BN(atomics),
+      atomics: BigInt(atomics),
       fractionalDigits: fractionalDigits,
     };
   }
```
webmaster128, line 121: Slightly out of scope, but we can now also change the constuctor argument to `atomics: string | bigint` for more efficientcy in a few places.

packages/math/src/integers.ts
```diff
@@ -227,45 +228,55 @@ export class Uint64 implements Integer, WithByteConverters {
...
 
   public toBigInt(): bigint {
-    return BigInt(this.toString());
+    return this.data;
   }
 
   public toNumber(): number {
-    return this.data.toNumber();
+    const num = Number(this.data);
+    if (!Number.isSafeInteger(num)) {
```
webmaster128, line 276: Let's do this check before the conversion to be on the safe side of JS magic. Something along the lines of

```ts
if (this.data > BigInt(Number.MAX_SAFE_INTEGER)) {
  // ...
}
```

packages/crypto/src/slip10.ts
dynst: Was trying to make the code self-documenting but I guess 'BigInt' is right in the function name there.

packages/math/src/decimal.ts
webmaster128: I'm not aware of any inputs that can cause any issues. But the number type is ugly as a result of bigint math. E.g. we can have a diff value that cannot be converted to number without loss.

With sucht a simple 3liner we don't need to know how number and Math.sign behave together with bigint

```ts
if (a.data.atomics === b.data.atomics) return 0;
else if (a.data.atomics < b.data.atomics) return -1;
else return 1;
```

packages/crypto/src/slip10.ts
webmaster128: Oh, yeah sure you are right

packages/math/src/decimal.ts
dynst: Okay, used if statements for the sake of future code archaeologists who don't realize `Number(2n ** 970n - 2n ** 1024n)` clamps the number to `-Infinity`.

webmaster128:
Looks good to me, thank you!

We just need an entry in the CHANGELOG.md. I will read through it again tomorrow and then merge

Issue #584, opened 2020-12-16
Switch Ed25519 implementation to zip215 compliant library


webmaster128:
Right now we use Ed25519 from libsodium. [Tendermint migrates to zip215 ](https://github.com/tendermint/tendermint/pull/5632) and so should we in order to be able to verify Tendermint's signatures in the client.

There is the [ed25519-zebra](https://crates.io/crates/ed25519-zebra) crate, which we should be able to compile to Wasm and integrate into CosmJS.

dynst:
Luckily [`@noble/curves`](https://www.npmjs.com/package/@noble/curves) is ZIP215-compliant!

https://github.com/cosmos/cosmjs/blob/f27783e8d92982a2cebcdd8dfa710aecfc9d1032/packages/crypto/src/libsodium.ts#L2

PR #1722, opened 2025-07-20
migrate off libsodium

### LINKED ISSUE #1031: Find more lightweight libraries to replace libsodium.js
libsodium.js is amazing [but big](https://github.com/cosmos/cosmjs/issues/904). This could be mitigated by using our own build. However, I assume the better path is to break it down into the algorithms we need and have them in separate modules. This allows optimizers to better remove unused code. Also only Argon2 needs Wasm speed.

Replacements for the following algos are needed

- [ ] Argon2 (see https://github.com/cosmos/cosmjs/pull/907; looks surprisingly good now with the latest Rust compiler)
- [ ] Ed25519 (use different implementation, see https://github.com/cosmos/cosmjs/issues/584)
- [ ] xchacha20poly1305_ietf

### LINKED ISSUE #1478: Find strategy to remove libsodium being bundled
Currently libsodium is brought in for 2 imports:
* Argon2I
* Ed25519 key pairs

But neither of these code utilities are used by most apps. I cant even see usage of these within cosmJS. This is causing a large import to appear in our bundles. (190kb gzipped, 500kb after parsing)

I'm new to the JS ecosystem, so these are two options I see right now:
- Move the libsodium dependency into another package thats not imported everywhere
- Require libsodium wrappers to be an external dependency, and its up to the caller to include libsodium (https://webpack.js.org/configuration/externals/)

### LINKED ISSUE #1585: Replace most of libsodium with Noble cryptography
https://github.com/cosmos/cosmjs/blob/5c1ec56189d3024a6c6ad0ec0d60df30c09c7cc0/packages/crypto/src/index.ts#L5-L13

Ed25519 -> https://github.com/paulmillr/noble-ed25519
Argon2id -> https://github.com/paulmillr/noble-hashes#argon2

You already use libraries from the same author so I assume you're familiar with the work quality.

I'm suggesting this because libsodium is massive which makes it extremely hard to justify cosmjs unless the project is very large.

dynst:
Closes #1031
Closes #1478
Closes #1585

Related but not closed: #584


dynst:
This conflicts with #1720 but let's get that one merged first.

webmaster128:
Looks straight forward at first glance. However, I would be very surprised if the pure-JS implementation of Argon2 is as fast as the WebAssembly one. So this PR has the potential to hit existing users hard. Which then leads to the questions: who uses CosmJS pro private key encryption at all? Should we remove that functionality alltogether?

packages/crypto/src/libsodium.ts
```diff
@@ -115,34 +109,28 @@ export const xchacha20NonceLength = 24;
...
-      message,
-      additionalData,
-      null, // secret nonce: unused and should be null (https://download.libsodium.org/doc/secret-key_cryptography/aead/chacha20-poly1305/xchacha20-poly1305_construction)
-      nonce,
-      key,
-    );
+    let ciphertext = new Uint8Array(message.length + 16);
+    ciphertext = cipher.encrypt(message, ciphertext);
+
+    return ciphertext;
```
webmaster128, line 119: I think we don't need to pass an output object here. It is optional and will be created internally if undefined.

```suggestion
    return cipher.encrypt(message);
```

packages/crypto/src/libsodium.ts
```diff
@@ -115,34 +109,28 @@ export const xchacha20NonceLength = 24;
...
     nonce: Uint8Array,
   ): Promise<Uint8Array> {
-    await sodium.ready;
+    const additionalAuthenticatedData = undefined;
+
+    const cipher = xchacha20poly1305(key, nonce, additionalAuthenticatedData);
 
-    const additionalData = null;
+    let plaintext = new Uint8Array(ciphertext.length - 16);
+    plaintext = cipher.decrypt(ciphertext, plaintext);
```
webmaster128, line 132: same here

dynst:
The basic parameters run so slowly the unit tests fail to finish in 15s. Or 60s. Or even 180s.

https://github.com/cosmos/cosmjs/blob/1264c2c90c15b6065292be89fb6a5bcd4c49060e/packages/amino/src/secp256k1hdwallet.ts#L37-L48

```
[@cosmjs/amino]: **************************************************
[@cosmjs/amino]: *                    Failures                    *
[@cosmjs/amino]: **************************************************
[@cosmjs/amino]: 
[@cosmjs/amino]: 1) Secp256k1HdWallet deserialize can restore
[@cosmjs/amino]:   - Error: Timeout - Async function did not complete within 15000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)
```

So just to keep CI happy `hash-wasm` looks necessary actually.

packages/proto-signing/src/wallet.ts
```diff
@@ -28,7 +28,13 @@ export async function executeKdf(password: string, configuration: KdfConfigurati
     case "argon2id": {
       const options = configuration.params;
       if (!isArgon2idOptions(options)) throw new Error("Invalid format of argon2id params");
-      return Argon2id.execute(password, cosmjsSalt, options);
+
+      // Emulate a slower implementation. The fast WASM code may get removed.
+      // This approximates the speed of using a pure JS implementation (@noble/hashes) in Node 22.
+      const screamTest = new Promise((resolve) => setTimeout(resolve, options.opsLimit * 250));
```
webmaster128, line 34: 💅 we can use `sleep` from `@cosmjs/utils` which does exactly that but is a bit more readable

webmaster128 (approved):

PR #1729, opened 2025-07-23
Modernize child_process import in reconnecting socket test

### LINKED ISSUE #1724: removing require() calls
Ignoring the .js scripts, there are 3 call sites in TypeScript:

https://github.com/cosmos/cosmjs/blob/d708b6aab562a6242d07b821d8f570e9eb8a5413/packages/faucet/src/api/webserver.ts#L3

https://github.com/cosmos/cosmjs/blob/d708b6aab562a6242d07b821d8f570e9eb8a5413/packages/crypto/src/random.ts#L17-L20

https://github.com/cosmos/cosmjs/blob/d708b6aab562a6242d07b821d8f570e9eb8a5413/packages/socket/src/reconnectingsocket.spec.ts#L11-L13

Are these absolutely needed?

dynst:
One-third of #1724. Closes #1724

Looks like this test code was inherited from #246

https://github.com/webpack/webpack/releases/v5.92.0

webpack added a feature that adds `node:` prefixes "in runtime code," but still compile-time rejects it if it's already present in the code, and the target is a browser. Even if it's a dynamic import, which can fail.

### DIFF
--- a/packages/socket/src/reconnectingsocket.spec.ts
+++ b/packages/socket/src/reconnectingsocket.spec.ts
@@ -3,7 +3,7 @@ import assert from "assert";
 import { ReconnectingSocket } from "./reconnectingsocket";
 
 /** @see https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback */
-type Exec = (command: string, callback: (error: null | (Error & { readonly code: number })) => void) => void;
+type Exec = (command: string, callback: (error: null | (Error & { readonly code?: number })) => void) => void;
 
 let exec: Exec | undefined;
 let childProcessAvailable: boolean;
--- a/packages/socket/src/reconnectingsocket.spec.ts
+++ b/packages/socket/src/reconnectingsocket.spec.ts
@@ -97,13 +97,19 @@ describe("ReconnectingSocket", () => {
     const startServerCmd = `${dirPath}/start.sh`;
     const stopServerCmd = `${dirPath}/stop.sh`;
 
-    it("automatically reconnects if no connection can be established at init", (done) => {
+    it("automatically reconnects if no connection can be established at init", async () => {
+      let pass!: () => void, fail!: (reason?: any) => void;
+      const ret = new Promise<void>((resolve, reject) => {
+        pass = resolve;
+        fail = reject;
+      });
+
       pendingWithoutChildProcess();
       pendingWithoutSocketServer();
 
       exec!(stopServerCmd, (stopError) => {
         if (stopError && stopError.code !== codePkillNoProcessesMatched) {
-          done.fail(stopError);
+          fail(stopError);
         }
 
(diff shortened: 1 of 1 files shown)


packages/socket/src/reconnectingsocket.spec.ts
```diff
@@ -97,13 +85,26 @@ describe("ReconnectingSocket", () => {
...
     const stopServerCmd = `${dirPath}/stop.sh`;
 
-    it("automatically reconnects if no connection can be established at init", (done) => {
-      pendingWithoutChildProcess();
+    it("automatically reconnects if no connection can be established at init", async () => {
+      let pass!: () => void, fail!: (reason?: any) => void;
+      const ret = new Promise<void>((resolve, reject) => {
+        pass = resolve;
+        fail = reject;
+      });
```
webmaster128, line 93: I think it is a bit less boilerplate and more natural to use `(done) =>` with an inner unnamed async function. See e.g.

```ts
it("can subscribe to block header events", (done) => {
    pendingWithoutTendermint();

    const testStart = ReadonlyDate.now();

    (async () => {
      const events: responses.NewBlockHeaderEvent[] = [];
      const client = Tendermint37Client.create(rpcFactory());
      const stream = client.subscribeNewBlockHeader();
      expect(stream).toBeTruthy();
      // ...
``` 

in this codebase

packages/socket/src/reconnectingsocket.spec.ts
```diff
@@ -1,34 +1,22 @@
...
-  exec = require("child_process").exec;
-  assert.strict(typeof exec === "function");
-  childProcessAvailable = true;
-} catch {
-  childProcessAvailable = false;
-}
+type Exec = (command: string, callback: (error: null | (Error & { readonly code?: number })) => void) => void;
+
+const getExec = async (): Promise<Exec> => {
+  const exec = (await import("child_process")).exec;
```
webmaster128, line 7: We could catch the exception here and then just return `Exec | undefined`. This would probably lead to cleaner caller code

webmaster128 (commented):
Nice. Could you update to latest main and remove the dependency changes from the PR? Then we can get this in and close #1724 – great stuff

packages/socket/src/reconnectingsocket.spec.ts
dynst: Ideally all this code would be refactored to be Promise-y instead of callback-y to avoid this boilerplate kludge that combines the 2 styles, do you not think this is a step in the right direction and a reminder to more fully rewrite the test later?

webmaster128: No super strong opinion. I was just remembering that we had those cases before and here it looks like a lot of boilerplate. But at the end of the day I am happy either way

PR #1735, opened 2025-07-24
Use native fetch directly

### LINKED ISSUE #1734: cross-fetch issues
`cross-fetch` still depends on an old version of `node-fetch` (because 3.0 is ESM-only, but that isn't a problem if you drop node 18 support), and that bites some users.

https://github.com/lquixada/cross-fetch/issues/177

https://github.com/Agoric/agoric-sdk/issues/11638

https://github.com/cosmos/cosmjs/issues/1627#issuecomment-3108797769

dynst:
Update from using the cross-fetch conditional polyfill to unconditional native fetch.

Closes #1734

### DIFF
--- a/.pnp.cjs
+++ b/.pnp.cjs
@@ -1062,10 +1062,6 @@ const RAW_RUNTIME_STATE =
       "npm:0.9.0"\
     ],\
     [\
-      "cross-fetch",\
-      "npm:4.1.0"\
-    ],\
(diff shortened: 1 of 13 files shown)


webmaster128:
I accidentally marked this as ready for review. Are we happy with the diff? Looks good to me.

Could you add a CHANGELOG.md entry to this PR?

PR #1742, opened 2025-07-25
chore: fix some minor issues in comments

houpo-bob:
fix some minor issues in comments

### DIFF
--- a/packages/stargate/src/modules/ibc/aminomessages.ts
+++ b/packages/stargate/src/modules/ibc/aminomessages.ts
@@ -27,7 +27,7 @@ export interface AminoMsgTransfer extends AminoMsg {
     /**
      * The timeout as a (revision_number, revision_height) pair.
      *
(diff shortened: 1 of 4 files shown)


packages/stargate/src/modules/ibc/aminomessages.ts
```diff
@@ -27,7 +27,7 @@ export interface AminoMsgTransfer extends AminoMsg {
     /**
      * The timeout as a (revision_number, revision_height) pair.
      *
-     * This fied is is non-optional (https://github.com/cosmos/cosmos-sdk/blob/v0.42.7/x/ibc/applications/transfer/types/tx.pb.go#L49).
+     * This fied is non-optional (https://github.com/cosmos/cosmos-sdk/blob/v0.42.7/x/ibc/applications/transfer/types/tx.pb.go#L49).
```
dynst, line 30:
```suggestion
     * This field is non-optional (https://github.com/cosmos/cosmos-sdk/blob/v0.42.7/x/ibc/applications/transfer/types/tx.pb.go#L49).
```

PR #1747, opened 2025-07-29
chore: Fix typos in crypto and json-rpc modules

VolodymyrBg:
In pbkdf2.ts:

- interprete → interpret
- crypro → crypto

In dummyservice.worker.ts:

- requst → request


webmaster128:
Thanks! The code in `pbkdf2.ts` was removed in the maintime, so only the other fix is left. I just updated the PR and will merge when green.

PR #1753, opened 2025-07-30
crypto: eliminate old pbkdf2 fallbacks

### LINKED ISSUE #1341: Remove pbkdf2Sha512Crypto
We have pbkdf2Sha512Subtle (modern browsers and Node.js) and pbkdf2Sha512Noble (pure-JS implementation) available. pbkdf2Sha512Crypto is only needed for cases where there is no subtle and we still want a fast implementation. However, given the pain with bundlers that comes from importing Node modules, it would be valuable to remove this.

This was brought up by @mvid in https://github.com/cosmos/cosmjs/pull/1307 for React Native environments.

### LINKED ISSUE #1550: Error: Can't resolve 'crypto'
I'm trying to install this library in my React project, but I'm receiving the error below when I run `yarn start`:

```
ERROR in ./node_modules/@cosmjs/crypto/build/pbkdf2.js 50:67-84
Module not found: Error: Can't resolve 'crypto' in '***\node_modules\@cosmjs\crypto\build'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "crypto": require.resolve("crypto-browserify") }'
        - install 'crypto-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "crypto": false }
```

Also I tried to setup webpack, but did not work.

webpack.config.js

```js
module.exports = [
  {
    plugins: [],
    resolve: {
      fallback: {
        buffer: false,
        crypto: false,
        events: false,
        path: false,
        stream: false,
        string_decoder: false
      }
    }
  }
]
```

Version:

"@cosmjs/stargate": "^0.32.2"

### LINKED ISSUE #1583: React Native Expo build error
Hello everyone
I am getting above error

```Error: Unable to resolve module crypto from /home/expo/workingdir/build/node_modules/@cosmjs/proto-signing/node_modules/@cosmjs/crypto/build/pbkdf2.js: crypto could not be found within the project or in these directories:
node_modules/@cosmjs/proto-signing/node_modules
node_modules
  33 | async function getCryptoModule() {
  34 |     try {
> 35 |         const crypto = await Promise.resolve().then(() => __importStar(require("crypto")));
    |                                                                                 ^
  36 |         // We get `Object{default: Object{}}` as a fallback when using
  37 |         // `crypto: false` in Webpack 5, which we interprete as unavailable.
  38 |         if (typeof crypto === "object" && Object.keys(crypto).length <= 1) {
Error: Unable to resolve module crypto from /home/expo/workingdir/build/node_modules/@cosmjs/proto-signing/node_modules/@cosmjs/crypto/build/pbkdf2.js: crypto could not be found within the project or in these directories:
  node_modules/@cosmjs/proto-signing/node_modules
  node_modules
  33 | async function getCryptoModule() {
  34 |     try {
> 35 |         const crypto = await Promise.resolve().then(() => __importStar(require("crypto")));
     |                                                                                 ^
  36 |         // We get `Object{default: Object{}}` as a fallback when using
  37 |         // `crypto: false` in Webpack 5, which we interprete as unavailable.
  38 |         if (typeof crypto === "object" && Object.keys(crypto).length <= 1) {
    at ModuleResolver.resolveDependency (/home/expo/workingdir/build/node_modules/metro/src/node-haste/DependencyGraph/ModuleResolution.js:138:15)
    at DependencyGraph.resolveDependency (/home/expo/workingdir/build/node_modules/metro/src/node-haste/DependencyGraph.js:231:43)
    at /home/expo/workingdir/build/node_modules/metro/src/lib/transformHelpers.js:156:21
    at resolveDependencies (/home/expo/workingdir/build/node_modules/metro/src/DeltaBundler/buildSubgraph.js:42:25)
    at visit (/home/expo/workingdir/build/node_modules/metro/src/DeltaBundler/buildSubgraph.js:83:30)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Promise.all (index 4)
at async visit (/home/expo/workingdir/build/node_modules/metro/src/DeltaBundler/buildSubgraph.js:92:5)
    at async Promise.all (index 0)
    at async visit (/home/expo/workingdir/build/node_modules/metro/src/DeltaBundler/buildSubgraph.js:92:5)```

dynst:
Follow-up to #1340 #1342 #1355

Re-closes #1341. Bundlers still give annoying messages about using a native Node.js module.

Closes #1550, closes #1583


webmaster128:
What about keeping subtle + noble (pure JS fallback)? This would remove the problematic node.js specific import and keep the solution flexible enough for environments that don't have subtle (whatever that might be).

webmaster128:
> Weird CI failure https://github.com/cosmos/cosmjs/actions/runs/16615628039/job/47007535517

Those kind of errors pop up when the blockchain does during the test run. Then the client cannot connect anymore. This happens occasionally and just disappears with a re-run.

PR #1755, opened 2025-07-30
Add packages/*/dist/ to eslint ignore patterns

webmaster128:
In https://github.com/cosmos/cosmjs/pull/1705 we missed the dist directory which is created in browser test builds (by webpack). If you have one of those dist folders in a local checkout, the lint will take forever.


webmaster128:
@dynst fyi

Issue #1756, opened 2025-07-30
eslint9


webmaster128:
I just gave it a try using the config migration tooling. But the result is very convoluted and linting does not work due to all sorts of missing dependencies and config issues. I think it would be far easier to create a fresh eslint9 config setup from scratch and then copying over our rules.

See https://typescript-eslint.io/getting-started

dynst:
Isn't this simply gated on updating the incompatible `eslint-plugin-import`, new versions of which complain about #1667

webmaster128:
If that becomes a blocker for the eslint9 upgrade, I'd just disable `import/no-cycle`.

webmaster128:
@dynst thanks a lot for all the flat config changed. This makes the eslint9 upgrade super easy (#1782)

PR #1757, opened 2025-07-30
flat config

dynst:
Update `eslint-plugin-import`, which only got flat config support in June with v2.32.

https://github.com/import-js/eslint-plugin-import/releases/tag/v2.32.0

New dependencies added:

```sh
yarn add @eslint/compat @eslint/js@^8.57.1 @eslint/eslintrc globals espree --dev
```

But the output of running `yarn dlx @eslint/migrate-config@1.3.8 .eslintrc.js` is riddled with errors that needed manual fixing (it generates a default import, it has to be changed to a namespace import `import * as espree` - zero difference if this were CJS, but it's ESM so it's two different things).

Part of #1756

<del>I'm not sure how these errors were silenced before, but something about how the new flat config file is setup has enabled a bunch of lints: `no-dupe-class-members` `no-redeclare` `no-undef`.</del> That's fixed now!


eslint.config.mjs
```diff
@@ -0,0 +1,178 @@
...
+    },
+
+    languageOptions: {
+      globals: {
+        ...globals.jasmine,
+        ...globals.node,
+        ...globals.worker,
+      },
+
+      parser: espree,
```
webmaster128, line 31: As espree is the default parser (https://eslint.org/docs/v8.x/use/configure/parser) I think we can just remove this setting and remove the explicit `espree` + `@types/espree` dependency, right?

eslint.config.mjs
```diff
@@ -0,0 +1,178 @@
...
+        "always",
+        {
+          line: {
+            markers: ["/ <reference"],
+          },
+        },
+      ],
+
+      "import/no-cycle": "off",
+      "import/no-named-as-default-member": "off",
```
webmaster128, line 63: After #1760 we can reset this rule to the default setting

webmaster128 (commented):
Very nice stuff

PR #1758, opened 2025-07-30
add exports field to all packages

dynst:
The `"main"` field has been replaceable by `"exports"` since Node 12.

https://nodejs.org/api/packages.html#community-conditions-definitions

Node 20 is `nodenext` as of TypeScript 5.8. (TypeScript 5.9 will add an explicit `node20` setting.)

https://www.typescriptlang.org/docs/handbook/modules/reference.html#the-module-compiler-option

Needs #1759 for `"exports"` support from the eslint plugin

Needs #1776 to configure webpack to work with `await import()` in a unit test.


webmaster128:
I pulled out the last commit into #1819. I guess this is the essence of this PR, right? Are you fine with merging #1819 instead of this?

PR #1767, opened 2025-08-01
enable strict lints

dynst:
[`no-non-null-assertion`](https://typescript-eslint.io/rules/no-non-null-assertion/) is disabled in many places, but it wasn't actually enabled anywhere. Now it is, as part of `strict-type-checked`.

The [`explicit-module-boundary-types`](https://typescript-eslint.io/rules/explicit-module-boundary-types/) lint isn't part of `strict-type-checked`, but it's also explicitly disabled in a bunch of places https://github.com/cosmos/cosmjs/commit/11ec6f1393534b68e9abbe07b4bce7d48edfdc15 and wasn't actually enabled anywhere.

Maybe they were enabled in an older version of the eslint plugin before the updates.

### DIFF
--- a/packages/crypto/src/sha.ts
+++ b/packages/crypto/src/sha.ts
@@ -1,5 +1,4 @@
-import { sha256 as nobleSha256 } from "@noble/hashes/sha256";
-import { sha512 as nobleSha512 } from "@noble/hashes/sha512";
+import { sha256 as nobleSha256, sha512 as nobleSha512 } from "@noble/hashes/sha2.js";
 
 import { HashFunction } from "./hash";
 import { toRealUint8Array } from "./utils";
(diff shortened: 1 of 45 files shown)


webmaster128 (approved):
👏

packages/crypto/src/sha.ts
```diff
@@ -1,5 +1,4 @@
-import { sha256 as nobleSha256 } from "@noble/hashes/sha256";
-import { sha512 as nobleSha512 } from "@noble/hashes/sha512";
+import { sha256 as nobleSha256, sha512 as nobleSha512 } from "@noble/hashes/sha2.js";
```
webmaster128, line 1: Those changes seem to be causing issues for bundlers as in @noble/hashes version 1.x we must not use the file extension. See

https://github.com/paulmillr/noble-hashes/blob/1.8.0/package.json#L126

A user is getting

> Error
  [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './sha2.js' is not defined
  by "exports" in …

I will send a PR to fix this in a bit

Issue #1004, opened 2022-01-23
Compile to ES modules


webmaster128:
Right now, CosmJS produces CommonJS builds. This works well for all versions of Node.js as well as bundlers like Webpack in order to bundle for the web.

As mentioned in https://vuejsdevelopers.com/2019/02/04/vue-es-module-browser-build/ ("What benefit is there of using Vue as an ES module?"), there is little obvious benefit for switching from CommonJS to ESM since the consumers are either node.js apps or bundlers. However, there is hope that ESM helps optimizing bundle size through tree shaking because ESM can better do static analysis on what is used and what is not (https://webpack.js.org/guides/tree-shaking/). In #904 you see that right now web apps include a lot of unused code that is only needed for key storage.

Other libraries use bundlers to bundle ES modules for the web. This is not desired here as we don't have a single entry point and do not want to decide for that user what should be included in such a bundle.

**Do we need CommonJS+ESM hybrid builds?**

- Node.js 12 does not support ESM natively without feature flags. But this version reaches EOL at 2022-04-30.
- Webpack support ESM inputs since version 2 (["The webpack 2 release came with built-in support for ES2015 modules (alias harmony modules)"](https://webpack.js.org/guides/tree-shaking/))

**Does this reduce bundle size**

- [ ] Test in a next.js project with [the analyze plugin](https://github.com/cosmos/cosmjs/issues/904#issuecomment-1018681955)

**Testing**

- [x] **Jasmine supports ESM** build testing (see https://github.com/jasmine/jasmine-npm/issues/170)
- [ ] **JSON test vectors cannot be used easily in tests**. `const bip39Vectors = require("./src/testdata/bip39.json");` works and TS transpiles `import bip39Vectors from "./testdata/bip39.json";` to `const bip39_json_1 = __importDefault(require("./testdata/bip39.json"))`. But an untranspiled `import bip39Vectors from "./testdata/bip39.json";` output is not working. See https://nodejs.org/api/esm.html#no-json-module-loading and https://nodejs.org/api/esm.html#json-modules. A workaround for tests is:
  ```ts
  import { createRequire } from "module";
  const require = createRequire(import.meta.url);
  const bip39Vectors = require("../src/testdata/bip39.json");
  ```
  but this does not work in Webpack/Karma.
[...]

PR #1785, opened 2025-08-07
import cosmjs-types with full file path

dynst:
```sh
sed -i -E 's/(from "cosmjs-types.+)";$/\1.js";/' `grep cosmjs-types/ -rlI packages/`
```

Omitting a file extension only works when transpiled to CJS.

Prerequisite for #1004


dynst:
Was there something blocking this?

webmaster128:
I did not look into / think about this topic yet. I think this answer is helpful to understand why this change is needed: https://stackoverflow.com/questions/63459159/omit-the-file-extension-es6-module-nodejs

To me it feels ugly to have to use cosmjs-types like this, especially as we are not only importing pure-JS code but also its types. Maybe we should auto-generate the paths for `exports` there and then map the path to the .js file.

PR #1786, opened 2025-08-07
simplify eslint flat config

dynst:
And enable import lints on JS files, though unclear why one particular import is triggering a lint.


webmaster128 (commented):
If we remove `@typescript-eslint/parser` and `@eslint/compat` imports here, we can remove the dependencies from package.json too right?

webmaster128:
Sorry for the delay. I think this is good to get merged once the conflicts are resolved

PR #1794, opened 2025-08-12
remove readonly tags from json-rpc types

dynst:
If they should be readonly in certain places, those places should use `Readonly<>`.

`JsonCompatibleDictionary` is now interchangeable with `Record<string, JsonValue>`, which could now be used in a lot more places to remove the use of `any`.


webmaster128:
Interesting, looks good at first glance

> I'm guessing migrating to using real json validation like `zod` would be overkill.

Yeah, agreed. Also I don't like external libraries in our public types which makes it very hard to replace dependencies.

dynst:
> external libraries in our public types which makes it very hard to replace dependencies.

And even just upgrading them, zod had a whole thing about how carefully the upgrade had to be designed due to how entangled it is with APIs.

https://github.com/colinhacks/zod/issues/4371
https://zod.dev/v4

PR #1813, opened 2025-08-25
drop buffer polyfill from crypto

dynst:
Buffer hasn't been used since #960
https://github.com/cosmos/cosmjs/pull/960/commits/4a6253f16bf1c7bb1a0bbd68d331ab712165306f

### DIFF
--- a/.pnp.cjs
+++ b/.pnp.cjs
@@ -3513,7 +3513,6 @@ const RAW_RUNTIME_STATE =
           ["@types/karma-jasmine-html-reporter", "npm:1.5.1"],\
           ["@types/libsodium-wrappers-sumo", "npm:0.7.5"],\
           ["@types/node", "npm:22.10.6"],\
-          ["buffer", "npm:6.0.3"],\
           ["glob", "npm:11.0.3"],\
           ["hash-wasm", "npm:4.12.0"],\
           ["jasmine", "npm:4.6.0"],\
--- a/packages/crypto/package.json
+++ b/packages/crypto/package.json
@@ -53,7 +53,6 @@
     "@types/karma-jasmine-html-reporter": "^1",
     "@types/libsodium-wrappers-sumo": "^0.7.5",
     "@types/node": "*",
-    "buffer": "^6.0.3",
     "glob": "^11",
     "jasmine": "^4",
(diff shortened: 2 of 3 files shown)


webmaster128:
Thanks! The `"buffer"` import is not used directly anmore indeed. But this is added as a `devDependency` also because of

```ts
      new webpack.ProvidePlugin({
        Buffer: ["buffer", "Buffer"],
      }),
```

in the webpack config. But then again the combination with

```
    resolve: {
      fallback: {
        buffer: false,
```

does not make much sense.

So if I understand things correctly we should remove the Buffer plugin whenever we remive the `buffer` `devDependency`.

dynst:
> The `"buffer"` import is not used directly anymore indeed.

Not quite, it was never imported at all, the `Buffer` global was being used by the ripemd160 code in the linked commit, it only existed for the webpack build, yeah.

> if I understand things correctly we should remove the Buffer plugin whenever we remive the `buffer` `devDependency`.


But every webpack config file includes the Buffer plugin. I suspect they have to, because lower-level packages use the global, and the plugin provides the `Buffer` global.

https://webpack.js.org/plugins/provide-plugin/

`resolve.fallback.buffer = false` is a separate thing that, in webpack 4, turns off the polyfill of importing the `'buffer'` package, but doesn't affect the global. webpack is pointlessly confusing like that. https://webpack.js.org/configuration/resolve/#resolvefallback

It isn't actually needed anymore in webpack 5 though, where the default is `false` anyway.

> ```
> BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
> This is no longer the case. Verify if you need this module and configure a polyfill for it.```

webmaster128:
Thanks for all the explanations. This makes a lot of sense.

Coming back to the very specific case in this PR without increasing scope. Once we remove the `buffer` devDependency from @cosmjs/crypto, this line does not work anymore:
https://github.com/cosmos/cosmjs/blob/v0.36.1/packages/crypto/webpack.web.config.cjs#L20
Or it works by accident because some other package brings in the `buffer` dependency and can break any time later.

So if we remove the dependency, we also need to remove the Buffer plugin from the package. Not sure if this can be done without affecting any lower level dependency but probably worth a try.

PR #1814, opened 2025-08-27
Feat/support cosmos evm

chiulam:
### Summary
This PR adds support for Ethereum-style secp256k1 addresses in CosmJS, enabling compatibility with Ethereum-derived address formats in Cosmos chains.

### Changes
- Added `rawEthSecp256k1PubkeyToRawAddress()` function to convert compressed Ethereum secp256k1 public keys to raw addresses
- Updated `pubkeyToRawAddress()` to handle `EthSecp256k1Pubkey` type
- Added support for keccak256 hashing for Ethereum address derivation
- Imported necessary dependencies (`keccak256`, `Secp256k1`) from `@cosmjs/crypto`

### Technical Details
- Ethereum addresses are derived by taking the keccak256 hash of the uncompressed public key (without the 0x04 prefix) and using the last 20 bytes
- The implementation follows Ethereum's address derivation standard while maintaining compatibility with existing Cosmos address formats
- Added proper validation for compressed public key length (33 bytes)

### Breaking Changes
None - this is a purely additive change that maintains backward compatibility.


packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
```diff
@@ -703,7 +704,12 @@ export class SigningCosmWasmClient extends CosmWasmClient {
     if (!accountFromSigner) {
       throw new Error("Failed to retrieve account from signer");
     }
-    const pubkey = encodePubkey(encodeSecp256k1Pubkey(accountFromSigner.pubkey));
+    let pubkey;
+    if (accountFromSigner.algo == "eth_secp256k1" || accountFromSigner.algo == "ethsecp256k1" ) {
```
BigtoC, line 708: It would be good to add a comment here to explain why you need `eth_secp256k1` and `ethsecp256k1`.

packages/stargate/src/signingstargateclient.ts
```diff
@@ -397,7 +397,12 @@ export class SigningStargateClient extends StargateClient {
     if (!accountFromSigner) {
       throw new Error("Failed to retrieve account from signer");
     }
-    const pubkey = encodePubkey(encodeSecp256k1Pubkey(accountFromSigner.pubkey));
+    let pubkey;
+    if (accountFromSigner.algo == "eth_secp256k1" || accountFromSigner.algo == "ethsecp256k1") {
```
BigtoC, line 401: It would be good to add a comment here to explain why you need `eth_secp256k1` and `ethsecp256k1`.

BigtoC (approved):

packages/amino/src/pubkeys.ts
```diff
@@ -24,10 +24,21 @@ export function isSecp256k1Pubkey(pubkey: Pubkey): pubkey is Secp256k1Pubkey {
   return (pubkey as Secp256k1Pubkey).type === "tendermint/PubKeySecp256k1";
 }
 
+export interface EthSecp256k1Pubkey extends SinglePubkey {
+  readonly type: "os/PubKeyEthSecp256k1";
+  readonly value: string;
+}
+
+export function isEthSecp256k1Pubkey(pubkey: Pubkey): pubkey is EthSecp256k1Pubkey {
+  return (pubkey as EthSecp256k1Pubkey).type === "os/PubKeyEthSecp256k1";
```
Cordtus, line 33: Is this a typo? `os/PubKeyEthSecp256k1`

packages/amino/src/signerutils.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Gets the correctly encoded amino pubkey for an account based on its algorithm.
+ * 
+ * This utility automatically selects the appropriate encoding function based on whether
+ * the account uses Ethereum secp256k1 keys or standard secp256k1 keys.
+ *
+ * @param account The account data from a signer
+ * @returns The amino-encoded pubkey (EthSecp256k1Pubkey or Secp256k1Pubkey)
+ */
+export function getAminoPubkey(account: AccountData): any {
```
dynst, line 33: There's no good reason for the return type here to be `any`.

dynst (changes requested):
This function isn't updated:

https://github.com/cosmos/cosmjs/blob/67234a84d92bda47709784661f69a3a527311289/packages/amino/src/pubkeys.ts#L55-L58

BigtoC:
> This function isn't updated:
> 
> https://github.com/cosmos/cosmjs/blob/67234a84d92bda47709784661f69a3a527311289/packages/amino/src/pubkeys.ts#L55-L58

I have updated the code. Please have a look.

packages/amino/src/addresses.ts
```diff
@@ -20,11 +20,24 @@ export function rawSecp256k1PubkeyToRawAddress(pubkeyData: Uint8Array): Uint8Arr
   return ripemd160(sha256(pubkeyData));
 }
 
+export function rawEthSecp256k1PubkeyToRawAddress(pubkeyData: Uint8Array): Uint8Array {
```
webmaster128, line 23: ~An "raw Eth Secp256k1 Pubkey" is uncompressed by convention. All Ethereum ecosystem uses uncompressed pubkeys. So either the implementation is wrong but maybe it is a question of findung a better name for the function~

I am wrong I guess and compressed pubkeys are used consostently in Cosmos, also evm code. Any thoughts on that matter?

packages/amino/src/pubkeys.ts
webmaster128: Thanks.

What does "os/" mean in this context?

packages/amino/src/pubkeys.ts
```diff
@@ -24,10 +24,21 @@ export function isSecp256k1Pubkey(pubkey: Pubkey): pubkey is Secp256k1Pubkey {
   return (pubkey as Secp256k1Pubkey).type === "tendermint/PubKeySecp256k1";
 }
 
+export interface EthSecp256k1Pubkey extends SinglePubkey {
+  readonly type: "os/PubKeyEthSecp256k1";
+  readonly value: string;
```
webmaster128, line 29: Question: is this a compressed or uncompressed encoding of the pubkey? In earlier versions of Ethermint this was uncompressed (but barely documented). Now looking at the code above it seems to be compressed.

packages/amino/src/pubkeys.ts
```diff
@@ -24,10 +24,21 @@ export function isSecp256k1Pubkey(pubkey: Pubkey): pubkey is Secp256k1Pubkey {
...
+export function isEthSecp256k1Pubkey(pubkey: Pubkey): pubkey is EthSecp256k1Pubkey {
+  return (pubkey as EthSecp256k1Pubkey).type === "os/PubKeyEthSecp256k1";
+}
+
 export const pubkeyType = {
-  /** @see https://github.com/tendermint/tendermint/blob/v0.33.0/crypto/ed25519/ed25519.go#L22 */
-  secp256k1: "tendermint/PubKeySecp256k1" as const,
   /** @see https://github.com/tendermint/tendermint/blob/v0.33.0/crypto/secp256k1/secp256k1.go#L23 */
+  secp256k1: "tendermint/PubKeySecp256k1" as const,
+  /** @see https://github.com/cosmos/evm/blob/main/crypto/ethsecp256k1/ethsecp256k1.go#L36 */
```
webmaster128, line 39:
```suggestion
  /** @see https://github.com/cosmos/evm/blob/v1.0.0-rc2/crypto/ethsecp256k1/ethsecp256k1.go#L35-L36 */
```

packages/amino/src/pubkeys.ts
```diff
@@ -53,7 +64,7 @@ export interface SinglePubkey extends Pubkey {
 }
 
 export function isSinglePubkey(pubkey: Pubkey): pubkey is SinglePubkey {
-  const singPubkeyTypes: string[] = [pubkeyType.ed25519, pubkeyType.secp256k1, pubkeyType.sr25519];
+  const singPubkeyTypes: string[] = [pubkeyType.ed25519, pubkeyType.secp256k1, pubkeyType.ethsecp256k1, pubkeyType.sr25519];
```
webmaster128, line 67: Could you add a unit test for this addition?

packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
```diff
@@ -1,4 +1,5 @@
-import { encodeSecp256k1Pubkey, makeSignDoc as makeSignDocAmino } from "@cosmjs/amino";
+/* eslint-disable @typescript-eslint/naming-convention */
```
webmaster128, line 1: Is this disable really needed?

packages/proto-signing/src/directethsecp256k1hdwallet.ts
```diff
@@ -0,0 +1,369 @@
...
+    const encryptedData = await encrypt(dataToEncryptRaw, encryptionKey, encryptionConfiguration);
+
+    const out: DirectSecp256k1HdWalletSerialization = {
+      type: serializationTypeV1,
+      kdf: kdfConfiguration,
+      encryption: encryptionConfiguration,
+      data: toBase64(encryptedData),
+    };
+    return JSON.stringify(out);
+  }
```
webmaster128, line 344: Could we please remove all references to kdf/serialize/deserialize from the newly added wallet? I know it is there for consistency but we will remove that functionality sooninsh and don't want new users to start with it. See https://github.com/cosmos/cosmjs/issues/1796

packages/proto-signing/src/directethsecp256k1wallet.ts
```diff
@@ -0,0 +1,64 @@
...
+import { toBech32 } from "@cosmjs/encoding";
+import { SignDoc } from "cosmjs-types/cosmos/tx/v1beta1/tx";
+
+import { AccountData, DirectSignResponse, OfflineDirectSigner } from "./signer";
+import { makeSignBytes } from "./signing";
+
+/**
+ * A wallet that holds a single secp256k1 keypair.
+ *
+ * If you want to work with BIP39 mnemonics and multiple accounts, use DirectSecp256k1HdWallet.
```
webmaster128, line 12:
```suggestion
 * If you want to work with BIP39 mnemonics and multiple accounts, use DirectEthSecp256k1HdWallet.
```

packages/proto-signing/src/pubkey.ts
```diff
@@ -30,6 +32,14 @@ export function encodePubkey(pubkey: Pubkey): Any {
       typeUrl: "/cosmos.crypto.secp256k1.PubKey",
       value: Uint8Array.from(CosmosCryptoSecp256k1Pubkey.encode(pubkeyProto).finish()),
     });
+  } else if (isEthSecp256k1Pubkey(pubkey)) {
+    const pubkeyProto = CosmosCryptoSecp256k1Pubkey.fromPartial({
```
webmaster128, line 36:
```suggestion
    // Note: This code block is hacky because we should use the correct EVM proto type
    // https://github.com/cosmos/evm/blob/v1.0.0-rc2/proto/cosmos/evm/crypto/v1/ethsecp256k1/keys.proto#L12-L17.
    // However, we do not have that available here or in cosmjs-types yet and the classic secp256k1 pubkey has the same structure except for documentation and annotations:
    // https://github.com/cosmos/cosmos-sdk/blob/v0.53.4/proto/cosmos/crypto/secp256k1/keys.proto#L14-L30
    //
    // Actually this type is so simple we should just have an Anybuf for TS instead of code generation (https://github.com/webmaster128/anybuf)

    const pubkeyProto = CosmosCryptoSecp256k1Pubkey.fromPartial({
```

webmaster128: This just adds a comment, no code change

packages/proto-signing/src/signer.ts
```diff
@@ -1,7 +1,7 @@
 import { OfflineAminoSigner, StdSignature } from "@cosmjs/amino";
 import { SignDoc } from "cosmjs-types/cosmos/tx/v1beta1/tx";
 
-export type Algo = "secp256k1" | "ed25519" | "sr25519";
+export type Algo = "secp256k1" | "eth_secp256k1" | "ed25519" | "sr25519";
```
webmaster128, line 4: In this case we do not need "ethsecp256k1"?

packages/amino/src/pubkeys.ts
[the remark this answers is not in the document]
BigtoC: Should we also remove type casting in this line? https://github.com/cosmos/cosmjs/blob/main/packages/amino/src/pubkeys.ts#L24
```js
return (pubkey as Secp256k1Pubkey).type === "tendermint/PubKeySecp256k1";
```

aim586034-art (changes requested):
Tio

packages/cosmwasm-stargate/src/signingcosmwasmclient.ts
aim586034-art: Mungkin nanti saya susuli.

PR #1815, opened 2025-09-24
Clarify ESLint import plugin setup in flat config comments

alizfara112:
Replace the stale comment with accurate guidance: we use fixupPluginRules(importt) and selectively reuse importt.flatConfigs.typescript.rules; we intentionally don’t enable importt.flatConfigs.recommended to avoid rule duplication/conflicts.


dynst:
This actually conflicts with #1786

PR #1817, opened 2025-10-01
Remove .js extensions from @noble/hashes imports

webmaster128:


CHANGELOG.md
```diff
@@ -6,6 +6,18 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- @cosmjs/crypto: Fix import path of @noble/hashes to avoid bunding issue
+
+  > Error \[ERR_PACKAGE_PATH_NOT_EXPORTED\]: Package subpath './sha2.js' is not
+  > defined by "exports" in …
```
dynst, line 14: This error only occurs with versions <1.8.0. It was fixed in

https://github.com/paulmillr/noble-hashes/releases/tag/1.8.0

https://github.com/paulmillr/noble-hashes/commit/f40873ddff4a256d069b84eee46f768d44e19fce

webmaster128: Oh, very interesting. This explains why it only was a problem for one long time user and we did not see it in our CI.

We can bump the dependency to ^1.8.0 no problem.

PR #1847, opened 2025-10-15
amino: remove some type casts

dynst:
These did nothing; the discriminator `type` member is on the base type for this exact reason.


webmaster128:
Oh they do something: avoid typos in the type name string literal

dynst:
That does sound useful, at least when first writing the code. Added the missing unit tests to prevent potential future typos!

PR #1854, opened 2025-10-15
Package code in the dist/ directory

dynst:
And silence a warning about an unbound variable.

Split out of #1833 without transpiler changes.

### DIFF
--- a/packages/amino/package.json
+++ b/packages/amino/package.json
@@ -34,7 +34,7 @@
     "test": "yarn build-or-skip && yarn test-node",
     "coverage": "nyc --reporter=text --reporter=lcov yarn test --quiet",
     "build": "rm -rf ./dist/ && tsc",
-    "build-or-skip": "[ -n \"$SKIP_BUILD\" ] || yarn build",
+    "build-or-skip": "[ -n \"${SKIP_BUILD:-}\" ] || yarn build",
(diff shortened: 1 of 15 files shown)


webmaster128:
Why is "dist" better than "build"? Isn't is confusing to write all build artifacts into a "distribution" folder including those that should never be distributed like tests?

packages/encoding/karma.conf.cjs
```diff
@@ -15,7 +15,7 @@ module.exports = function (config) {
     frameworks: ["jasmine"],
 
     // list of files / patterns to load in the browser
-    files: ["dist/web/tests.js"],
+    files: ["build/web/tests.js"],
```
webmaster128: I like this change btw. As this is a purely dev build artifact, we can bundle from `build/` to `build/web/` or something like that to free the "dist" name

webmaster128:
I just merged the SKIP_BUILD change to main as it is a nice improvement independent of the rest of this PR

webmaster128:
Thank you. I manually took two out of 3 ideas from here and added them to main already. That should make things clearer now.

I'm not convinced that changing the primary build folder from build/ to dist/ is an improvement. Instead we build everything into build/ for now and in there we can have all sorts of secondary gabarge like Karma test artifacts. This frees the term dist/ for a potential future post-build step for publishing.

PR #1856, opened 2025-10-16
@scure/bip32

### LINKED ISSUE #1721: deprecating bip39 & bip32 code
What about undoing #966?

Now that a stronger (and less bloated?) dependency ecosystem exists, seems like most of this code could just be replaced with `@scure/bip39` and `@scure/bip32`, especially after migrating #1272.

dynst:
Closes #1721


dynst:
I thought replacing custom code with 2 small high quality paulmillr-written packages might ease maintainability similar to the switch to using `@scure/bip39`, as well as reduce code redundancy in app bundles that already pull in these packages. But good point. Feel free to close this if it's not worth it.

PR #1861, opened 2025-10-16
stargate, faucet: add missing .name field to Error subclasses

dynst:
They still had the default name `"Error"`.

### DIFF
--- a/packages/stargate/src/stargateclient.ts
+++ b/packages/stargate/src/stargateclient.ts
@@ -28,6 +28,7 @@ export class TimeoutError extends Error {
 
   public constructor(message: string, txId: string) {
     super(message);
+    this.name = "TimeoutError";
     this.txId = txId;
   }
 }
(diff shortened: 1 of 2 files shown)


packages/stargate/src/stargateclient.ts
```diff
@@ -179,6 +180,7 @@ export class BroadcastTxError extends Error {
 
   public constructor(code: number, codespace: string, log: string | undefined) {
     super(`Broadcasting transaction failed with code ${code} (codespace: ${codespace}). Log: ${log}`);
+    this.name = "BroadcastTxError";
```
webmaster128:
```suggestion
    this.name = this.constructor.name;
```

webmaster128 (commented):
Nice catch. Turns out we can even avoid the duplication of the name for ease of maintenance ([source](https://stackoverflow.com/questions/1382107/whats-a-good-way-to-extend-error-in-javascript#comment84156899_32750746)).

Issue #1750, opened 2025-07-29
lints


dynst:
These lints are recommended by the eslint TypeScript plugin but are currently disabled because they flag so many parts of the codebase.

- [ ] https://typescript-eslint.io/rules/no-empty-function/
- [ ] https://typescript-eslint.io/rules/no-empty-object-type/
- [ ] https://typescript-eslint.io/rules/no-explicit-any/
- [ ] https://typescript-eslint.io/rules/no-misused-promises/
- [ ] https://typescript-eslint.io/rules/no-redundant-type-constituents/
- [ ] https://typescript-eslint.io/rules/no-unnecessary-type-assertion/
- [ ] https://typescript-eslint.io/rules/no-unsafe-argument/
- [ ] https://typescript-eslint.io/rules/no-unsafe-assignment/
- [ ] https://typescript-eslint.io/rules/no-unsafe-call/
- [ ] https://typescript-eslint.io/rules/no-unsafe-enum-comparison/
- [ ] https://typescript-eslint.io/rules/no-unsafe-member-access/
- [ ] https://typescript-eslint.io/rules/no-unsafe-return/
- [ ] https://typescript-eslint.io/rules/prefer-promise-reject-errors/
- [ ] https://typescript-eslint.io/rules/restrict-template-expressions/
- [ ] https://typescript-eslint.io/rules/unbound-method/

Which should be turned on someday, if any?

webmaster128:
Nice! I like no-unsafe-member-access and no-unsafe-enum-comparison

dynst:
Currently 156 errors from running `no-unsafe-member-access`

PR #1867, opened 2025-10-20
simplify Promise code

dynst:
Replace `setTimeout()` with `await sleep()` and use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

This resolves most but not all warnings from the `@typescript-eslint/no-misused-promises` lint #1750


webmaster128 (approved):
Great thank you

PR #1905, opened 2025-11-11
enable 'no-unsafe-call' lint

dynst:
Part of #1750. Most warnings were in the ledger-amino web code.


packages/ledger-amino/src/demo/web.ts
```diff
@@ -8,8 +8,11 @@ import TransportWebUSB from "@ledgerhq/hw-transport-webusb";
 
 import { LedgerSigner } from "../ledgersigner";
 
-declare const window: any;
-declare const document: any;
+const getElement = (id: string): HTMLInputElement => {
+  const e = document.getElementById(id);
```
webmaster128: Does this compile? How does TS know `document.getElementById` exists here?

webmaster128 (commented):
Nice

PR #1911, opened 2025-11-18
add dynamic gasfee pricing as a signing client option - Feemarket/Osmosis

ertemann:
## Summary

Adds automatic gas price discovery for chains with dynamic feemarket modules (Osmosis EIP-1559 and Skip feemarket). Instead of manually setting static gas prices, CosmJS can now query current prices from the chain, preventing transaction failures from outdated pricing.

## Changes

- New feemarket.ts: Query dynamic gas prices from Osmosis EIP-1559 and Skip feemarket modules
- Updated clients: SigningStargateClient and SigningCosmWasmClient now support DynamicGasPriceConfig
- Refactoring: Extracted calculateFeeForTransaction() helper
- Tests: Some unit tests and an end-to-end validation on Osmosis and Neutron testnets

Usage

```
const client = await SigningStargateClient.connectWithSigner(rpc, signer, {
  gasPrice: {
    denom: "uosmo",
    multiplier: 1.3,
    minGasPrice: GasPrice.fromString("0.0025uosmo"),
    maxGasPrice: GasPrice.fromString("0.1uosmo"),
  }
});

// Gas price queried automatically per transaction
await client.signAndBroadcast(address, messages, "auto");

```

## Features
Graceful fallback to minGasPrice if feemarket query fails
Min/max constraint enforcement with configurable multiplier
backward compatible - existing code unchanged
Automatic chain detection for Osmosis variants

Inspired by [Hermes relayer PR #4004](https://github.com/informalsystems/hermes/pull/4004).

PR #1926, opened 2025-12-18
Add dynamic gas pricing as a signing client option - Feemarket/Osmosis

webmaster128:
Follow-up of #1911. Closes #1911

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,15 @@ and this project adheres to
   after ensuring `ArrayBuffer` is used. ([#1883])
 - @cosmjs/math: Add `Decimal.adjustFractionalDigits` which allows you to change
   the fractional digits of a Decimal without changing its value. ([#1916])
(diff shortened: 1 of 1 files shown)


packages/stargate/src/feemarket.ts
```diff
@@ -0,0 +1,224 @@
...
+export function multiplyDecimalByNumber(
+  value: Decimal,
+  multiplier: number,
+  fractionalDigits: number,
+): Decimal {
+  // Normalize value to target fractional digits
+  const normalizedValue = value.adjustFractionalDigits(fractionalDigits);
+
+  // Convert multiplier to Decimal with same fractional digits
+  const multiplierDecimal = Decimal.fromUserInput(multiplier.toString(), fractionalDigits);
```
webmaster128: I tried `multiplier.toFixed(fractionalDigits)` but it leads to some inconvenient effekts like this:

```
> 0.1.toFixed(18)
'0.100000000000000006'
```

i.e. we swap one problem with another. Maybe it is good enough as-is, assuming users don't put in `multiplier` values that lead to scientific notation.

PR #1940, opened 2025-12-31
Bring back main/types fields for bundlephobia support (0.37)

webmaster128:
Starting with 0.37.0 we get the error

> EntryPointError
> 
> We could not guess a valid entry point for this package. Perhaps the author hasn't specified one in its package.json ?

from bundlephobia (e.g. https://bundlephobia.com/package/@cosmjs/cosmwasm-stargate@0.37.0).

You can consider this a bundlephobia issue but it is annoying for us and there is no harm in having main/types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+- Bring back main/types fields for bundlephobia support ([#1940])
+
+[#1940]: https://github.com/cosmos/cosmjs/pull/1940
+
 ## [0.37.0] - 2025-10-29
 
 ### Added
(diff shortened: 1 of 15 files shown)


dynst:
For posterity: support for `"exports"` was added in [TypeScript 4.7](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/) in 2022. Meanwhile Bundlephobia has had an issue about it since 2020 https://github.com/pastelsky/bundlephobia/issues/379 but Bundlephobia has been all but unmaintained abandonware for years, until new development began again in January 2026. As of May, support for `"exports"` was being worked on in https://github.com/pastelsky/package-build-stats/pull/73.

Also Bundlephobia only needs `"main"` to know where the entrypoint for the package is - I don't think it benefits from `"types"` at all.

PR #1953, opened 2026-03-06
Migrate to pnpm

### LINKED ISSUE #1922: Yarn Plug'n'Play causes issues for on Mac
When you checkout this repo and then on a Mac do `yarn install`, you get `.yarn/cache/@unrs-resolver-binding-linux-x64-gnu-npm-1.11.1-93a00570de-10c0.zip` deleted and `.yarn/cache/@unrs-resolver-binding-darwin-arm64-npm-1.11.1-c78d4bd2cb-10c0.zip` added. This is indicating an uncommitted change. However, if you commit this, the Linux users and the CI system are in the same situtation but the other way around.

See `git status`:

```
On branch main
Your branch is up to date with 'origin/main'.

You are currently bisecting, started from branch 'main'.
  (use "git bisect reset" to get back to the original branch)

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        deleted:    .yarn/cache/@unrs-resolver-binding-linux-x64-gnu-npm-1.11.1-93a00570de-10c0.zip

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        .yarn/cache/@unrs-resolver-binding-darwin-arm64-npm-1.11.1-c78d4bd2cb-10c0.zip

no changes added to commit (use "git add" and/or "git commit -a")
```

I work around this issue for months by not committing the change. But especially when there are other dependency changes this is hard to deal with. For external contributors this is even harder to unserstand and treat correctly.

The source is [unrs-resolver with the system dependent optional dependencies](https://github.com/cosmos/cosmjs/blob/ed46c382066d9cc47b187f36bfd0b21e4e713800/yarn.lock#L7666-L7731) in combination with Yarn Plug'n'Play.

If anyone has suggestions how to better handle this case, happy to hear.

2wheeh:
Yarn PnP has known quirks around ESM (#1804, https://yarnpkg.com/getting-started/qa#does-yarn-support-esm, https://github.com/yarnpkg/berry/discussions/4044#discussioncomment-11395531) and platform-specific cache issues (#1922). Moving to pnpm avoids these concerns entirely, so figured it's worth proposing the switch.  

Closes #1922


package.json
```diff
@@ -25,21 +25,21 @@
     "packages/*"
   ],
   "scripts": {
-    "format": "yarn workspaces foreach --all --topological-dev --parallel --verbose run format",
-    "format-text": "prettier --write \"./*.md\" \"./docs/*.md\" && yarn workspaces foreach --all --topological-dev --parallel --verbose run format-text",
+    "format": "pnpm -r run format",
```
2wheeh, line 28: pnpm's -r (recursive) flag runs commands across all workspace packages with topological sorting enabled by default (--sort is true), so no extra flags are needed.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,5 @@
+packages:
+  - packages/*
+
+catalog:
+  "@types/node": ~24
```
2wheeh, line 5: Previously all packages used `"@types/node": "*"`, and `yarn.lock` happened to pin it to v22. 

While updating, `*` resolved to v25 which introduced breaking type changes (Uint8Array<ArrayBufferLike> incompatible with BufferSource). 

Since CI runs Node.js 24, pin `@types/node: ~24`

PR #1963, opened 2026-06-19
Validate RFC3339 date/time component ranges

haoshengzhen:
## Summary

This adds explicit range validation to `fromRfc3339` before constructing a `Date`.

Previously, inputs that matched the RFC3339-shaped regex but contained out-of-range values could be silently normalized by JavaScript `Date`, for example month `13`, February `30`, hour `24`, or timezone offset `+24:00`. Those inputs are now rejected with the existing RFC3339 format error.

The tests also cover a valid leap day to ensure valid calendar dates continue to parse correctly.

## Testing

- `git diff --check`
- Targeted local RFC3339 validation check covering valid leap-day/year cases and out-of-range components


haoshengzhen:
@codingki @webmaster128 Hi, Could you please review this PR at your convenience? Thank you very much.

